PR #5, opened 2022-09-23
First working version

EmreOzkose:


cmake/kaldi-native-fbank.cmake
```diff
@@ -0,0 +1,41 @@
+function(download_kaldi_native_fbank)
```
csukuangfj, line 1: Could you replace it with
https://github.com/k2-fsa/sherpa-ncnn/blob/master/cmake/kaldi-native-fbank.cmake

Compiling kaldi-native-fbank from source makes the user's life easier.

---

You need to add something like the following line:
https://github.com/k2-fsa/sherpa-ncnn/blob/0a65dbe30a57e3da70aa8a28d756b51f2b680929/sherpa-ncnn/csrc/CMakeLists.txt#L11
```
target_link_libraries(sherpa-ncnn-core kaldi-native-fbank-core ncnn)
```

That is, you need to link with `kaldi-native-fbank-core `.

csukuangfj:
If you think it is ready, you can merge it first. The refactoring can be done in a separate PR.

Thanks!

csukuangfj:
Would you mind adding github actions to ensure that it compiles and and decode some wave files?
It can be added in a separate PR.

You can find github actions are added in sherpa-ncnn, sherpa, icefall, and k2.

PR #46, opened 2023-02-20
Add python-api-examples: speech-recognition-from-microphone.py

yfyeung:


python-api-examples/speech-recognition-from-microphone.py
```diff
@@ -0,0 +1,66 @@
...
+        sample_rate=16000,
+        feature_dim=80,
+    )
+    return recognizer
+
+
+def main():
+    print("Started! Please speak")
+    recognizer = create_recognizer()
+    sample_rate = recognizer.sample_rate
```
csukuangfj: Please use 16000 directly.

recognizer in sherpa-onnx does not have the attribute `sample_rate`.

You can find the details of OnlineRecognizer at
https://github.com/k2-fsa/sherpa-onnx/blob/master/sherpa-onnx/python/sherpa_onnx/online_recognizer.py#L17

python-api-examples/speech-recognition-from-microphone.py
```diff
@@ -0,0 +1,66 @@
...
+    samples_per_read = int(0.1 * sample_rate)  # 0.1 second = 100 ms
+    last_result = ""
+    with sd.InputStream(channels=1, dtype="float32", samplerate=sample_rate) as s:
+        while True:
+            samples, _ = s.read(samples_per_read)  # a blocking read
+            samples = samples.reshape(-1)
+            recognizer.accept_waveform(sample_rate, samples)
```
csukuangfj: please follow
https://github.com/k2-fsa/sherpa-onnx/blob/063c68c91257ec5cf528f94e0b28eca095f8e1fd/python-api-examples/decode-file.py#L48-L50

In `sherpa-onnx`, we have a `stream` per utterance, which is used to support batch processing.

python-api-examples/speech-recognition-from-microphone.py
yfyeung: OK, I will fix it.

python-api-examples/speech-recognition-from-microphone.py
yfyeung: OK, I will fix it.

csukuangfj:
Thanks!

PR #59, opened 2023-02-23
Support Android

csukuangfj:


csukuangfj:
It works well on an Android `x86_64` simulator. However, it stops working after a few seconds on some Android phone.

Will try to debug the issue later.

PR #88, opened 2023-03-07
fix cmake parse path string error

frankyoujian:
@csukuangfj Verified locally, this change works for both conditions where Downloads directory have .zip or not.


csukuangfj:
Thanks!

PR #108, opened 2023-03-31
fix garbled console output with chinese characters

dongqianzhuan:
chinese characters can normal show in console

### DIFF
--- a/python-api-examples/speech-recognition-from-microphone-with-endpoint-detection.py
+++ b/python-api-examples/speech-recognition-from-microphone-with-endpoint-detection.py
@@ -125,10 +125,10 @@ def main():
 
             if result and (last_result != result):
                 last_result = result
-                display.print(segment_id, result)
-
+                print("\r{}:{}".format(segment_id, result), end="", flush=True)
             if is_endpoint:
                 if result:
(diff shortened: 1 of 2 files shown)


csukuangfj:
Thanks!

danpovey:
what is the point of the sherpa_onnx.Display ?  He seems to be using normal printing instead?

dongqianzhuan:
> what is the point of the sherpa_onnx.Display ? He seems to be using normal printing instead?
i think it is fprintf bug,in the different format console ,may be have some problem.like chineses in winows cmd console

csukuangfj:
> He seems to be using normal printing instead?

The key to his usage of printing is `end=""` and `\r`, though it cannot handle the case when a sentence expands multiple lines.

> what is the point of the sherpa_onnx.Display ? 

`sherpa_onnx.Display()` clears the previously displayed results from the terminal. To give you a concrete example,
when the recognition result is `what day is today`, instead of displaying it in the following form
```
what
what day
what day is
what day is today
```
`sherpa_onnx.Display()` displays the whole sentence in a single line.

csukuangfj:
The issue is that on **some** (not all) windows systems, `sherpa_onnx.Display()` cannot handle the encoding correctly, while python's `print` works fine.

PR #120, opened 2023-04-09
Use log probs for paraformer

pengzhendong:


sherpa-onnx/csrc/math.h
```diff
@@ -103,5 +103,18 @@ std::vector<int32_t> TopkIndex(const T *vec, int32_t size, int32_t topk) {
   return index;
 }
 
+template <class T>
```
csukuangfj: Please follow
https://github.com/k2-fsa/sherpa-onnx/blob/80060c276d2d5d75c9b2f24a132d7694ed1e4942/sherpa-onnx/csrc/offline-ctc-greedy-search-decoder.cc#L35-L39

to use `std::max_element` to replace `ArgMax()`.

You don't need to reimplement the wheel.

sherpa-onnx/csrc/offline-paraformer-greedy-search-decoder.cc
```diff
@@ -6,26 +6,30 @@
...
+  std::vector<int64_t> shape = log_probs.GetTensorTypeAndShapeInfo().GetShape();
   int32_t batch_size = shape[0];
   int32_t num_tokens = shape[1];
+  int32_t vocab_size = shape[2];
 
   std::vector<OfflineParaformerDecoderResult> results(batch_size);
 
-  const int64_t *p = token_num.GetTensorData<int64_t>();
+  const float *p = log_probs.GetTensorData<float>();
   for (int32_t i = 0; i != batch_size; ++i) {
```
csukuangfj:
```suggestion
  
  for (int32_t i = 0; i != batch_size; ++i) {
    const float *p = log_probs.GetTensorData<float>() + i * num_tokens * vocab_size;
```

sherpa-onnx/csrc/math.h
pengzhendong: `p_log_probs` is `const float *`, why do we use `static_cast` again?

csukuangfj: > `p_log_probs` is `const float *`, why do we use `static_cast` again?

yes, you are right. please remove the cast.

csukuangfj:
Thanks!

PR #131, opened 2023-04-26
share GetHypsRowSplits interface and fix getting Topk not taking logprob

pingfengluo:


csukuangfj:
Thanks!
By the way, could you use std::partial_sort to replace std::sort in top_k?
(It can be done in a separate PR).

You can use the following code as a reference implementation.
https://github.com/flashlight/text/blob/main/flashlight/lib/text/decoder/LexiconDecoder.cpp#L43
```cpp
    std::iota(idx.begin(), idx.end(), 0);
    if (N > opt_.beamSizeToken) {
      std::partial_sort(
          idx.begin(),
          idx.begin() + opt_.beamSizeToken,
          idx.end(),
          [&t, &N, &emissions](const size_t& l, const size_t& r) {
            return emissions[t * N + l] > emissions[t * N + r];
          });
    }

```

pingfengluo:
> Thanks! By the way, could you use std::partial_sort to replace std::sort in top_k? (It can be done in a separate PR).
> 
> You can use the following code as a reference implementation. https://github.com/flashlight/text/blob/main/flashlight/lib/text/decoder/LexiconDecoder.cpp#L43
> 
> ```c++
>     std::iota(idx.begin(), idx.end(), 0);
>     if (N > opt_.beamSizeToken) {
>       std::partial_sort(
>           idx.begin(),
>           idx.begin() + opt_.beamSizeToken,
>           idx.end(),
>           [&t, &N, &emissions](const size_t& l, const size_t& r) {
>             return emissions[t * N + l] > emissions[t * N + r];
>           });
>     }
> ```

ok, i will make another PR

PR #136, opened 2023-05-01
Add stack support

jingzhaoou:
The current code only has `Cat`, etc. `Stack` should be useful when supporting streaming conformer models (specifically, in `StackStates` and `UnStackStates`). Please share your feedback and I will modify my code accordingly. 

Added some tentative code for streaming conformer model support.

### DIFF
--- a/android/SherpaOnnx/app/src/main/java/com/k2fsa/sherpa/onnx/MainActivity.kt
+++ b/android/SherpaOnnx/app/src/main/java/com/k2fsa/sherpa/onnx/MainActivity.kt
@@ -182,9 +182,10 @@ class MainActivity : AppCompatActivity() {
         val config = OnlineRecognizerConfig(
             featConfig = getFeatureConfig(sampleRate = sampleRateInHz, featureDim = 80),
             modelConfig = getModelConfig(type = type)!!,
+            lmConfig = getOnlineLMConfig(type = type),
             endpointConfig = getEndpointConfig(),
             enableEndpoint = true,
-            decodingMethod = "greedy_search",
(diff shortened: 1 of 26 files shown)


csukuangfj:
Thanks! It looks good to me.

Could you please fix the style issues? After fixing it, we can merge this PR. 

You can check the style issues locally by running:
```bash
cd /path/to/sherpa-onnx
./scripts/check_style_cpplint.sh 2
```

---

By the way, do you plan to also add `unstack` in a separate pull-request?

jingzhaoou:
I just changed my code and passed the style checker. I will ensure all the workflows are passed (they take a while to run in GitHub). 

> By the way, do you plan to also add `unstack` in a separate pull-request?

If there is any existing streaming conformer model that I can use to test my code, I can work on the `stack/unstack` code right away. 

I can combine things in a single PR or separate PRs. Just let me know what you suggest/prefer. Thanks.

jingzhaoou:
I will try out this conformer model and see how far I can go with the streaming flow. Fingers crossed.

https://k2-fsa.github.io/sherpa/onnx/pretrained_models/offline-transducer/conformer-transducer-models.html#csukuangfj-sherpa-onnx-conformer-en-2023-03-18-english

jingzhaoou:
@csukuangfj  I got the offline conformer model to work. However, the model does not seem to work for the streaming flow. I will wait for your model then. 

I committed my code related to streaming conformer model support. I don't have a way to verify things work or not. I will wait till I can verify everything before merging my PR. 

As always, any suggestions are highly appreciated.

Issue #40, opened 2023-02-20
[Help wanted] Support TensorRT


csukuangfj:
# TODO

- [ ] Support GPU via TensorRT

See https://onnxruntime.ai/docs/execution-providers/TensorRT-ExecutionProvider.html

yuekaizhang:
I would like take on this. 

- [ ] Support the Onnxruntime CUDA provider.

manickavela29:
Hi @csukuangfj ,

Observed that currently only CUDA EP support is there and TensorRT EP support is not there for onnxruntime.
is there ay active developments going on for TensorRT GPU backend?

csukuangfj:
> is there ay active developments going on for TensorRT GPU backend?

We don't have a plan to support it in the near future. Would you like to contribute?

manickavela29:
I tried adding triggering onnxruntime's tensorrt ep for zipfromer but the model performance was very bad,
debugging further with standalone onnxruntime in python for Encoder models, will update if I see some good results.

manickavela29:
Hi @csukuangfj,
TensorRT has several parameters, and these will be only valid if TensorRT provider is chosen,
so I need your suggestion on either of below 2.

1. Putting TRT configs as part of the model-config.cc file [model-config.cc](https://github.com/k2-fsa/sherpa-onnx/blob/1371c6b3f08e33784380f9cd76832bdf83c166bf/sherpa-onnx/csrc/online-model-config.h#L27)
2. Creating a new config for TRT and exposing the required parameters from it.

Thank you

csukuangfj:
Could you create a new config for tensorrt and add this config as a member field of `OnlineModelConfig` and `OfflineModelConfig`?

You can set the default values of this config as the one used in
https://github.com/k2-fsa/sherpa-onnx/blob/b7148174739275dfc997af726be364245511239c/sherpa-onnx/csrc/session.cc#L137-L150

manickavela29:
yes, I will send the PR for configs separately in some time.

manickavela29:
Current perf Cuda vs Trt

csrc/online-zipformer2-transducer-model.cc:RunEncoder:445 Encoder Duration : **1.930044** ms
csrc/online-zipformer2-transducer-model.cc:RunEncoder:445 Encoder Duration :  **0.034984** ms
csrc/online-zipformer2-transducer-model.cc:RunEncoder:445 Encoder Duration :  **0.034912** ms
[...]

PR #138, opened 2023-05-04
[WIP] Add onnxruntime-gpu support

yuekaizhang:
This PR added onnxruntime-gpu infer support for paraformer.  Related to #40.

Also, added python scripts to decode manifest files like this https://github.com/k2-fsa/sherpa/blob/master/sherpa/bin/pruned_transducer_statelessX/decode_manifest.py. (I could remove it from this PR if you don't want this part. )

TODO:
- [ ] Update docs
- [ ] Support other models
- [ ] Add perf results


sherpa-onnx/csrc/offline-paraformer-model.cc
```diff
@@ -23,6 +23,10 @@ class OfflineParaformerModel::Impl {
     sess_opts_.SetIntraOpNumThreads(config_.num_threads);
     sess_opts_.SetInterOpNumThreads(config_.num_threads);
 
+#ifdef ENABLE_GPU
```
csukuangfj, line 26: Could you prepend it with `SHERPA_ONNX`?
That is, use `SHERPA_ONNX_ENABLE_GPU`.

All macros in sherpa-onnx are expected to start with `SHERPA_ONNX_`

CMakeLists.txt
```diff
@@ -19,6 +19,11 @@ option(SHERPA_ONNX_ENABLE_PORTAUDIO "Whether to build with portaudio" ON)
 option(SHERPA_ONNX_ENABLE_JNI "Whether to build JNI internface" OFF)
 option(SHERPA_ONNX_ENABLE_C_API "Whether to build C API" ON)
 option(SHERPA_ONNX_ENABLE_WEBSOCKET "Whether to build webscoket server/client" ON)
+option(SHERPA_ONNX_ENABLE_GPU "Enable ONNX Runtime GPU support" OFF)
+
+if(SHERPA_ONNX_ENABLE_GPU)
+  add_definitions(-DENABLE_GPU)
```
csukuangfj, line 25: Could you please move it to
https://github.com/k2-fsa/sherpa-onnx/blob/aa7108729bd07e773f03ea1a49eb8bb27367ff8b/sherpa-onnx/csrc/CMakeLists.txt#L76

`add_definitions` defines a macro in the global scope but we want to limit it only to `sherpa-onnx-core`.

jingzhaoou:
I tried to compile the code with `-DSHERPA_ONNX_ENABLE_GPU:BOOL=ON` and ran into the following error:

```
CMake Error at cmake/onnxruntime.cmake:231 (message):
  GPU download is not supported for now.
```

Can you share which onnxruntime I should download? Thanks.

yuekaizhang:
> I tried to compile the code with `-DSHERPA_ONNX_ENABLE_GPU:BOOL=ON` and ran into the following error:
> 
> ```
> CMake Error at cmake/onnxruntime.cmake:231 (message):
>   GPU download is not supported for now.
> ```
> 
> Can you share which onnxruntime I should download? Thanks.

https://github.com/microsoft/onnxruntime/releases/download/v1.14.1/onnxruntime-linux-x64-gpu-1.14.1.tgz Or, you could wait this PR finished.

sherpa-onnx/csrc/offline-paraformer-model.cc
```diff
@@ -23,6 +23,10 @@ class OfflineParaformerModel::Impl {
...
+    OrtCUDAProviderOptions cuda_options;
+    sess_opts_.AppendExecutionProvider_CUDA(cuda_options);
```
jingzhaoou, line 28: I see many `Ort::SessionOptions` and `Ort::Allocator` handlings scattered throughout the code base. Maybe we should derive a session/allocator management class, which can be consistently applied to all models. Otherwise, we may end up having a number of such `#if` statements. Just my 2 cents. I can help to work on this. I happen to be looking at enable Intel OpenVINO these days. :-)

jingzhaoou: Note that this is probably out of scope of this PR. Just want to share some thoughts. Thanks.

yuekaizhang: @csukuangfj What do you think? In this way, we may not need to set session separately for every model.

csukuangfj: I suggest that we remove the macro and then add a new field to
https://github.com/k2-fsa/sherpa-onnx/blob/0992063de839adc41e55d0ee8a0ee1b9e4248a94/sherpa-onnx/csrc/offline-model-config.h#L15
and 
https://github.com/k2-fsa/sherpa-onnx/blob/0992063de839adc41e55d0ee8a0ee1b9e4248a94/sherpa-onnx/csrc/online-transducer-model-config.h#L13

```c++
string provider = "cpu";
```

And add the following code
```c++
if (config_.provider == "cuda") {
    OrtCUDAProviderOptions cuda_options;
    sess_opts_.AppendExecutionProvider_CUDA(cuda_options);
}
```

If we want to support other providers, such as
- `OpenVINO`
- `DirectML`
- ARM compute library
- `NNAPI`

we can add more if/else statements.

(Or we can define some enums and use a switch to replace if/else).

---

> I see many Ort::SessionOptions and Ort::Allocator handlings scattered throughout the code base. Maybe we should derive a session/allocator management class, which can be consistently applied to all models

Can we define a function
```cpp
Ort::SessionOptions GetSessionOptions(const OfflineModelConfig& config) {

}
```
and an overload
```cpp
Ort::SessionOptions GetSessionOptions(const OnlineTransducerModelConfig& config) {

}
```
to reduce duplicate code?

As for the allocator, I think the current approach is fine.

yuekaizhang:
Closed it since resolved by #153. Would make separate PRs to add benchmarks.

PR #141, opened 2023-05-05
Stack and streaming conformer support

jingzhaoou:
The current code only has `Cat`, etc. `Stack` should be useful when supporting streaming conformer models (specifically, in `StackStates` and `UnStackStates`). Please share your feedback and I will modify my code accordingly. 

Added some tentative code for streaming conformer model support. 

PS: sorry that I am new to do development on a forked repo. I need to close [my previous PR](https://github.com/k2-fsa/sherpa-onnx/pull/136) (created from the `master` branch) and open a new one.

This PR is ready to merge. I verified it to work with a streaming conformer model published by @csukuangfj .


csukuangfj:
@jingzhaoou  Do you have time to finish this pull-request?

jingzhaoou:
@csukuangfj  I just saw the new streaming conformer model you published, which is very exciting. Yeah, I have the time to finish this pull-request and am working on it now. Thanks a lot!

sherpa-onnx/csrc/stack.cc
```diff
@@ -76,13 +76,13 @@ Ort::Value Stack(OrtAllocator *allocator,
   T *dst = ans.GetTensorMutableData<T>();
 
   for (int32_t i = 0; i != leading_size; ++i) {
-      for (int32_t n = 0; n != static_cast<int32_t>(values.size()); ++n) {
-          const T *src = values[n]->GetTensorData<T>();
-          src += i * trailing_size;
+    for (int32_t n = 0; n != static_cast<int32_t>(values.size()); ++n) {
```
csukuangfj, line 79: I suggest that you handle the case `values.size() == 1` specially.

If the input vector contains only a single tensor, we return a copy of it after changing its shape.

jingzhaoou:
@csukuangfj  I dumped the metadata of your streaming conformer model. It only has `left_context=64`. Should it also contain `right_context=4`? For now, I just hard coded the right context size. 

I added more code to support streaming conformer and currently ran into the following error:

```
state: 18,64,1,512,
state: 18,30,1,512,
===> model_->RunEncoder (begin)
terminate called after throwing an instance of 'Ort::Exception'
  what():  Unexpected input data type. Actual: (tensor(int32)) , expected: (tensor(int64))
``` 

I think the `encoder_inputs` is not constructed correctly. I will continue to dig further tomorrow. If you have any suggestions for me,I would highly appreciate it. 

```cpp
std::pair<Ort::Value, std::vector<Ort::Value>>
OnlineConformerTransducerModel::RunEncoder(Ort::Value features,
                                           std::vector<Ort::Value> states,
                                           Ort::Value processed_frames) {
  std::array<Ort::Value, 4> encoder_inputs = {
      std::move(features), std::move(states[0]), std::move(states[1]),
      std::move(processed_frames)};
```

csukuangfj:
Please refer to
- https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/pruned_transducer_stateless5/export-onnx-streaming.py#L327
- https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/pruned_transducer_stateless5/onnx_pretrained-streaming.py

for usages.

---

As for you error, you can fix it by changing

https://github.com/k2-fsa/sherpa-onnx/blob/3c38ad07b69c5ccd5f1a730fc687c802d44e5e0f/sherpa-onnx/csrc/online-recognizer.cc#L191
to
```cpp
 std::vector<int64_t> all_processed_frames(n); 
```

That is, use `int64_t` to replace `int32_t`.

sherpa-onnx/csrc/online-transducer-model.h
```diff
@@ -71,7 +71,8 @@ class OnlineTransducerModel {
    */
   virtual std::pair<Ort::Value, std::vector<Ort::Value>> RunEncoder(
       Ort::Value features,
-      std::vector<Ort::Value> states) = 0;  // NOLINT
+      std::vector<Ort::Value> states,
+      Ort::Value processed_frames) = 0;  // NOLINT
```
csukuangfj, line 76: please document that `processed_frames` is counted before subsampling and it is a 1-D tensor with data type `int64_t`.

sherpa-onnx/csrc/online-lstm-transducer-model.cc
```diff
@@ -227,7 +227,8 @@ std::vector<Ort::Value> OnlineLstmTransducerModel::GetEncoderInitStates() {
 
 std::pair<Ort::Value, std::vector<Ort::Value>>
 OnlineLstmTransducerModel::RunEncoder(Ort::Value features,
-                                      std::vector<Ort::Value> states) {
+                                      std::vector<Ort::Value> states,
+                                      Ort::Value processed_frames) {
```
csukuangfj:
```suggestion
                                      std::vector<Ort::Value> states,
                                      Ort::Value /*processed_frames*/) {
```
otherwise, the compiler will complain the input parameter `processed_frames` is never used.

sherpa-onnx/csrc/online-zipformer-transducer-model.cc
```diff
@@ -434,7 +434,8 @@ std::vector<Ort::Value> OnlineZipformerTransducerModel::GetEncoderInitStates() {
 
 std::pair<Ort::Value, std::vector<Ort::Value>>
 OnlineZipformerTransducerModel::RunEncoder(Ort::Value features,
-                                           std::vector<Ort::Value> states) {
+                                           std::vector<Ort::Value> states,
+                                           Ort::Value processed_frames) {
```
csukuangfj:
```suggestion
                                           Ort::Value /*processed_frames*/) {
```

sherpa-onnx/csrc/online-recognizer.cc
```diff
@@ -210,9 +215,26 @@ class OnlineRecognizer::Impl {
                                             features_vec.size(), x_shape.data(),
                                             x_shape.size());
 
+    std::array<int64_t, 1> processed_frames_shape{(int64_t) all_processed_frames.size()};
```
csukuangfj:
```suggestion
    std::array<int64_t, 1> processed_frames_shape{all_processed_frames.size()};
```

sherpa-onnx/csrc/online-conformer-transducer-model.h
```diff
@@ -0,0 +1,108 @@
...
+  int32_t T_ = 0;
+  int32_t decode_chunk_len_ = 0;
+  int32_t cnn_module_kernel_ = 0;
+  int32_t context_size_ = 0;
+  int32_t left_context_ = 0;
+  // TODO: to get from model
+  int32_t right_context_ = 4;
+  int32_t encoder_dim_ = 0;
+  int32_t pad_length_ = 0;
+  int32_t vocab_size_ = 0;
```
csukuangfj, line 103: I assume these fields can be read from the model, right?

sherpa-onnx/csrc/online-lstm-transducer-model.cc
jingzhaoou: Changed the code as suggested. Thanks, @csukuangfj.

sherpa-onnx/csrc/online-zipformer-transducer-model.cc
jingzhaoou: Changed the code as suggested. Thanks, @csukuangfj.

sherpa-onnx/csrc/online-conformer-transducer-model.h
jingzhaoou: Except for `right_context` (missing in the model metadata), all the other parameters are retrieved from the model. I recommended earlier to add `right_context` to the ONNX export script. 

```cpp
  SHERPA_ONNX_READ_META_DATA(left_context_, "left_context");
  SHERPA_ONNX_READ_META_DATA(encoder_dim_, "encoder_dim");
  SHERPA_ONNX_READ_META_DATA(pad_length_, "pad_length");
  SHERPA_ONNX_READ_META_DATA(cnn_module_kernel_, "cnn_module_kernel");
......
  SHERPA_ONNX_READ_META_DATA(vocab_size_, "vocab_size");
```

sherpa-onnx/csrc/online-transducer-model.h
jingzhaoou: Added document as suggested. 

```cpp
   * @param processed_frames  Processed frames before subsampling. It is a 1-D tensor with data type int64_t.
```

sherpa-onnx/csrc/online-recognizer.cc
jingzhaoou: `all_processed_frames.size()` returns an unsigned integer. Static cast is required to avoid compiler warnings. I changed the code to look like

```cpp
    std::array<int64_t, 1> processed_frames_shape{
      static_cast<int64_t>(all_processed_frames.size())};
```

Let me know if that looks good. Thanks.

sherpa-onnx/csrc/online-conformer-transducer-model.h
csukuangfj: @jingzhaoo 

I just updated the encoder model to include `right_context` in its metadata.

By the way, I don't think you need right_context in sherpa-onnx.

Is this PR ready to merge?

jingzhaoou: @csukuangfj thanks for adding `right_context` to the model metadata. Look like both `left_context` and `right_contex` are not required in `sherpa-onnx`. 

I just committed all I have into the PR. Things are not working yet. I think stacking and unstacking states are not working as expected. I ran into segmentation fault:

```
===> StackStates (start)                                                                                                                    
===> StackStates (end)                                                                                                                      
state: 18,64,1,512,                                                                                                                         
state: 18,30,1,512,                                                                                                                         
===> model_->RunEncoder (begin)                                                                                                             
===> model_->RunEncoder (end)                                                                                                               
===> model_->Decoder (begin)                                                                                                                
===> model_->Decoder (end)                                                                                                                  
===> model_->UnStackStates (begin)                                                                                                          
===> model_->UnStackStates (end)                                                                                                            
===> set (start: 1, 1, 64)                                                                                                                  
===> set (end)                                                                                                                              
Segmentation fault (core dumped) 
```

I jhate to hold off any progress of this project. I suggest that we can merge the PR (or through other means) so that you can debug the parts that I get wrong. Since this is a new feature, I don't see it breaking any existing features. Please let me know what you suggest. Thanks.

jingzhaoou: When the code runs to 

```cpp
    for (int32_t i = 0; i != n; ++i) {
      ss[i]->SetResult(results[i]);
      ss[i]->SetStates(std::move(next_states[i]));
    }
```

`next_states.size()` is 64, rather than 1. Thus, the output of `next_states = model_->UnStackStates(pair.second)` is incorrect due to some reason. I am still thinking why that is the case. Not much clue yet.

csukuangfj: > I jhate to hold off any progress of this project. I suggest that we can merge the PR (or through other means) so that you can debug the parts that I get wrong.

No worry. I am helping debug it. I will try to find out what goes wrong.

jingzhaoou: @csukuangfj Thanks a lot! I looked at `pair = model_->RunEncoder(...)`. `pair.second` has two `Ort::Value`s with shapes of `(18, 64, 1, 512)` and `(18, 30, 1, 512)`, which should be correct based on my understanding of the model init process. 

After the following two `Unbind`s:

```cpp
  std::vector<Ort::Value> attn_vec = Unbind(allocator_, &states[0], 2);
  std::vector<Ort::Value> conv_vec = Unbind(allocator_, &states[1], 2);
```

`attn_vec[0]` has shape `(18, 64, 1, 512)` while `conv_vec[0]` has shape `(18, 30, 1, 512)`. This seems incorrect. `Unbind` has a different behavior from `torch::unbind`. I suspect we may need to write a different `Unbind` function here.

sherpa-onnx/csrc/stack.cc
jingzhaoou: @csukuangfj  once things are working, I will add code to handle the `values.size() == 1` case specifically. Thanks.

sherpa-onnx/csrc/online-conformer-transducer-model.h
csukuangfj: @jingzhaoo 

I just realized there is no need to implement `Stack()`.
We can change the init states to 4-D tensors and use Cat() to stack them.

---

Please see
https://github.com/jingzhaoou/sherpa-onnx/pull/1


---

Also, please remove the debug statements so that we can merge it.

jingzhaoou: @csukuangfj Indeed! After merging your PR to my branch, things are working correctly. You are so awesome! I am cleaning my debug statements now. 

The stack code can stay just in case we need it later on. Let me know if you think otherwise.

csukuangfj: > The stack code can stay just in case we need it later on.

Yes, please keep the stack code.

jingzhaoou:
@csukuangfj  I just removed all the debug statements. Let me know if you would like any changes before merging this PR. Thanks.

csukuangfj:
> @csukuangfj I just removed all the debug statements. Let me know if you would like any changes before merging this PR. Thanks.

@jingzhaoo 

Thanks! 

Could you please add a test to
https://github.com/k2-fsa/sherpa-onnx/blob/master/.github/scripts/test-online-transducer.sh

---

You can use the model from
https://huggingface.co/csukuangfj/sherpa-onnx-streaming-conformer-en-2023-05-09

jingzhaoou:
@csukuangfj it is almost midnight here. I will work on adding the requested tests in my tomorrow morning.

csukuangfj:
> @csukuangfj it is almost midnight here. I will work on adding the requested tests in my tomorrow morning.

Thanks!

jingzhaoou:
@csukuangfj Great, will do then.

sherpa-onnx/csrc/stack.cc
jingzhaoou: @csukuangfj  I spent some time to think about how to deal with the special `values.size() == 1` case efficiently. `Cat()` can do a simple `Clone()`. But for `Stack()`, I don't see how we can have a shortcut to speed things up. I still need to get `v0_shape` as `Stack()` can happen in `dim=0` or `dim=1`, which result in different outputs. 

Let me know if you have other suggestions. For now, I recommend not to add code to deal with this special case. Thanks.

PR #145, opened 2023-05-07
Implement context biasing with a Aho Corasick automata

pkufool:


sherpa-onnx/csrc/offline-recognizer-transducer-impl.h
```diff
@@ -72,8 +73,15 @@ class OfflineRecognizerTransducerImpl : public OfflineRecognizerImpl {
...
-  std::unique_ptr<OfflineStream> CreateStream() const override {
-    return std::make_unique<OfflineStream>(config_.feat_config);
+  std::unique_ptr<OfflineStream> CreateStream(
+      const std::vector<std::vector<int32_t>> &context_list) const override {
+    // We create context_graph at this level, because we might have default
+    // context_graph(will add later if need) that belong to the whole model
+    // rather than each stream.
+    ContextGraphPtr context_graph =
+        std::make_shared<ContextGraph>(ContextGraph(config_.context_score));
+    context_graph->Build(context_list);
```
csukuangfj:
```suggestion
    auto context_graph =
        std::make_shared<ContextGraph>(config_.context_score);
    context_graph->Build(context_list);
```

sherpa-onnx/csrc/offline-recognizer.h
```diff
@@ -56,7 +58,8 @@ class OfflineRecognizer {
   explicit OfflineRecognizer(const OfflineRecognizerConfig &config);
 
   /// Create a stream for decoding.
-  std::unique_ptr<OfflineStream> CreateStream() const;
+  std::unique_ptr<OfflineStream> CreateStream(
```
csukuangfj, line 64: I suggest that you create an overload version of CreateStream().
People who don't use context graph related stuff don't need to construct a vector. (You don't pay for what you don't need.)

pkufool:
I only support python offline interface, I am going to find some solid validatin set, but I think it is almost ready for you guys to have a look.

pkufool:
A simple test:

Command without contexts:
```bash
python python-api-examples/offline-decode-files.py \
  --tokens exp-zh/tokens.txt \
  --modeling-unit char \
  --encoder exp-zh/encoder-epoch-99-avg-1.onnx \
  --decoder exp-zh/decoder-epoch-99-avg-1.onnx \
  --joiner exp-zh/joiner-epoch-99-avg-1.onnx \
  --decoding-method modified_beam_search  \
  exp-zh/test_wavs/3.wav exp-zh/test_wavs/4.wav exp-zh/test_wavs/5.wav exp-zh/test_wavs/6.wav
```
The result:
```
Started!                                                                                                                                                       
contexts_list : []                                                                                                                                             
contexts_list : []                                                                                                                                             
contexts_list : []                                                                                                                                             
contexts_list : []
Done!
exp-zh/test_wavs/3.wav
文森特卡所是全球知名的法国性格派演员
----------
exp-zh/test_wavs/4.wav
蒋友伯被拍到带着女儿出游
----------
exp-zh/test_wavs/5.wav
周望军就落实控物价
----------
exp-zh/test_wavs/6.wav
朱立南在上市见面会上表示
----------
num_threads: 1
decoding_method: modified_beam_search
Wave duration: 20.328 s
Elapsed time: 2.806 s
Real time factor (RTF): 2.806/20.328 = 0.138
```

Command with contexts:
```bash
python python-api-examples/offline-decode-files.py \
  --tokens exp-zh/tokens.txt \
  --modeling-unit char \
  --encoder exp-zh/encoder-epoch-99-avg-1.onnx \
  --decoder exp-zh/decoder-epoch-99-avg-1.onnx \
  --joiner exp-zh/joiner-epoch-99-avg-1.onnx \
  --decoding-method modified_beam_search  \
  --contexts "文森特卡索/蒋友柏/朱立南/周望君" \
  exp-zh/test_wavs/3.wav exp-zh/test_wavs/4.wav exp-zh/test_wavs/5.wav exp-zh/test_wavs/6.wav
```

The results:
```
Started!                                                                                                                                                       
contexts_list : [[1875, 2141, 401, 451, 1336], [1093, 944, 3381], [2167, 361, 919], [460, 387, 2188]]                                                          
contexts_list : [[1875, 2141, 401, 451, 1336], [1093, 944, 3381], [2167, 361, 919], [460, 387, 2188]]                                                          
contexts_list : [[1875, 2141, 401, 451, 1336], [1093, 944, 3381], [2167, 361, 919], [460, 387, 2188]]                                                          
contexts_list : [[1875, 2141, 401, 451, 1336], [1093, 944, 3381], [2167, 361, 919], [460, 387, 2188]]                                                          
Done!                                                                                                                                                          
exp-zh/test_wavs/3.wav                                                                                                                                         
文森特卡索是全球知名的法国性格派演员                                                                                                                           
----------                                                                                                                                                     
exp-zh/test_wavs/4.wav                                                                                                                                         
蒋友伯被拍到带着女儿出游                                                                                                                                       
----------                                                                                                                                                     
exp-zh/test_wavs/5.wav                                                                                                                                         
周望君就落实空物价                                                                                                                                             
----------                                                                                                                                                     
exp-zh/test_wavs/6.wav                                                                                                                                         
朱立南在上市见面会上表示                                                                                                                                       
----------                                                                                                                                                     
num_threads: 1                                                                                                                                                 
decoding_method: modified_beam_search                                                                                                                          
Wave duration: 20.328 s                                                                                                                                        
Elapsed time: 2.655 s                                                                                                                                          
Real time factor (RTF): 2.655/20.328 = 0.131 
```

pkufool:
@csukuangfj  I tested the correctness of this feature, but I might be too busy to implement it for other apis and server engine, so could you give a review for this and let's merge it first, then some other guys can continue to wrap it to other language.

python-api-examples/offline-decode-files.py
```diff
@@ -153,6 +194,23 @@ def assert_file_exists(filename: str):
...
+    sp = None
+    if "bpe" in args.modeling_unit:
+        assert_file_exists(args.bpe_model)
+        sp = spm.SentencePieceProcessor()
+        sp.load(args.bpe_model)
+    tokens = {}
+    with open(args.tokens, "r") as f:
+        for line in f.readlines():
+            toks = line.strip().split()
+            assert len(toks) == 2, len(toks)
```
csukuangfj, line 208: please also add
```python3
assert toks[0] not in tokens, f"Duplicate token: {toks} "
```

python-api-examples/offline-decode-files.py
```diff
@@ -239,7 +304,13 @@ def main():
         duration = len(samples) / sample_rate
         total_duration += duration
 
-        s = recognizer.create_stream()
+        print(f"contexts_list : {contexts_list}")
```
csukuangfj: Please move it into the statement at line 308.

sherpa-onnx/csrc/context-graph-test.cc
```diff
@@ -0,0 +1,48 @@
...
+namespace sherpa_onnx {
+
+TEST(ContextGraph, TestBasic) {
+  std::vector<std::string> contexts_str(
+      {"S", "HE", "SHE", "SHELL", "HIS", "HERS", "HELLO", "THIS", "THEM"});
+  std::vector<std::vector<int32_t>> contexts;
+  for (int32_t i = 0; i < contexts_str.size(); ++i) {
+    std::vector<int32_t> tmp;
+    for (int32_t j = 0; j < contexts_str[i].size(); ++j) {
+      tmp.push_back((int32_t)contexts_str[i][j]);
```
csukuangfj:
```suggestion
      tmp.push_back(contexts_str[i][j]);
```

sherpa-onnx/csrc/context-graph-test.cc
```diff
@@ -0,0 +1,48 @@
...
+    }
+    contexts.push_back(tmp);
```
csukuangfj:
```suggestion
    contexts.push_back(std::move(tmp));
```

sherpa-onnx/csrc/context-graph-test.cc
```diff
@@ -0,0 +1,48 @@
...
+  }
+  auto context_graph = ContextGraph(1);
+  context_graph.Build(contexts);
+
+  auto queries = std::map<std::string, float>{
+      {"HEHERSHE", 14}, {"HERSHE", 12}, {"HISHE", 9},   {"SHED", 6},
+      {"HELL", 2},      {"HELLO", 7},   {"DHRHISQ", 4}, {"THEN", 2}};
+
+  for (auto iter : queries) {
```
csukuangfj:
```suggestion
  for (const auto &iter : queries) {
```

sherpa-onnx/csrc/context-graph-test.cc
```diff
@@ -0,0 +1,48 @@
...
+    float total_scores = 0;
+    auto state = context_graph.Root();
+    for (auto q : iter.first) {
+      auto res = context_graph.ForwardOneStep(state, (int32_t)q);
```
csukuangfj:
```suggestion
      auto res = context_graph.ForwardOneStep(state, q);
```

sherpa-onnx/csrc/context-graph.h
```diff
@@ -0,0 +1,64 @@
...
+        local_node_score(local_node_score),
+        is_end(is_end) {}
+};
+
+class ContextGraph {
+ public:
+  ContextGraph() = default;
+  explicit ContextGraph(float context_score) : context_score_(context_score) {
+    root_ = std::make_shared<ContextState>(-1, 0, 0, 0, false);
+    root_->fail = root_;
```
csukuangfj: This creates a cycle and prevents `root_` from being freed, i.e., causing memory leaks.

csukuangfj: Turns out there are many cycles in the graph.

To avoid memory leak, please replace
```
  ContextStatePtr fail;
```
with
```
  ContextState *fail = nullptr;
```

sherpa-onnx/csrc/context-graph.h
```diff
@@ -0,0 +1,64 @@
...
+  }
+  ~ContextGraph() {}
```
csukuangfj:
```suggestion
  ~ContextGraph() = default
```
or just delete this line.

sherpa-onnx/csrc/context-graph.h
```diff
@@ -0,0 +1,64 @@
...
+  void Build(const std::vector<std::vector<int32_t>> &token_ids);
```
csukuangfj:
```suggestion
  void Build(const std::vector<std::vector<int32_t>> &token_ids) const;
```

sherpa-onnx/csrc/context-graph.h
```diff
@@ -0,0 +1,64 @@
...
+                                                   int32_t token_id) const;
+  std::pair<float, ContextStatePtr> Finalize(
+      const ContextStatePtr &state) const;
+
+  ContextStatePtr Root() const { return root_; }
+
+ private:
+  float context_score_;
+  ContextStatePtr root_;
+  void FillFailOutput();
```
csukuangfj:
```suggestion
  void FillFailOutput() const;
```

sherpa-onnx/csrc/context-graph.cc
```diff
@@ -0,0 +1,103 @@
...
+std::pair<float, ContextStatePtr> ContextGraph::Finalize(
+    const ContextStatePtr &state) const {
+  float score = -state->node_score;
+  if (state->is_end) {
+    score = 0;
+  }
+  return std::make_pair(score, root_);
+}
+
+void ContextGraph::FillFailOutput() {
```
csukuangfj:
```suggestion
void ContextGraph::FillFailOutput() const {
```

sherpa-onnx/csrc/context-graph.cc
```diff
@@ -0,0 +1,103 @@
...
+// Copyright (c)  2023  Xiaomi Corporation
+
+#include "sherpa-onnx/csrc/context-graph.h"
+
+#include <cassert>
+#include <queue>
+#include <utility>
+
+namespace sherpa_onnx {
+void ContextGraph::Build(const std::vector<std::vector<int32_t>> &token_ids) {
```
csukuangfj:
```suggestion
void ContextGraph::Build(const std::vector<std::vector<int32_t>> &token_ids) const {
```

sherpa-onnx/csrc/context-graph.cc
```diff
@@ -0,0 +1,103 @@
...
+  float score = -state->node_score;
+  if (state->is_end) {
+    score = 0;
+  }
+  return std::make_pair(score, root_);
+}
+
+void ContextGraph::FillFailOutput() {
+  std::queue<ContextStatePtr> node_queue;
+  for (auto kv : root_->next) {
```
csukuangfj:
```suggestion
  for (auto &kv : root_->next) {
```

sherpa-onnx/csrc/context-graph.h
csukuangfj: I think the same goes for
```cpp
ContextStatePtr output;
```

sherpa-onnx/csrc/hypothesis.h
```diff
@@ -39,11 +40,18 @@ struct Hypothesis {
...
+  // the minimum of tokens in a chunk for streaming RNN LM
+  int32_t lm_rescore_min_chunk = 2;  // a const
+
   int32_t num_trailing_blanks = 0;
 
   Hypothesis() = default;
-  Hypothesis(const std::vector<int64_t> &ys, double log_prob)
-      : ys(ys), log_prob(log_prob) {}
+  Hypothesis(const std::vector<int64_t> &ys, double log_prob,
+             const ContextStatePtr &context_state = nullptr)
```
csukuangfj:
```suggestion
            ContextStatePtr context_state = nullptr)
```

sherpa-onnx/csrc/offline-recognizer-transducer-impl.h
```diff
@@ -72,6 +73,16 @@ class OfflineRecognizerTransducerImpl : public OfflineRecognizerImpl {
     }
   }
 
+  std::unique_ptr<OfflineStream> CreateStream(
+      const std::vector<std::vector<int32_t>> &context_list) const override {
+    // We create context_graph at this level, because we might have default
+    // context_graph(will add later if need) that belong to the whole model
```
csukuangfj:
```suggestion
    // context_graph(will be added later if needed) that belongs to the whole model
```

sherpa-onnx/csrc/context-graph.h
csukuangfj: Are users able to call `Build()` multiple times?

sherpa-onnx/csrc/offline-recognizer-transducer-impl.h
```diff
@@ -117,7 +128,8 @@ class OfflineRecognizerTransducerImpl : public OfflineRecognizerImpl {
                                -23.025850929940457f);
 
     auto t = model_->RunEncoder(std::move(x), std::move(x_length));
-    auto results = decoder_->Decode(std::move(t.first), std::move(t.second));
+    auto results =
+        decoder_->Decode(std::move(t.first), std::move(t.second), ss, n);
```
csukuangfj, line 132: I suggest we pass a nullptr instead of ss when context graph is empty. Otherwise, it does extra work in modified beam search.

sherpa-onnx/python/csrc/offline-recognizer.cc
```diff
@@ -35,7 +36,14 @@ void PybindOfflineRecognizer(py::module *m) {
   using PyClass = OfflineRecognizer;
   py::class_<PyClass>(*m, "OfflineRecognizer")
       .def(py::init<const OfflineRecognizerConfig &>(), py::arg("config"))
-      .def("create_stream", &PyClass::CreateStream)
+      .def("create_stream", [](PyClass &self) { return self.CreateStream(); })
```
csukuangfj:
```suggestion
      .def("create_stream", [](const PyClass &self) { return self.CreateStream(); })
```

sherpa-onnx/python/csrc/offline-recognizer.cc
```diff
@@ -35,7 +36,14 @@ void PybindOfflineRecognizer(py::module *m) {
...
+      .def(
+          "create_stream",
+          [](PyClass &self,
```
csukuangfj, line 43:
```suggestion
          [](const PyClass &self,
```

sherpa-onnx/python/sherpa_onnx/__init__.py
```diff
@@ -1,5 +1,78 @@
+from typing import Dict, List, Optional
+
 from _sherpa_onnx import Display
 
 from .online_recognizer import OnlineRecognizer
 from .online_recognizer import OnlineStream
 from .offline_recognizer import OfflineRecognizer
+
+
+def encode_contexts(
```
csukuangfj: Could you move this funciton into a separate file and import it in `__init__.py`?

I suggest we keep the `__init__.py` file as simple as possible.

sherpa-onnx/python/sherpa_onnx/__init__.py
```diff
@@ -1,5 +1,78 @@
...
+    """
+    Encode the given contexts (a list of string) to a list of a list of token ids.
+
+    Args:
+      modeling_unit:
+        The valid values are bpe, char, bpe+char
+      contexts:
+        The given contexts list (a list of string).
+      sp:
+        A instance of SentencePieceProcessor.
```
csukuangfj:
```suggestion
        An instance of SentencePieceProcessor.
```

sherpa-onnx/csrc/context-graph-test.cc
```diff
@@ -19,9 +19,9 @@ TEST(ContextGraph, TestBasic) {
   for (int32_t i = 0; i < contexts_str.size(); ++i) {
     std::vector<int32_t> tmp;
     for (int32_t j = 0; j < contexts_str[i].size(); ++j) {
-      tmp.push_back((int32_t)contexts_str[i][j]);
+      tmp.push_back(contexts_str[i][j]);
     }
-    contexts.push_back(tmp);
+    contexts.emplace_back(tmp);
```
csukuangfj:
```suggestion
    contexts.emplace_back(std::move(tmp));
```

sherpa-onnx/csrc/offline-recognizer-transducer-impl.h
pkufool: I think pass ss is OK, because some streams would have context graph, some streams would not, we will need a loop to get the context graph for each stream anyway.

sherpa-onnx/csrc/context-graph.cc
```diff
@@ -10,6 +10,7 @@
 
 namespace sherpa_onnx {
 void ContextGraph::Build(const std::vector<std::vector<int32_t>> &token_ids) {
+  assert(false == is_populated_);
```
csukuangfj: I suggest that you use `SHERPA_ONNX_CHECK()` or `SHERPA_ONNX_DCHECK()` to replace `assert()`.
It will give you a more informative message on assertion failure.

sherpa-onnx/csrc/context-graph.h
pkufool: No, I add a flag to guarantee that it can be called only once.

sherpa-onnx/csrc/context-graph.h
pkufool: Don't want to use the raw pointer, I use a weak_ptr to fix this.

sherpa-onnx/csrc/context-graph-test.cc
```diff
@@ -0,0 +1,48 @@
...
+TEST(ContextGraph, TestBasic) {
+  std::vector<std::string> contexts_str(
+      {"S", "HE", "SHE", "SHELL", "HIS", "HERS", "HELLO", "THIS", "THEM"});
+  std::vector<std::vector<int32_t>> contexts;
+  for (int32_t i = 0; i < contexts_str.size(); ++i) {
+    std::vector<int32_t> tmp;
+    for (int32_t j = 0; j < contexts_str[i].size(); ++j) {
+      tmp.push_back(contexts_str[i][j]);
+    }
+    contexts.push_back(std::move(tmp));
```
csukuangfj:
```suggestion
contexts.emplace_back(context_str[i].begin(), contexts_str[i].end());
```
is much simpler.

.github/workflows/run-python-test.yaml
```diff
@@ -54,7 +54,7 @@ jobs:
       - name: Install Python dependencies
         shell: bash
         run: |
-          python3 -m pip install --upgrade pip numpy
+          python3 -m pip install --upgrade pip numpy sentencepiece==0.1.96
```
csukuangfj, line 57: Please also change
https://github.com/k2-fsa/sherpa-onnx/blob/1a1b9fd2362fa00910e7461d7cfb58f83b024a33/setup.py#L38-L40

python-api-examples/offline-decode-files.py
```diff
@@ -153,6 +194,24 @@ def assert_file_exists(filename: str):
...
 
+def encode_contexts(args, contexts: List[str]) -> List[List[int]]:
+    sp = None
+    if "bpe" in args.modeling_unit:
+        assert_file_exists(args.bpe_model)
+        sp = spm.SentencePieceProcessor()
+        sp.load(args.bpe_model)
+    tokens = {}
+    with open(args.tokens, "r") as f:
+        for line in f.readlines():
```
csukuangfj:
```suggestion
        for line in f:
```

python-api-examples/offline-decode-files.py
```diff
@@ -182,10 +241,17 @@ def main():
...
     assert args.num_threads > 0, args.num_threads
+
+    contexts_list = []
     if args.encoder:
         assert len(args.paraformer) == 0, args.paraformer
         assert len(args.nemo_ctc) == 0, args.nemo_ctc
 
+        contexts = [x.strip().upper() for x in args.contexts.split("/") if x.strip()]
+        if contexts:
+            print (f"Contexts list: {contexts}")
```
csukuangfj:
```suggestion
            print(f"Contexts list: {contexts}")
```

sherpa-onnx/csrc/context-graph.cc
csukuangfj: 这个可以改成 const 不？

sherpa-onnx/csrc/context-graph.h
csukuangfj: I suggest that you use `std::unique_ptr<>`.

Here are the reasons:

The pointers in the map `next`  are allocated in the `Build()` method. Once allocated they will not be deallocated as long as `root_` is alive. 

So `std::unique_ptr<>` is a perfect choice for it. It is smaller in size than `std::shared_ptr<>`. Also, when you copy a shared pointer in `ForwardOneStep()` and `FillFailOutput()`, you need to pay for increasing/decreasing an atomic counter.

Also, you don't need to use `weak_ptr` once switching to `unique_ptr`.


Is there a reason to choose `std::shared_ptr` instead of `std::unique_ptr`?

If you want to return a ContextState to a stream, you only need to return the underlying raw pointer managed by the unique ptr.
Since the `root_` is kept as a member of the stream, so you don't need to worry about dangling pointers.

python-api-examples/offline-decode-files.py
```diff
@@ -153,6 +194,24 @@ def assert_file_exists(filename: str):
...
 
 
+def encode_contexts(args, contexts: List[str]) -> List[List[int]]:
+    sp = None
+    if "bpe" in args.modeling_unit:
+        assert_file_exists(args.bpe_model)
+        sp = spm.SentencePieceProcessor()
+        sp.load(args.bpe_model)
+    tokens = {}
+    with open(args.tokens, "r") as f:
```
csukuangfj:
```suggestion
    with open(args.tokens, "r", encoding="utf-8") as f:
```

sherpa-onnx/python/sherpa_onnx/utils.py
```diff
@@ -0,0 +1,72 @@
...
+    if "bpe" in modeling_unit:
+        assert sp is not None
+    if "char" in modeling_unit:
+        assert tokens_table is not None
+        assert len(tokens_table) > 0, len(tokens_table)
+
+    if "char" == modeling_unit:
+        for context in contexts:
+            ids = [
+                tokens_table[txt] if txt in tokens_table else tokens_table["<unk>"]
```
csukuangfj, line 37: Shall we consider the case when `txt` is a space?

python-api-examples/offline-decode-files.py
pkufool: Thanks for the detail explanation, will change to `unique_ptr`, I also find some documents here in case someone else is intrested in it.

http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-summary-smartptrs

The items in red boxes are relevant to this comment, other items are also very helpful.

image

sherpa-onnx/python/sherpa_onnx/utils.py
pkufool: I think for `bpe` and `char` based models, there will not be spaces in context phrases, no?

csukuangfj: Is "hello world" a valid phrase in this case? 

If not, could you give an example?

pkufool: I think "hello world" will be tokenized to something like "hel, lo, \_wor, ld" if using bpe, space will be represented as "\_" (a unicode version underline).

csukuangfj: what if the modelling unit is `char`?

pkufool: For a Chinese phrase, there is no spaces at all.

csukuangfj: I mean for non-CJK languages, e.g., English, German, French, etc.

sherpa-onnx/csrc/context-graph.cc
pkufool: > 这个可以改成 const 不？

It modifies `is_populated_`.

sherpa-onnx/python/sherpa_onnx/utils.py
pkufool: OK, I think there is no need to support all the languages, modeling units for now, the fact is we only have bpe models for English, char models for Chinese and bpe+char models for English+Chinese, for all these models we don't need a space. We could surely add `space` if some models need it, this is just a util function that converts phrases to token ids, I'd rather keep it as simple as it is, we can modify it any time when needed, won't affect anything.

csukuangfj: Then please document that
- char is for CJK languages, not for English like languages
- bpe is for English-like languages
- bpe+char is for CJK + English-like languages.

sherpa-onnx/csrc/offline-recognizer-transducer-impl.h
```diff
@@ -72,6 +73,16 @@ class OfflineRecognizerTransducerImpl : public OfflineRecognizerImpl {
     }
   }
 
+  std::unique_ptr<OfflineStream> CreateStream(
+      const std::vector<std::vector<int32_t>> &context_list) const override {
+    // We create context_graph at this level, because we might have default
+    // context_graph(will be added later if needed) that belongs to the whole
+    // model rather than each stream.
+    auto context_graph = std::make_shared<ContextGraph>(config_.context_score);
+    context_graph->Build(context_list);
```
csukuangfj: Could you pass `context_list` as an argument to the constructor of ContextGraph and make `Build()` a private method?

sherpa-onnx/python/sherpa_onnx/utils.py
pkufool: Ok, I think documenting "char is for CJK language" is enough, for bpe system we don't care the languages.

sherpa-onnx/csrc/offline-recognizer-transducer-impl.h
pkufool: Yet, might be a good idea.

sherpa-onnx/python/sherpa_onnx/utils.py
csukuangfj: Actually, if you handle spaces, then `char` also works for non-CJK languages.

pkufool: ~~Ok,  I will map `space` to `blank`.~~

Wait,  I was thinking can we just map `space` to `blank`,  is there a special token for `space` in non-CJK languages char modeling units? 

So, I'd rather not adding `space` for now, let's add it when needed.

sherpa-onnx/csrc/offline-recognizer-transducer-impl.h
csukuangfj: in that case, I suggest putting the loop outside of `Decode()`.
Otherwise, you need to check for each frame inside `Decode()`.

If there are no streams requiring that, we don't need to do extra work for each frame inside `Decode()`.

sherpa-onnx/csrc/offline-transducer-modified-beam-search-decoder.cc
```diff
@@ -17,23 +18,39 @@ namespace sherpa_onnx {
...
-    Ort::Value encoder_out, Ort::Value encoder_out_length) {
+    Ort::Value encoder_out, Ort::Value encoder_out_length,
+    OfflineStream **ss /*=nullptr */, int32_t n /*= 0*/) {
   PackedSequence packed_encoder_out = PackPaddedSequence(
       model_->Allocator(), &encoder_out, &encoder_out_length);
 
   int32_t batch_size =
       static_cast<int32_t>(packed_encoder_out.sorted_indexes.size());
 
+  if (ss != nullptr) assert(batch_size == n);
```
csukuangfj: Please replace `assert` with `SHERPA_ONNX_CHECK_EQ()`.

sherpa-onnx/csrc/offline-recognizer.cc
```diff
@@ -26,6 +26,9 @@ void OfflineRecognizerConfig::Register(ParseOptions *po) {
 
   po->Register("max-active-paths", &max_active_paths,
                "Used only when decoding_method is modified_beam_search");
+  po->Register("context-score", &context_score,
+               "The bonus score for each token in context word/phrase. "
+               "Used only when decoding_method is modified_beam_search");
```
csukuangfj, line 31: Please also update line 55.

sherpa-onnx/python/sherpa_onnx/utils.py
csukuangfj: If you are not willing to make a change, please add
```python3
assert ' ' not in context
```
Otherwise, users will get undesired results without notice
if their modelling unit is `character` and give a phrase
`hello world`.

csukuangfj (approved):

sherpa-onnx/csrc/offline-recognizer-transducer-impl.h
pkufool: I just revisited the code, it seems make no difference to add a extra loop outside of `Decode()`,  the loop inside `Decode()` is for the whole sequence not each frame.  Shall we keep it as it is?

csukuangfj: Ok, but for the online stream, I suggest putting it outside of `Decode()`.

sherpa-onnx/csrc/context-graph.h
```diff
@@ -0,0 +1,66 @@
...
+  ContextState(int32_t token, float token_score, float node_score,
+               float local_node_score, bool is_end)
+      : token(token),
+        token_score(token_score),
+        node_score(node_score),
+        local_node_score(local_node_score),
+        is_end(is_end) {}
+};
+
+class ContextGraph {
```
danpovey, line 40: @pkufool I don't know if this is the right place, but I think we should have documentation of what is being done here.  My understanding is that you have implemented context biasing with Aho Corasick, but I don't know exactly what is being done by that algorithm.  I imagine it is somehow boosting the probabilities of strings that appear in the supplied context-biasing?  Do they have to be the complete strings, or will partial ones match?  And at what point is the boost applied?  What happens when different matches are present, that might overlap?

PR #151, opened 2023-05-12
Support CoreML for macOS

csukuangfj:


csukuangfj:
From https://github.com/microsoft/onnxruntime/issues/11343#issuecomment-1108935687
> For your information, the macOS binaries in our github release package were built in a way like:
>
> ./build.sh  --build --build_java --build_nodejs --use_coreml --cmake_extra_defines CMAKE_OSX_ARCHITECTURES=arm64  --parallel --build_shared_lib --config Release

The downloaded onnxruntime lib already supports CoreML.
https://github.com/k2-fsa/sherpa-onnx/blob/de1880948b44d25437510c15d52cef1a32bc0e6e/cmake/onnxruntime.cmake#L45

PR #153, opened 2023-05-12
Add onnxruntime gpu for cmake

yuekaizhang:
Resolved #138


sherpa-onnx/csrc/session.cc
```diff
@@ -28,15 +28,19 @@ static Ort::SessionOptions GetSessionOptionsImpl(int32_t num_threads,
...
-          // The CUDA provider is available, proceed with setting the options
-          OrtCUDAProviderOptions options;
-          options.device_id = 0;
-          // set more options on need
-          sess_opts.AppendExecutionProvider_CUDA(options);
+      std::vector<std::string> available_providers =
+          Ort::GetAvailableProviders();
+      if (std::find(available_providers.begin(), available_providers.end(),
+                    "CUDAExecutionProvider") != available_providers.end()) {
+        // The CUDA provider is available, proceed with setting the options
```
csukuangfj, line 36: what if we don't check it?

yuekaizhang: Then it would be a segmentation error. If we use cpu build but --provider=gpu.

csukuangfj:
Thanks! Is it ready to merge?

yuekaizhang:
> Thanks! Is it ready to merge?

Yes. It's ready.

PR #154, opened 2023-05-12
Support DirectML for windows

csukuangfj:
Note:  It does not seem to improve the speed when passing `--provider=directml` from the commandline.

### DIFF
--- a/cmake/onnxruntime.cmake
+++ b/cmake/onnxruntime.cmake
@@ -1,6 +1,7 @@
 function(download_onnxruntime)
   include(FetchContent)
 
+  set(LIB_PATH)
   if(CMAKE_SYSTEM_PROCESSOR STREQUAL aarch64)
     # For embedded systems
     set(possible_file_locations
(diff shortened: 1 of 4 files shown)


thewh1teagle:
> Note: It does not seem to improve the speed when passing `--provider=directml` from the commandline.

What's something heavy I can test this on? I wanted to test on whisper medium but this branch does not have it.
Can I use sherpa-onnx.exe binary to test something?

csukuangfj:
I think we should resolve the conflict and merge this PR.

If I recall correctly, it has not been merged because it does not run faster with DirectlML.

thewh1teagle:
> it has not been merged because it does not run faster with DirectlML.

I would like to check it and compare

thewh1teagle:
> If I recall correctly, it has not been merged because it does not run faster with DirectlML.

I tried on another c++ project that uses DirectML. it used 95% from the TPU (amd ryzen 5 4500u) 
But it wasn't fast. Maybe it does works slow.
The fastest way to use whisper I found is [CTranslate2](https://github.com/OpenNMT/CTranslate2) and even on CPU only!
Do you have idea how can we get similar results with sherpa?

thewh1teagle:
Any chance you could resolve the conflicts so I can test it with Whisper medium?

csukuangfj:
Closing via #1153

PR #162, opened 2023-05-25
Added provider option to sherpa-onnx and decode-file-c-api

jingzhaoou:


csukuangfj:
Thanks! The number of command line arguments is getting larger. Could you please follow sherpa-onnx-offline.cc to use ParseOptions?

As for the C code, could you use getopt?

jingzhaoou:
@csukuangfj  I updated my code to use `ParseOptions` in `sherpa-onnx.cc`. For `decode-file-c-api.cc`, I used `getopt_long()`, which works fine on my Ubuntu and MacBook. However, the build failed on Windows as Visual C does not provide `getopt.h`. Any suggestions how I should resolve it? Providing my own `getopt.h/getopt.cc`? Thanks.

jingzhaoou:
@csukuangfj should we consider https://likle.github.io/cargs? Or simply leave the C code as before?

sherpa-onnx/csrc/sherpa-onnx.cc
```diff
@@ -71,63 +62,68 @@ for a list of pre-trained models to download.
...
 
-  fprintf(stderr, "wav filename: %s\n", wav_filename.c_str());
-  fprintf(stderr, "wav duration (s): %.3f\n", duration);
+    fprintf(stderr, "wav filename: %s\n", wav_filename.c_str());
+    fprintf(stderr, "wav duration (s): %.3f\n", duration);
 
-  auto begin = std::chrono::steady_clock::now();
-  fprintf(stderr, "Started\n");
+    auto begin = std::chrono::steady_clock::now();
+    fprintf(stderr, "Started\n");
```
csukuangfj, line 84: Could you put this line before line 85. We may not want to time the I/O operation.
(It is a mistake for the original code.)

sherpa-onnx/csrc/sherpa-onnx.cc
```diff
@@ -71,63 +62,68 @@ for a list of pre-trained models to download.
...
 
-  // Call InputFinished() to indicate that no audio samples are available
-  s->InputFinished();
+    // Call InputFinished() to indicate that no audio samples are available
+    s->InputFinished();
 
-  while (recognizer.IsReady(s.get())) {
-    recognizer.DecodeStream(s.get());
-  }
+    while (recognizer.IsReady(s.get())) {
```
csukuangfj, line 98: Could you follow
https://github.com/k2-fsa/sherpa-onnx/blob/5e2dc5ceea21145b205c3d712c38dff1e5d81d2b/python-api-examples/online-decode-files.py#L158
to decode multiple files in batches?

Also, could you call `fprintf` at the end so that I/O operations such as `fprintf` are not timed.

csukuangfj:
> @csukuangfj should we consider https://likle.github.io/cargs? Or simply leave the C code as before?

`cargs` looks fine to me.

csukuangfj (changes requested):

c-api-examples/decode-file-c-api.c
```diff
@@ -5,23 +5,70 @@
 // This file shows how to use sherpa-onnx C API
 // to decode a file.
 
-#include <getopt.h>
+#include <cargs.h>
```
csukuangfj:
```suggestion
#include "cargs.h"
```

Please use `""` for non-system headers.

(You may need to sort the headers alphabetically after changing `<` to `"`)

cmake/cargs.cmake
```diff
@@ -0,0 +1,18 @@
...
+  FetchContent_Declare(cargs URL ${cargs_URL} )
+
+  FetchContent_GetProperties(cargs)
+  if(NOT cargs_POPULATED)
+    message(STATUS "Downloading cargs ${cargs_URL}")
+    FetchContent_Populate(cargs)
+  endif()
+  message(STATUS "cargs is downloaded to ${cargs_SOURCE_DIR}")
+  add_subdirectory(${cargs_SOURCE_DIR} ${cargs_BINARY_DIR} EXCLUDE_FROM_ALL)
+  # include_directories(${asio_SOURCE_DIR}/asio/include)
```
csukuangfj:
```suggestion
```

cmake/cargs.cmake
```diff
@@ -0,0 +1,18 @@
+function(download_cargs)
+  include(FetchContent)
+
+  set(cargs_URL "https://github.com/likle/cargs/archive/refs/tags/v1.0.3.tar.gz")
+
+  FetchContent_Declare(cargs URL ${cargs_URL} )
```
csukuangfj:
```suggestion
  FetchContent_Declare(cargs URL ${cargs_URL})
```

c-api-examples/decode-file-c-api.c
```diff
@@ -60,13 +95,33 @@ int32_t main(int32_t argc, char *argv[]) {
...
+    identifier = cag_option_get(&context);
+    value = cag_option_get_value(&context);
+    switch (identifier) {
+      case 't': config.model_config.tokens = value; break;
+      case 'e': config.model_config.encoder = value; break;
+      case 'd': config.model_config.decoder = value; break;
+      case 'j': config.model_config.joiner = value; break;
+      case 'n': config.model_config.num_threads = atoi(value); break;
+      case 'p': config.model_config.provider = value; break;
+      case 'm': config.decoding_method = value; break;
```
csukuangfj, line 114: Shall we add a `default` statement to handle unknown options?

jingzhaoou: All the values are set up their defaults before reaching this `while`. So, we don't need a `default` statement. I can add a `default` statement that does nothing. Thanks.

sherpa-onnx/csrc/sherpa-onnx.cc
jingzhaoou: Changed the code as suggested. Thanks, @csukuangfj .

c-api-examples/decode-file-c-api.c
jingzhaoou: Added a `default` statement that does nothing. :-)

sherpa-onnx/csrc/sherpa-onnx.cc
jingzhaoou: @csukuangfj I adjusted the `fprintf` statements so that they are not timed. 

Regarding the decoding of multiple files in batches, I can work on that for sure. But that is a bit beyond the original scope of this PR. Can I work on that after the current PR is merged? Thanks.

csukuangfj: Sure.

csukuangfj:
Thanks!

PR #168, opened 2023-06-05
rnnlm model inference supports num_threads setting

keanucui:
It's support passing num_threads argument to rnnlm onnx model, including offline rnnlm rescore and online rnnlm rescore.


keanucui:
@csukuangfj I changed asio.cmake.  I'll restore it when it's ready to merge.

PR #185, opened 2023-06-25
support streaming zipformer2

kakashidan:
It works but the code is not clean enough. I will refactor the code later

### DIFF
--- a/sherpa-onnx/csrc/online-zipformer2-transducer-model.cc
+++ b/sherpa-onnx/csrc/online-zipformer2-transducer-model.cc
@@ -1,0 +1,460 @@
+// sherpa-onnx/csrc/online-zipformer2-transducer-model.cc
+//
+// Copyright (c)  2023  Xiaomi Corporation
+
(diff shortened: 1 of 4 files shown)


sherpa-onnx/csrc/online-zipformer2-transducer-model.cc
```diff
@@ -0,0 +1,460 @@
+// sherpa-onnx/csrc/online-zipformer2-transducer-model.cc
+//
+// Copyright (c)  2023  Xiaomi Corporation
```
csukuangfj, line 3: Please also add your own information here.

csukuangfj:
Let us merge it first. You can fix any issues in a separate pull-request.

sherpa-onnx/csrc/online-zipformer2-transducer-model.cc
kakashidan: ok. thanks

PR #186, opened 2023-06-26
Add detection conditions under the x86 platform.

neuxys:
When downloading onnxruntime, add detection conditions under the x86 platform.


csukuangfj:
Thanks for your contribution!

PR #203, opened 2023-07-08
Publish pre-compiled xcframework for iOS via GitHub actions

csukuangfj:


csukuangfj:
The pre-compiled xcframework can be found at
https://huggingface.co/csukuangfj/sherpa-onnx-libs/tree/main

PR #210, opened 2023-07-12
Use static libraries for MFC examples

csukuangfj:
[StreamingSpeechRecognition.exe.zip](https://github.com/k2-fsa/sherpa-onnx/files/12026971/StreamingSpeechRecognition.exe.zip)

I've attached a pre-compiled `exe` above.  It is self-contained and has no external dependencies.
(Note: You can only run it on Windows x64 for now. Will add support for 32-bit Windows)

You only need to download the pre-trained models.

---

[Updated]: I have updated the attached exe.


csukuangfj:
By the way, the pre-compiled executable is also available at the following address

https://github.com/k2-fsa/sherpa-onnx/releases


Screenshot 2023-07-13 at 16 01 14

Issue #211, opened 2023-07-13
Slow Model Initialization


w11wo:
Hi, I've just been playing with sherpa-onnx and I find that the model initialization is kinda slow compared to sherpa-ncnn. I noticed this in different desktop OS's (tested on macOS, Linux, and Windows). Oddly enough, I didn't notice a slow initialization on my iOS device. I wonder if we could somehow speed it up?

A minimal reproducible example is as follows, in Python.

```py
import sherpa_onnx
from pathlib import Path

model_path = Path("/root/sherpa-onnx-streaming-zipformer-en-2023-06-26")
recognizer = sherpa_onnx.OnlineRecognizer(
    tokens=str(model_path / "tokens.txt"),
    encoder=str(model_path / "encoder-epoch-99-avg-1-chunk-16-left-64.onnx"),
    decoder=str(model_path / "decoder-epoch-99-avg-1-chunk-16-left-64.onnx"),
    joiner=str(model_path / "joiner-epoch-99-avg-1-chunk-16-left-64.onnx"),
    num_threads=1,
    enable_endpoint_detection=False,
    rule1_min_trailing_silence=2.4,
    rule2_min_trailing_silence=1.2,
    rule3_min_utterance_length=30,
    decoding_method="modified_beam_search",
    max_active_paths=4,
    provider="cpu",
)
```

This sample code takes about 4-5 seconds to initialize the model.

The model is [csukuangfj/sherpa-onnx-streaming-zipformer-en-2023-06-26](https://huggingface.co/csukuangfj/sherpa-onnx-streaming-zipformer-en-2023-06-26).

Thanks!

jingzhaoou:
I noticed similar delays on the model initialization. `sherpa-onnx` itself has a very small amount of processing during warm-up. This is probably just an ONNX runtime thing IMO. There might be some graph optimizations, mapping operators to different execution providers, etc. that take time. When I added TensorRT execution provider to a session, the model initialization was much much longer.

w11wo:
Hi @jingzhaoou, thanks for the info. I agree that it might be just an ONNX thing and not a sherpa-onnx thing.

I noticed that the initialization for the int8 quantized model is also slightly faster, so it's probably related to the model graph and the operators. I'm still trying to find out if there's a way to speed it up. Cheers!

csukuangfj:
There is an easy way to reduce the initialization time by half.

[...]

PR #215, opened 2023-07-14
Reduce model initialization time for online speech recognition

w11wo:
Implemented `model_type` parameter for online transducer models as suggested by @csukuangfj in [this issue](https://github.com/k2-fsa/sherpa-onnx/issues/211#issuecomment-1635661879) and based largely on [this PR](https://github.com/k2-fsa/sherpa-onnx/pull/213).

Add a new argument `--model-type` so that it only needs to load the model once.
Otherwise, it needs to load the model twice, where the first loading is to determine the model type.

I have tested [csukuangfj/sherpa-onnx-streaming-zipformer-en-2023-06-26](https://huggingface.co/csukuangfj/sherpa-onnx-streaming-zipformer-en-2023-06-26) via the Python API on Linux, by specifying `model_type="zipformer2"` during initialization.

The model loading time is reduced from ~6 seconds to ~3 seconds for the fp32 model; while for the int8 model the time is reduced from 4.4 seconds to 2.1 seconds.


csukuangfj:
Thanks!

PR #226, opened 2023-07-27
Updated hypothesis key generation to be the same as sherpa

jingzhaoou:
As suggested by the subject. This makes it easier to compare results between `sherpa` and `sherpa-onnx`.

### DIFF
--- a/sherpa-onnx/csrc/hypothesis.h
+++ b/sherpa-onnx/csrc/hypothesis.h
@@ -60,10 +60,10 @@ struct Hypothesis {
   std::string Key() const {
     // TODO(fangjun): Use a hash function?
     std::ostringstream os;
-    std::string sep = "-";
+    std::string sep = "";
     for (auto i : ys) {
-      os << i << sep;
-      sep = "-";
+      os << sep << i;
+      if (sep.empty()) { sep = "-"; }
     }
     return os.str();
   }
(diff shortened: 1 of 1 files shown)


sherpa-onnx/csrc/hypothesis.h
```diff
@@ -60,10 +60,10 @@ struct Hypothesis {
   std::string Key() const {
     // TODO(fangjun): Use a hash function?
     std::ostringstream os;
-    std::string sep = "-";
```
csukuangfj, line 63: please change it to
```
std::string sep;
```
and leave the code in the for loop unchanged.

jingzhaoou: @csukuangfj I changed the code as suggested, which looks like the following

```cpp
  std::string Key() const {
    // TODO(fangjun): Use a hash function?
    std::ostringstream os;
    std::string sep;
    for (auto i : ys) {
      os << i << sep;
      sep = "-";
    }
    return os.str();
  }
```

For `0, 0, 120`, the key is `"00-120-"`. I assume `0-0-120` is the correct join result. Let me know if I miss anything. Thanks.

csukuangfj: Please change
```
os << i << sep;
```
to
```
os << sep << i;
```

Thanks!

jingzhaoou: @csukuangfj  Just changed as suggested. Seem working well. Thanks a lot!

csukuangfj (commented):
Thanks!

PR #230, opened 2023-08-01
Support real time hotwords on python

frankyoujian:
Try supporting real time hot words function on python. Test locally, both Chinese and English hot words work well.

### DIFF
--- a/python-api-examples/speech-recognition-from-microphone.py
+++ b/python-api-examples/speech-recognition-from-microphone.py
@@ -10,6 +10,9 @@ import argparse
 import sys
 from pathlib import Path
 
+from typing import List, Tuple
+import sentencepiece as spm
+
 try:
     import sounddevice as sd
 except ImportError:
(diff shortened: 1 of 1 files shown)


python-api-examples/speech-recognition-from-microphone.py
```diff
@@ -70,6 +73,66 @@ def get_args():
         help="Valid values are greedy_search and modified_beam_search",
     )
 
+    parser.add_argument(
+        "--num-threads",
```
pkufool: You did not use this argument, please delete it.

python-api-examples/speech-recognition-from-microphone.py
```diff
@@ -116,6 +211,9 @@ def main():
             if last_result != result:
                 last_result = result
                 print("\r{}".format(result), end="", flush=True)
+            is_endpoint = recognizer.is_endpoint(stream)
+            if is_endpoint:
+                recognizer.reset(stream)
```
pkufool: I think current implementation did not handle `Reset` for context biasing properly, you have to add some C++ code to `Reset` in `OnlineRecognizer` to make the contexts biasing work after recognizer being reset.  I mean, the context biasing can not work any more after resetting, you can try it.

Here are some reference code piece:
https://github.com/k2-fsa/sherpa-onnx/blob/c88e672733efe4ec6b345d4cdc6c9d65c07495a8/sherpa-onnx/csrc/online-recognizer.cc#L173-L186

pkufool (changes requested):
Thanks!  There are still some problems, they might not relate to your contribution, but it would be good if you can help to fix them.

python-api-examples/speech-recognition-from-microphone.py
frankyoujian: @pkufool  Thanks for explanations. I did not think too much about relations between `Reset` and contexts biasing. Will delete this part.

python-api-examples/speech-recognition-from-microphone.py
frankyoujian: Updated

python-api-examples/speech-recognition-from-microphone.py
pkufool: You can reset here,  I mean there are some issues for current `recognizer.reset()`,  hope that you can help to fix it.

frankyoujian: I will fix this in next PR. Since my current knowledge of context biasing logic is not that solid.

pkufool (approved):

PR #234, opened 2023-08-04
Reinitialize context state after Reset stream when using contexts

frankyoujian:
Reinitialize context state after Reset stream when using contexts, so as to make contexts work again after reset.


pkufool (approved):
@frankyoujian Can you try this on your previous PR, I mean add the "reset logic" you deleted in that PR back and see if contexts is still working after reseting. Thanks!

frankyoujian:
@pkufool I tried locally, it will crash if the python code invokes the old pybind binary which is installed by `pip3 install`. The crash stack is at https://github.com/k2-fsa/sherpa-onnx/blob/master/sherpa-onnx/csrc/context-graph.cc#L34 , since the `state` is NULL after reset while there is no protection in `ContextGraph::ForwardOneStep`.

My thought is, after we release the official pybind binary which means consumers can install new version binary containing this fix by `pip3 install`, then we add the "reset logic" in previous PR.

csukuangfj:
@frankyoujian 

Could you please fix the conflicts?

Changes should be added to
https://github.com/k2-fsa/sherpa-onnx/blob/a324e0cb0d1a47c63aa9edcfb9bac685a9daa40f/sherpa-onnx/csrc/online-recognizer-transducer-impl.h#L227-L237

frankyoujian:
@csukuangfj Done. Reopen a PR due to merge error.

csukuangfj:
> @csukuangfj Done. Reopen a PR due to merge error.

Thanks!

PR #281, opened 2023-08-17
Fix a bug for multilingual ASR

csukuangfj:
We should use
```
logits[:, -1, :]
```
for the `sot_sequence`, not `logits[:, 0, :]`.


csukuangfj:
Thanks to @zzzacwork  for pointing out this bug.

PR #284, opened 2023-08-17
fix issue in upload and offlind-record pages of python streaming server

hexins:
fix issue in upload and offlind-record pages of python streaming server


python-api-examples/http_server.py
```diff
@@ -23,17 +23,20 @@
     ("/js/bootstrap.min.js", "application/javascript"),
     ("/js/bootstrap.min.js.map", "application/javascript"),
     ("/js/jquery-3.6.0.min.js", "application/javascript"),
-    ("/js/offline_record.js", "application/javascript"),
-    ("/js/offline_record.js", "application/javascript"),
+    ("/js/offline_record_streaming.js", "application/javascript"),
```
csukuangfj: Offline and non-streaming are synonymous.

Please either use `offline_record.js` or use `non_streaming_record.js`.

hexins: I think offline record just means we record the voice into one buffer, like a file, 
then we still have streaming and non-streaming mode to transcribe the buffer, 
streaming mode is to send the buffer bytes in stream and receive the result at the same time, 
while non-streaming mode is to send the whole buffer firstly 
and then receive the whole result text in one response, 
so we should still need two different js files for offline record.

PR #316, opened 2023-09-18
Fix LogAdd

kamirdin:
Using 0 as the initial value,  should not perform addition when both values are 0.  It will trigger a bug in the modified beam search algorithm when the language model is not loaded


csukuangfj:

PR #340, opened 2023-09-25
Supports cmake compilation compatible with v3.13.

poor1017:
Solved the problem that the program cannot be compiled with static onnx library. If your cmake version happens to be 3.25, then you may not have this problem, because this version of cmake adds support for `LINUX` variables.


csukuangfj:
Thanks for your contribution!

PR #350, opened 2023-10-05
Proper convolution mode for fast GPU processing

nshmyrev:
Default convolution mode is very slow, as a result sherpa-onnx-gpu is much slower than CPU. With this fix the decoding speed is ok.

See also

https://github.com/microsoft/onnxruntime/issues/12880#issuecomment-1330442188

https://onnxruntime.ai/docs/execution-providers/CUDA-ExecutionProvider.html#cudnn_conv_algo_search


csukuangfj:
Thank you for your first contribution!

PR #388, opened 2023-10-25
include cstdint (debian, gcc-13.2)

rouseabout:
fixes: sherpa-onnx/csrc/text-utils.cc:217:9: error: ‘uint8_t’ does not name a type


csukuangfj:

PR #413, opened 2023-11-07
add --tts-rule-fsts argument at offline-tts.py

longshiming:
add --tts-rule-fsts argument at offline-tts.py


csukuangfj:
Thanks for your contribution!

PR #431, opened 2023-11-17
Judge before UseCachedDecoderOut

HieDean:
When the cached decoder out of the entire batch exists, then use it.

### DIFF
--- a/sherpa-onnx/csrc/online-transducer-greedy-search-decoder.cc
+++ b/sherpa-onnx/csrc/online-transducer-greedy-search-decoder.cc
@@ -89,9 +89,24 @@ void OnlineTransducerGreedySearchDecoder::Decode(
   int32_t num_frames = static_cast<int32_t>(encoder_out_shape[1]);
   int32_t vocab_size = model_->VocabSize();
 
-  Ort::Value decoder_input = model_->BuildDecoderInput(*result);
-  Ort::Value decoder_out = model_->RunDecoder(std::move(decoder_input));
-  UseCachedDecoderOut(*result, &decoder_out);
+  Ort::Value decoder_out{nullptr};
+  bool is_batch_decoder_out_cached = true;
+  for (const auto &r : *result) {
+    if (!r.decoder_out) {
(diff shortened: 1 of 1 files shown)


sherpa-onnx/csrc/online-transducer-greedy-search-decoder.cc
```diff
@@ -89,9 +89,23 @@ void OnlineTransducerGreedySearchDecoder::Decode(
...
+  for (const auto &r : *result) {
+    if (!r.decoder_out) {
+      is_batch_decoder_out_cached = false;
+      break;
+    }
+  }
+  if (is_batch_decoder_out_cached) {
+    auto &r = result->front();
+    std::vector<int64_t> decoder_out_shape = r.decoder_out.GetTensorTypeAndShapeInfo().GetShape();
+    decoder_out = Ort::Value::CreateTensor<float>(model_->Allocator(), decoder_out_shape.data(), decoder_out_shape.size());
```
csukuangfj, line 104: This line only allocates memory for decoder_out but it does not initialize it.

Please use
https://github.com/k2-fsa/sherpa-onnx/blob/eeda1e190e7fb9132442190aa3e10699b486c00e/sherpa-onnx/csrc/cat-test.cc#L126
to concatenate the `r.decoder_out` along the batch dimension.

csukuangfj:
By the way, could you follow
https://github.com/k2-fsa/sherpa-onnx/blob/eeda1e190e7fb9132442190aa3e10699b486c00e/sherpa-onnx/csrc/online-wenet-ctc-model.cc#L74-L75
to replace `Clone()` with `View()` in other code in a separate PR?

sherpa-onnx/csrc/online-transducer-greedy-search-decoder.cc
HieDean: Sorry for my mistake. `Cat()` seems to be a more overloaded option.

csukuangfj:
Thank you for your first contribution!

PR #437, opened 2023-11-20
[Not for merge][Test] For blank skipping in sherpa-onnx test

frankyoujian:


Aurora-6:
Is there any problem in merging the code here？Can these changes be used for ctc blank skip decoding？

frankyoujian:
> Is there any problem in merging the code here？Can these changes be used for ctc blank skip decoding？

@Aurora-6 These changes should be fine, but the program will crash in frame reducer module when run frame_reducer.onnx. Seems there is some tricky issue in frame reducer module.

Aurora-6:
https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/pruned_transducer_stateless7_ctc_bs/export_onnx.py 
So there is a problem with the frame_reducer.onnx exported from this file?

PR #445, opened 2023-11-24
Lock before push_back the deque for thread safety

HieDean:
Lock before push_back the deque for thread safety.


csukuangfj:

PR #460, opened 2023-12-03
Break text into sentences for tts.

csukuangfj:
This is for models that are not using piper-phonemize as their front-end.


GeorgeS2019:
image

why German removed?

csukuangfj:
We are using piper-phonemizer, which uses espeak-ng.

We have German TTS APK.

GeorgeS2019:
I can not find from the list of pretrained models

csukuangfj:
> I can not find from the list of pretrained models

Please see https://github.com/k2-fsa/sherpa-onnx/releases/tag/tts-models

PR #481, opened 2023-12-13
Add two GLaDOS TTS models

csukuangfj:
See also
https://github.com/rhasspy/piper/issues/187


csukuangfj:
See also https://github.com/rhasspy/piper/issues/187

PR #489, opened 2023-12-15
Keep multiple threads from calling into espeak-ng at the same time

csukuangfj:
Turns out espeak-ng is not thread-safe. This PR adds a lock for espeak-ng.


csukuangfj:
Tested by one user using Python with multi threads and it works fine. Merging.

thewh1teagle:
Does sherpa c-api is thread safe? Can I consider using it across threads with locks myself?
https://github.com/thewh1teagle/sherpa-rs/issues/6

csukuangfj:
Yes,  all functions in c-api.h are thread-safe.

PR #500, opened 2023-12-21
Add CMake option to disable/enable TTS support

### LINKED ISSUE #498: Conditional inclusion of TTS
Can a build option be added to disable the building of TTS related parts of sherpa-onnx? Something like `SHERPA_ONNX_ENABLE_TTS`? I'm building on sherpa-onnx as a library for speech recognition and would prefer to not have to maintain a divergent fork. What's especially problematic for me is the usage and linking to `espeak-ng`, which I believe is under a GPLv3 license and thus somewhat incompatible with the license of this project.

I can create a PR for this.

rkjaran:
The PR adds a CMake option `SHERPA_ONNX_ENABLE_TTS` which is `ON` by default to 
control whether to compile TTS support.

Closes #498

### DIFF
--- a/sherpa-onnx/c-api/c-api.h
+++ b/sherpa-onnx/c-api/c-api.h
@@ -600,6 +600,8 @@ SHERPA_ONNX_API void SherpaOnnxDestroySpeechSegment(
 SHERPA_ONNX_API void SherpaOnnxVoiceActivityDetectorReset(
     SherpaOnnxVoiceActivityDetector *p);
 
+#if defined(SHERPA_ONNX_ENABLE_TTS)
+
 // ============================================================
 // For offline Text-to-Speech (i.e., non-streaming TTS)
 // ============================================================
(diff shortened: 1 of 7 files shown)


sherpa-onnx/c-api/c-api.h
```diff
@@ -600,6 +600,8 @@ SHERPA_ONNX_API void SherpaOnnxDestroySpeechSegment(
 SHERPA_ONNX_API void SherpaOnnxVoiceActivityDetectorReset(
     SherpaOnnxVoiceActivityDetector *p);
 
+#if defined(SHERPA_ONNX_ENABLE_TTS)
```
rkjaran: I should probably invert this, so as to not break the examples.

sherpa-onnx/csrc/CMakeLists.txt
```diff
@@ -94,15 +92,20 @@ set(sources
...
-  offline-tts-impl.cc
-  offline-tts-model-config.cc
-  offline-tts-vits-model-config.cc
-  offline-tts-vits-model.cc
-  offline-tts.cc
-  wave-writer.cc
-)
+
+if(SHERPA_ONNX_ENABLE_TTS)
+  list(APPEND sources
```
csukuangfj, line 97: How about to build a separate library for tts, e.g., `libsherpa-onnx-core-tts.so`?

If you only want to use the ASR part, then you don't need to link with `libsherpa-onnx-core-tts.so`.

rkjaran: Yes. That makes more sense.

Issue #535, opened 2024-01-19
UTF encoding of tokens in python


vesis84:
Hello,
I use latest `master` of sherpa-onnx with pybind11 (v1.9.7).

The output from `OnlineRecognizerResult.text`  is with correct encoding:
(e.g.: `"Kdyť autobusu , já bych chtěl nahlásit o dovolená voda a požár zarážek ."`)

But the content of `OnlineRecognizerResult.tokens` is wrong:
(e.g.: `[' K', '<0x64>', '<0x79>', 'ť', ' a', '<0x75>', 'to', 'bu', '<0x73>', '<0x75>', ' ,', ' já', ' bych', ' chtěl', ' na', '<0x68>', 'lá', '<0x73>', '<0x69>', '<0x74>', ' o', ' dovol', 'en', 'á', ' v', '<0x6F>', 'da', ' a', ' po', 'ž', 'á', '<0x72>', ' za', 'rá', 'že', '<0x6B>', ' .']`)

For some reason, the single-char strings in range `[a-z]` come out as hexa-decimal codes.
This is what comes directly from the pybind11 property [OnlineRecognizerResult.tokens](https://github.com/k2-fsa/sherpa-onnx/blob/bf1dd3daf6f19f5abfbeeb3a0e5d66db997c29e7/sherpa-onnx/python/csrc/online-recognizer.cc#L23)

In the previous version `db41778e995277c6aee898ac75ff90fd05a6e837` (v1.8.11), it was still okay (~end of november).

I tried to reimplement `OnlineRecognizerResult.tokens` with: 
`PyUnicode_DecodeUTF8` as it is done in `OnlineRecognizerResult.text`. But that did not solve the issue.

Would you have an idea what might be possibly causing it ?
Best regards,
Karel

csukuangfj:
Are you using a transducer or a CTC model?

vesis84:
it is a transducer model

vesis84:
Well, should it be the other way round ?
`if (sym.size() == 1 && sym[0] != ' ' && sym[0] >= 0x80) {`
Otherwise the 1char BPE tokens like 'a' 'b' 'c' will still be converted to hex-code...

vesis84:
Yes, it solves it, should i open PR with it, would is also be in CTC ?
Does it break the 'byte output', that someone else needed ?

csukuangfj:
>  should i open PR with it, would is also be in CTC

Yes, please.

---

[...]

PR #550, opened 2024-01-26
Fixes issue #535 , fix hexa 1-char tokens in ASR output.

vesis84:
- This is to avoid output like : `[' K', '<0x64>', '<0x79>', 'ť', ' a', '<0x75>', 'to', 'bu', '<0x73>', '<0x75>', ... ]` with regular 500 BPE units.
- Don't rewrite 1-char output tokens in range `[ 0x20 (space) .. 0x7E (tilde) ]`


csukuangfj:

PR #560, opened 2024-01-29
Update test.py

20246688:
Because the model needs a certain length of input data to work correctly, just like we need a certain amount of time to do something. If the input data is too short, the model may feel unsatisfied, just like we only have a few minutes to complete a task, but actually need more time. So, we added some "virtual" data to make the model feel like it has enough time to handle the task.因为模型需要一定长度的输入数据才能正确工作，就好像我们需要一定长度的时间来做某些事情一样。如果输入数据太短，模型可能会感到不满足，就像是我们只有几分钟的时间来完成一项任务，但实际上需要更多的时间一样，所以我们添加了一些“虚拟”的数据，让模型觉得它有足够的时间来处理任务。

### DIFF
--- a/scripts/whisper/test.py
+++ b/scripts/whisper/test.py
@@ -259,7 +259,7 @@ def compute_features(filename: str) -> torch.Tensor:
 
     # We pad 50 frames at the end so that it is able to detect eot
     # You can use another value instead of 50.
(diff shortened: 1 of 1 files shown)


csukuangfj:

PR #571, opened 2024-02-02
Track token scores

KarelVesely84:
Hello,
I modified the code, so that the per-token scores are accessible via python API 
(ys_probs, lm_probs, context_scores).

This is tracked during the modified beam search decoding of online transduces,
and it is exported for the BestHypothesis.

Would you be interesting in having this functionality in the codebase ?
This can be used as confidences from the client side...

But I did not evaluate it yet, to see how "useful" these numbers are.
I am ready to do changes as well. I'd like to open the discussion...

Best regards
K. Veselý

### DIFF
--- a/sherpa-onnx/csrc/online-recognizer.cc
+++ b/sherpa-onnx/csrc/online-recognizer.cc
@@ -20,12 +20,12 @@ namespace sherpa_onnx {
 
 /// Helper for `OnlineRecognizerResult::AsJsonString()`
 template<typename T>
-const std::string& VecToString(const std::vector<T>& vec, int32_t precision = 6) {
+std::string VecToString(const std::vector<T>& vec, int32_t precision = 6) {
   std::ostringstream oss;
   oss << std::fixed << std::setprecision(precision);
   oss << "[ ";
   std::string sep = "";
-  for (auto item : vec) {
+  for (const auto& item : vec) {
     oss << sep << item;
     sep = ", ";
   }
@@ -35,13 +35,12 @@ const std::string& VecToString(const std::vector<T>& vec, int32_t precision = 6)
 
 /// Helper for `OnlineRecognizerResult::AsJsonString()`
 template<>  // explicit specialization for T = std::string
-const std::string& VecToString<std::string>(const std::vector<std::string>& vec,
-                                            int32_t) // ignore 2nd arg
-{
+std::string VecToString<std::string>(const std::vector<std::string>& vec,
+                                     int32_t) {  // ignore 2nd arg
   std::ostringstream oss;
   oss << "[ ";
   std::string sep = "";
-  for (auto item : vec) {
+  for (const auto& item : vec) {
     oss << sep << "\"" << item << "\"";
     sep = ", ";
   }
@@ -57,9 +56,10 @@ std::string OnlineRecognizerResult::AsJsonString() const {
   os << "\"timestamps\": " << VecToString(timestamps, 2) << ", ";
   os << "\"ys_probs\": " << VecToString(ys_probs, 6) << ", ";
   os << "\"lm_probs\": " << VecToString(lm_probs, 6) << ", ";
-  os << "\"constext_scores\": " << VecToString(context_scores, 6) << ", ";
+  os << "\"context_scores\": " << VecToString(context_scores, 6) << ", ";
   os << "\"segment\": " << segment << ", ";
-  os << "\"start_time\": " << std::fixed << std::setprecision(2) << start_time  << ", ";
+  os << "\"start_time\": " << std::fixed << std::setprecision(2)
+     << start_time  << ", ";
   os << "\"is_final\": " << (is_final ? "true" : "false");
   os << "}";
   return os.str();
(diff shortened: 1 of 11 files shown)


csukuangfj:
> Would you be interesting in having this functionality in the codebase ?

Yes, it is a useful feature and will benefit some users who want to get confidence for each token.

In addition to the modified_beam_search, could you also update the greedy_search?

KarelVesely84:
Okay, i extended it also for the `greedy_search`. 
I did not test it yet... I'll do it soon...

KarelVesely84:
Hi Fangyun @csukuangfj , 
now, the per-token confidences are accessible both for `greedy_search` and `modified_beam_search` 
of  transducer decoding.

I manually tested it with CPU compiled sherpa-onnx through python API.

In the test-workflow I see some segmentation fault for Online-CTC in linux-gpu test.
Is this because the `OnlineRecognizerResult` class was extended in my PR ?

Do you also run the tests locally to investigate ?
K.

csukuangfj:
Thanks! I will have time after today. Will look into it tomorrow.

sherpa-onnx/csrc/online-recognizer.cc
```diff
@@ -18,56 +18,50 @@
...
-
-  os << "\""
-     << "timestamps"
-     << "\""
-     << ": ";
-  os << "[";
-
+/// Helper for `OnlineRecognizerResult::AsJsonString()`
+template<typename T>
+const std::string& VecToString(const std::vector<T>& vec, int32_t precision = 6) {
```
csukuangfj:
```suggestion
std::string VecToString(const std::vector<T>& vec, int32_t precision = 6) {
```

sherpa-onnx/csrc/online-recognizer.cc
```diff
@@ -18,56 +18,50 @@
...
-    } else {
-      os << sep << "\"" << t << "\"";
-    }
+  oss << " ]";
+  return oss.str();
+}
+
+/// Helper for `OnlineRecognizerResult::AsJsonString()`
+template<>  // explicit specialization for T = std::string
+const std::string& VecToString<std::string>(const std::vector<std::string>& vec,
```
csukuangfj:
```suggestion
std::string VecToString<std::string>(const std::vector<std::string>& vec,
```

csukuangfj: Please never return a temporary reference from a function.

KarelVesely84: Aha, thank you. 

My bad. I remember in some context returning a local object as a const reference worked, but here I did it badly.
https://stackoverflow.com/questions/13318257/const-reference-to-temporary-vs-return-value-optimization

Best, K.

csukuangfj: > I remember in some context returning a local object as a const reference worked

Yes, you are right. But in this case, the temporary reference is from the function return value of a temporary object.
Think twice, there are two indirections.

Suppose `obj` is a local variable in a function. It is totally fine to call
```
return obj
```
to return a const reference.

But it is not valid to invoke
```
return obj.someMethod();
```
since `obj` is destroyed when the function returns.

KarelVesely84: Yes, you are right.

The code is fixed now. I tested the JSON output from python, and it looks fine.

Also now the lm_probs and contexts scores are filled only if LM and ContextGraph are used. 
Otherwise there are empty lists to save bandwidth.

Thank you for finding the bug,
Karel

csukuangfj:
I see that you have removed WIP. Do you think it is ready for review and merge?

KarelVesely84:
Ok, the workflow tests seem fine, including the linter style-check... It is ready...

csukuangfj:
I will have a second look after dinner. Thanks!

sherpa-onnx/csrc/online-recognizer.cc
```diff
@@ -18,56 +18,50 @@
...
+}
+
+/// Helper for `OnlineRecognizerResult::AsJsonString()`
+template<>  // explicit specialization for T = std::string
+std::string VecToString<std::string>(const std::vector<std::string>& vec,
+                                     int32_t) {  // ignore 2nd arg
+  std::ostringstream oss;
+  oss << "[ ";
+  std::string sep = "";
+  for (auto item : vec) {
```
csukuangfj:
```suggestion
  for (const auto i&tem : vec) {
```

sherpa-onnx/csrc/online-transducer-greedy-search-decoder.cc
```diff
@@ -138,6 +142,17 @@ void OnlineTransducerGreedySearchDecoder::Decode(
       } else {
         ++r.num_trailing_blanks;
       }
+
+      // export the per-token log scores
+      if (y != 0 && y != unk_id_) {
+        LogSoftmax(p_logit, vocab_size);  // renormalize probabilities,
+                                          // save time by doing it only for
+                                          // emitted symbols
+        float *p_logprob = p_logit;  // rename p_logit as p_logprob,
```
csukuangfj:
```suggestion
        const float *p_logprob = p_logit;  // rename p_logit as p_logprob,
```

csukuangfj (approved):
Thanks again!

Left some minor comments. Otherwise, it looks good to me.

KarelVesely84:
okay, the two suggestions are implemented.... thank you...

csukuangfj:
Thanks!

PR #648, opened 2024-03-08
add WebAssembly for Kws

lovemefan:


wasm/kws/assets/README.md
```diff
@@ -0,0 +1,27 @@
...
+# Kws
+
+The following is an example:
+```
+cd sherpa-onnx/wasm/kws
+git clone https://www.modelscope.cn/pkufool/sherpa-onnx-kws-zipformer-wenetspeech-3.3M-2024-01-01.git assets
+```
+
+You should have the following files in `assets` before you can run
+`build-wasm-simd-kws.sh`
```
csukuangfj, line 16: Please also check in this file.

You need to use
```bash
cd sherpa-onnx
git add -f ./build-wasm-simd-kws.sh
```

lovemefan: 👌

sherpa-onnx/csrc/keyword-spotter-transducer-impl.h
```diff
@@ -266,15 +266,25 @@ class KeywordSpotterTransducerImpl : public KeywordSpotterImpl {
   }
 
   void InitKeywords() {
-    // each line in keywords_file contains space-separated words
 
+#ifdef SHERPA_ONNX_ENABLE_WASM_KWS
+    // Due to the limitations of the wasm file system,
```
csukuangfj, line 270: Could you pass a keyword file from wasm?

We have been doing this for model files, such as `tokens.txt`.

wasm/kws/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
+if(NOT $ENV{SHERPA_ONNX_IS_USING_BUILD_WASM_SH})
+    message(FATAL_ERROR "Please use ./build-wasm-kws.sh to build for wasm KWS")
```
csukuangfj:
```suggestion
    message(FATAL_ERROR "Please use ./build-wasm-simd-kws.sh to build for wasm KWS")
```

csukuangfj (commented):
Thank you! Left some minor comments.

sherpa-onnx/csrc/keyword-spotter-transducer-impl.h
lovemefan: Sure, but  it needs to be recompiled when the keywords is modified, which is very inconvenient. Because token.txt and keywords.txt are both packaged into the sherpa-onnx-wasm-kws-main.data file, they cannot be modified and can only be recompiled to generate. What do you think?

csukuangfj: Ah, I see. Then please keep using your current approach.

.github/workflows/build-wheels-aarch64.yaml
```diff
@@ -17,14 +17,13 @@ concurrency:
 
 jobs:
   build_wheels_aarch64:
-    name: ${{ matrix.manylinux }} ${{ matrix.python-version }}
+    name: ${{ matrix.python-version }}
```
csukuangfj: Please use changes from the latest master for `*.yaml` files that you have changed.

lovemefan: sorry, I forgot to merge other files

sherpa-onnx/c-api/c-api.cc
```diff
@@ -473,10 +473,16 @@ SherpaOnnxKeywordSpotter* CreateKeywordSpotter(
     SHERPA_ONNX_LOGE("%s\n", spotter_config.ToString().c_str());
   }
 
+#ifndef SHERPA_ONNX_ENABLE_WASM_KWS
```
csukuangfj: Could you disable only the check for hotword file?
We can still validate model files and tokens.txt.

lovemefan: ok, thank you for your suggestions

csukuangfj:
Thank your for your first-time contribution!

csukuangfj (approved):

PR #650, opened 2024-03-08
Add microphone streaming ASR example for C API

csukuangfj:


livesflying:
you should add #include “” at the file

csukuangfj:
it's not an error. It's a symbolic link.

PR #673, opened 2024-03-18
Fix argument for beam search Decoding

bhaswa:


csukuangfj:

PR #687, opened 2024-03-22
Update MainActivity.kt

mablue:
Appending a default text to test field.
To faster check the voices


csukuangfj:

mablue:
This is the smallest thing I can do in return for the service you do to the society.

PR #707, opened 2024-03-27
add openfst.cmake file

hantengc:
1. When compiling locally, openfst is missing.so add this file to the sherpa-onnx/cmake folder


csukuangfj:

PR #714, opened 2024-03-28
delete incorrect logs

longshiming:
delete incorrect logs


csukuangfj:

PR #726, opened 2024-04-01
Appending TTS Engine Next Generation by @jing332

mablue:


csukuangfj:
Could you add it directly without using git submodule?

mablue:
yes


> Could you add it directly without using git submodule?

yes appended as normal folder

PR #748, opened 2024-04-10
Fix a bug in mean calculation of 'ys_probs'

aask1357:
A vector variable `best_hyp.ys_probs` contains the probabilities of each token in a matched keyword whose length is `level`.
The code checks if the mean of `best_hyp.ys_probs` is greater than or equal to a threshold.
### Problem
Currently, when calculating the mean, the index ranges from 1 to `level`.
This excludes the first token (at index 0) and includes an undefined value at the end.
In most cases,  this undefined value is zero so the code seems to work fine.
### Solution
Change the index to range  from 0 to `level`-1.


csukuangfj:
@pkufool Could you have a look?

pkufool:
@aask1357  Though I can't recall why I wrote down this code before，after reviewing the current code carefully, I think you are right, Thanks!

pkufool (approved):

PR #765, opened 2024-04-13
Replace torchaudio with soundfile in python-api-examples

gtf35:
`torchaudio` is a part of `PyTorch`. 

`PyTorch` is very large, but only a small part is used in `python-api-examples`.

In real projects,  it is no need to use `PyTorch`,

`soundfile` is smaller and has been used in the project,

So, I replace `torchaudio` with `soundfile` in `python-api-examples`,

Please review :)


csukuangfj:

PR #789, opened 2024-04-18
Adding temperature scaling on Joiner logits:

KarelVesely84:
- T is configurable
- so far best result NCE 0.122 (still not so high) 
  - the BPE scores were rescaled with 0.2 (but then also incorrect words get high confidence, visually reasonable histograms are for 0.5 scale) 
  - BPE->WORD score merging done by min(.) function (tried also prob-product, and also arithmetic, geometric, harmonic mean)

- without temperature scaling (i.e. scale 1.0), the best NCE was 0.032 (here product merging was the best)

Results seem consistent with: https://arxiv.org/abs/2110.15222

Everything tuned on a very small set of 100 sentences with 813 words and 10.2% WER, a Czech model.

I also experimented with blank posteriors mixed into the BPE confidences, but no NCE improvement found, so not pushing that.

Temperature scling added also to the Greedy search confidences.

### DIFF
--- a/sherpa-onnx/csrc/online-transducer-modified-beam-search-decoder.cc
+++ b/sherpa-onnx/csrc/online-transducer-modified-beam-search-decoder.cc
@@ -129,6 +129,22 @@ void OnlineTransducerModifiedBeamSearchDecoder::Decode(
         model_->RunJoiner(std::move(cur_encoder_out), View(&decoder_out));
 
     float *p_logit = logit.GetTensorMutableData<float>();
+
+    // copy raw logits, apply temperature-scaling (for confidences)
+    int32_t p_logit_items = vocab_size * num_hyps;
+    std::vector<float> logit_with_temperature(p_logit_items);
+    {
+      std::copy(p_logit,
+                p_logit + p_logit_items,
+                logit_with_temperature.begin());
+      // TODO(KarelVesely84): configure externally ?
+      float temperature_scale = 2.0;
+      for (float& elem : logit_with_temperature) {
+        elem /= temperature_scale;
+      }
+      LogSoftmax(logit_with_temperature.data(), vocab_size, num_hyps);
+    }
+
     if (blank_penalty_ > 0.0) {
       // assuming blank id is 0
       SubtractBlank(p_logit, vocab_size, num_hyps, 0, blank_penalty_);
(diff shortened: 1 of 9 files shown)


sherpa-onnx/csrc/online-transducer-modified-beam-search-decoder.cc
```diff
@@ -129,6 +129,22 @@ void OnlineTransducerModifiedBeamSearchDecoder::Decode(
         model_->RunJoiner(std::move(cur_encoder_out), View(&decoder_out));
 
     float *p_logit = logit.GetTensorMutableData<float>();
+
+    // copy raw logits, apply temperature-scaling (for confidences)
+    int32_t p_logit_items = vocab_size * num_hyps;
+    std::vector<float> logit_with_temperature(p_logit_items);
```
csukuangfj, line 137: Can we change p_logit in-place?

csukuangfj:
Thanks!

> T hard-coded to 2.0 (is it okay ?)


 Could you make it configurable and give it a default value 1.0 (like what we are doing for blank penalty)?
If it is 1.0, then the loop is skipped.

sherpa-onnx/csrc/online-transducer-modified-beam-search-decoder.cc
KarelVesely84: no, it cannot be done in-place

the idea is to apply temperature only for computation of confidences, 
the decoding continues to use the original values

this is why the logit values are copied to a new buffer

csukuangfj: Thanks for the explanation.

KarelVesely84:
> Thanks!
> 
> > T hard-coded to 2.0 (is it okay ?)
> 
> Could you make it configurable and give it a default value 1.0 (like what we are doing for blank penalty)? If it is 1.0, then the loop is skipped.

okay, working on it

i am not sure about the default 1.0, for 1.0 the confidences have worse quality than for 2.0,
and, as it is implemented, setting it to 2.0 will not affect the decoding at all, just the confidence
values will be different, to **proposing to set it to 2.0**
(same value appears in the Google article for NN-posterios confidences)

csukuangfj:
1.0 is for backward compatibility.

> as it is implemented, setting it to 2.0 will not affect the decoding at all,

In that case, 2.0 is fine with me.

KarelVesely84:
okay, the T parameter is now configurable

KarelVesely84:
there seems to be some problem with the workload tests, many fail with the 503 error leading to huggingface

KarelVesely84:
i fixed an error in the android build

KarelVesely84:
the tests look OK, seeing only unrelated errors

csukuangfj:

PR #899, opened 2024-05-21
Update offline-recognizer-transducer-impl.h

Egorgij21:
there were no initialization of hotwords for offline-recognizer-transducer for __ANDROID_API__ > 9. so i added it for my needs and it started work.


Egorgij21:
Screenshot 2024-05-21 at 21 03 32
someone already made these changes yesterday...

PR #903, opened 2024-05-22
Fix reading wave files generated by NAudio.

csukuangfj:
See also
https://github.com/naudio/NAudio/issues/1132


csukuangfj:
CC @eugeneYz

PR #904, opened 2024-05-22
Add Dart API for VAD

csukuangfj:


csukuangfj:
>  I wonder whether you need help with this Flutter / Dart port

Yes, definitely. Help from the community is always appreciated.

As for the Dart API, there are the following TODO items:

1. Wrap the following to Dart

- [ ] non-streaming ASR
- [ ] TTS
- [ ] Spoken language identification
- [ ] audio tagging
- [ ] keyword spotting

I am working on the streaming ASR part.

By the way, we are not using  [ffigen](https://pub.dev/packages/ffigen). We prefer to wrap the C API one-by-one and test it incrementally as we wrap it.

2.  Follow [Federated plugin](https://docs.flutter.dev/packages-and-plugins/developing-packages#federated-plugins)
to refactor our current approach.

   There are two examples of federated plugin that we can follow.
      - https://pub.dev/packages/url_launcher
      - https://pub.dev/packages/record

3. Write Flutter demos for the above APIs.   I am also working on the Flutter streaming ASR demo.

PR #917, opened 2024-05-24
Update offline-ctc-greedy-search-decoder.cc

Dadoou:
Bug fixes.
Z_O_O will be decoded as ZO instead of ZOO.
To fix this, prev_id should update every time.


csukuangfj:
You are right. Thanks!

Issue #41, opened 2023-02-20
[Help wanted] Support quantization


csukuangfj:
# TODO

- [ ] Support fp16 and/or int8 with TensorRT

manickavela29:
Hi @csukuangfj ,

Currently I am doing some experiments with Zipformer models, let me know if there are active developments going on from your end for TensorRT low precision support, else I will take up on this  task  :)

csukuangfj:
>  else I will take up on this task :)

Thanks! That would be great. We don't have a plan to support tensorrt in the near future.

PR #921, opened 2024-05-26
initial tensorrt ep commit

manickavela29:
- Adds in TensorRT EP
- Observed that Encoder Model is much faster with TRT than with CUDA backend
- Initialization of onnxrt session takes more time as TRT models are generated and converted

ToDo : 
- [x] Validate Accuracy of the model inference
- [x] Reducing the starting time

Ref : https://github.com/k2-fsa/sherpa-onnx/issues/40, https://github.com/k2-fsa/sherpa-onnx/issues/41
CC : @csukuangfj


sherpa-onnx/csrc/session.cc
```diff
@@ -115,6 +116,54 @@ static Ort::SessionOptions GetSessionOptionsImpl(int32_t num_threads,
...
+          "1",
+          ".",  // can be same as the engine cache folder
+      };
+      std::vector<std::string> available_providers =
+          Ort::GetAvailableProviders();
+      if (std::find(available_providers.begin(), available_providers.end(),
+                    "TensorrtExecutionProvider") != available_providers.end()) {
+        const auto& api = Ort::GetApi();
+
+        OrtTensorRTProviderOptionsV2* tensorrt_options;
```
csukuangfj: Please refer to
https://onnxruntime.ai/docs/api/c/struct_ort_api.html#aded45902af604f958176bec4792351c0

You need to invoke

[OrtApi::ReleaseTensorRTProviderOptions](https://onnxruntime.ai/docs/api/c/struct_ort_api.html#ae17a8857260554d44bb2b27a67e39020)
to release `tensorrt_options` to avoid memory leak.

sherpa-onnx/csrc/session.cc
```diff
@@ -115,6 +116,54 @@ static Ort::SessionOptions GetSessionOptionsImpl(int32_t num_threads,
...
+        Ort::ThrowOnError(api.CreateTensorRTProviderOptions(&tensorrt_options));
```
csukuangfj: Please follow
https://github.com/k2-fsa/sherpa-onnx/blob/b7148174739275dfc997af726be364245511239c/sherpa-onnx/csrc/session.cc#L98-L107

to replace `Ort::ThrowOnError`.

If there was an error, we just fallback to CPU instead of throwing an exception.

sherpa-onnx/csrc/session.cc
```diff
@@ -115,6 +116,54 @@ static Ort::SessionOptions GetSessionOptionsImpl(int32_t num_threads,
 #endif
       break;
     }
+    case Provider::kTRT: {
+      SHERPA_ONNX_LOGE("Came for TRT");
```
csukuangfj: Please remove the log when it is ready for merge.

sherpa-onnx/csrc/session.cc
manickavela29: Handled this in the latest commit

sherpa-onnx/csrc/session.cc
manickavela29: Handled this in the latest commit

manickavela29:
I will send a separate PR for handling configs of OnnxRT EP configs separately.
along with TRT config options.

In the middle of something, and it might take sometime

sherpa-onnx/csrc/session.cc
```diff
@@ -54,6 +54,61 @@ static Ort::SessionOptions GetSessionOptionsImpl(int32_t num_threads,
...
+        else if(statusU) { OrtStatusFailure(statusU,&os);}
+        else {
+          auto *statusTrt = OrtSessionOptionsAppendExecutionProvider_Tensorrt(
+                              sess_opts, 0);
+          if(statusTrt) { OrtStatusFailure(statusTrt,&os);}
+        }
+
+        api.ReleaseTensorRTProviderOptions(tensorrt_options);
+      }
+    }
```
csukuangfj, line 116: Please add a comment here describing that you omit `break` intentionally. If TensorRT is not available, it tries to use CUDA instead.

sherpa-onnx/csrc/session.cc
```diff
@@ -214,4 +220,13 @@ Ort::SessionOptions GetSessionOptions(
   return GetSessionOptionsImpl(config.num_threads, config.provider);
 }
 
+void OrtStatusFailure(OrtStatus *status,std::ostringstream *os) {
```
csukuangfj: This function is used only inside this file, so you can remove it from the header file, and move this function to the beginning of this file and mark it as `static`.

manickavela29: ok I will make the changes

manickavela29:
I think the build failures are coming in from the python dependency of nvinfer and similar, I will try to add some libraries and check

csukuangfj:
https://github.com/k2-fsa/sherpa-onnx/actions/runs/9354678305/job/25748084556#step:5:1121

```
[ 62%] Linking CXX executable ../../bin/sherpa-onnx
/opt/rh/devtoolset-10/root/usr/libexec/gcc/x86_64-redhat-linux/10/ld: ../../lib/libsherpa-onnx-core.so: undefined reference to `OrtSessionOptionsAppendExecutionProvider_Tensorrt'
collect2: error: ld returned 1 exit status
make[2]: *** [bin/sherpa-onnx] Error 1
make[1]: *** [sherpa-onnx/csrc/CMakeFiles/sherpa-onnx.dir/all] Error 2
make: *** [all] Error 2
```

I think an extra lib should be linked to for tensorrt.

manickavela29:
yes, it seems directly having  OrtSessionOptionsAppendExecutionProvider_Tensorrt() as in interface is not working out
and AppendExecutionProvider_TensorRT_V2 is actually a wrapper around that function.

so it should be alright now

sherpa-onnx/csrc/session.cc
```diff
@@ -53,6 +64,57 @@ static Ort::SessionOptions GetSessionOptionsImpl(int32_t num_threads,
...
+        const char* op_keys;
+        const char* op_values;
+      };
+
+      std::vector<TrtPairs> trt_options = {
+        {"device_id", "0"},
+        {"trt_max_workspace_size", "2147483648"},
+        {"trt_max_partition_iterations", "10"},
+        {"trt_min_subgraph_size", "5"},
+        {"trt_fp16_enable", "0"},
```
yuekaizhang, line 77: I think we could enable fp16 here to get faster.

manickavela29: Yess, I tried with fp16 also, and didn't see improvement in performance. 

But still with the config patch it will be an option for people to enable

sherpa-onnx/csrc/session.cc
```diff
@@ -8,6 +8,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+// #include <cuda.h>
```
csukuangfj: Could you remove this unused line?

sherpa-onnx/csrc/session.cc
```diff
@@ -21,6 +22,16 @@
 
 namespace sherpa_onnx {
 
+
+static void OrtStatusFailure(OrtStatus *status, std::ostringstream *os) {
```
csukuangfj:
```suggestion
static void OrtStatusFailure(OrtStatus *status, const char *s) {
```

You can pass `os->str().c_str()` to this function

sherpa-onnx/csrc/session.cc
```diff
@@ -53,6 +64,57 @@ static Ort::SessionOptions GetSessionOptionsImpl(int32_t num_threads,
...
+                tensorrt_options, option_keys.data(), option_values.data(),
+                option_keys.size());
+        sess_opts.AppendExecutionProvider_TensorRT_V2(*tensorrt_options);
+
+        if (statusC) { OrtStatusFailure(statusC, &os); }
+        if (statusU) { OrtStatusFailure(statusU, &os); }
+
+        api.ReleaseTensorRTProviderOptions(tensorrt_options);
+      }
+      // break; is ommited here intentially so that
```
csukuangfj:
```suggestion
      // break; is omitted here intentionally so that
```

By the way, if tensorrt is available, do we need to put a `break` here?

csukuangfj:
Please merge the latest master into your branch and the CI should pass or you can just ignore the failed tests.

Could you describe what users need to do to build sherpa-onnx with TensorRT support?

csukuangfj:
please leave a comment if you think it is ready to merge.

sherpa-onnx/csrc/session.cc
manickavela29: I would keep it so,
onnxruntime by default has the fall back mechanism for their EP's ,

if any subrgraph is failing at Tensorrt EP, it will come to CUDA,
this will be a good way to keep it future proof.


CPU EP must also be added at bottom for good practise.

manickavela29: I checked with Zipformer and it is holding good,
and since at icefall itself k2 dependency is removed most of the models should be consistent by default

manickavela29:
Regarding what is required to run TensorRT,

Hardware      : Nvidia GPU 😄 ,
Software lib  : TensorRT requires libnvinfer, libnvinfer-dispatch, libnvinfer-plugin, libnvonnxparsers8 (from my experience)

But seeing that CI/CD's are working fine, I don't think onnxrt is able to handle these, 
as long as someone is having a compatible GPU it should work fine.

manickavela29:
It can be merged,
after ensuring that basic builds are working fine

Thank you for the review 😄

manickavela29:
> Regarding what is required to run TensorRT,
> 
> Hardware : Nvidia GPU 😄 ,(not sure how compatible AMD GPUs are) Software lib : TensorRT requires libnvinfer, libnvinfer-dispatch, libnvinfer-plugin, libnvonnxparsers8 (from my experience)
> 
> But seeing that CI/CD's are working fine, I think onnxrt is able to handle these, as long as someone is having a compatible GPU it should work fine.
> 
> Just setting the appropriate provider argument should be enough, if there are any github pages/doc, this can be updated there
> 
> --provider=trt

Update this commit, by mistake I had typed ' don't' but edited it

manickavela29:
> Please merge the latest master into your branch and the CI should pass or you can just ignore the failed tests.
> 
> Could you describe what users need to do to build sherpa-onnx with TensorRT support

I think the builds are also good enough. 
Shall we merge it @csukuangfj

csukuangfj:

PR #933, opened 2024-05-28
Add Dart API for streaming ASR

csukuangfj:


csukuangfj:
Note:  https://github.com/llfbandit/record does not work well with Linux. 
For instance, it does not support pcm16bit.
Also, it supports only recording to a file on Linux.

Help from the community is appreciated to switch to
https://github.com/SKKbySSK/coast_audio

PR #936, opened 2024-05-29
Added tokens, tokens_arr and json for offline recognizer result

leohuang2013:
Added following new fields in offline recognizer result

- tokens
- tokens_arr 
- json


csukuangfj:

PR #941, opened 2024-05-29
Split online.cs and offline.csFile 

dfengpo:
拆分 scripts 下的dotnet 文件online.cs，offline.cs。


csukuangfj:

PR #889, opened 2024-05-17
Add C++ runtime for *streaming* faster conformer transducer from NeMo. 

sangeet2020:
This PR is to integrate Nemo's faster conformer transducer into sherpa-decoder.
More commits to be added.

PR #944, opened 2024-05-30
Fix nemo streaming transducer greedy search

csukuangfj:
See also #889

@sangeet2020


sangeet2020:
Thank you @csukuangfj .
I will study your changes and review my mistakes .

PR #992, opened 2024-06-12
Onnxrt execution provider configs

manickavela29:
Config files handling for onnxrt provider


sherpa-onnx/csrc/onnxrt-execution-provider-config.h
```diff
@@ -0,0 +1,94 @@
...
+  uint32_t trt_fp16_enable = 1;
+  uint32_t trt_detailed_build_log = 0;
+  uint32_t trt_engine_cache_enable = 1;
+  std::string trt_engine_cache_path = ".";
+  uint32_t trt_timing_cache_enable = 1;
+  std::string trt_timing_cache_path = ".";
+
+  std::string encoder;
+  std::string decoder;
+  std::string joiner;
```
csukuangfj: Could you remove model-related fields from this config?

This config can be used for different types of models, not limited to transducer models.

sherpa-onnx/csrc/onnxrt-execution-provider-config.h
```diff
@@ -0,0 +1,94 @@
+// sherpa-onnx/csrc/online-transducer-model-config.h
+//
+// Copyright (c)  2023  Xiaomi Corporation
```
csukuangfj: Please change it to your name.

sherpa-onnx/csrc/onnxrt-execution-provider-config.h
manickavela29: yess yes, it is a very early patch, and just created templates.
will remove them, will request for review once completed

csukuangfj: Thanks!

manickavela29:
Hi,
I haven't tried with any Offlinemodel, so I am extending the config support to only for OnlineModel,
but easily it can be extended for OfflineModel.
Let me know if you have any better suggestions for maintaining the config.

session.cc files also has several other models (vad,etc), since  these are minimal models, they are left as they are
if the support for trt/cuda backends can be extended for them.
let me know

sherpa-onnx/csrc/online-model-config.cc
```diff
@@ -87,6 +85,10 @@ bool OnlineModelConfig::Validate() const {
     return nemo_ctc.Validate();
   }
 
+  if (!provider_config.Validate()) {
+    return provider_config.Validate();
```
csukuangfj: You can just return `false` since you have already called `Validate()`.

sherpa-onnx/csrc/online-model-config.h
```diff
@@ -60,11 +62,11 @@ struct OnlineModelConfig {
         wenet_ctc(wenet_ctc),
         zipformer2_ctc(zipformer2_ctc),
         nemo_ctc(nemo_ctc),
+        provider_config(provider_config),
         tokens(tokens),
         num_threads(num_threads),
         warm_up(warm_up),
         debug(debug),
-        provider(provider),
```
csukuangfj, line 67: please also remove the `provider` argument from the argument list at line 54.

sherpa-onnx/csrc/online-recognizer.cc
```diff
@@ -113,6 +113,7 @@ void OnlineRecognizerConfig::Register(ParseOptions *po) {
 }
 
 bool OnlineRecognizerConfig::Validate() const {
+  SHERPA_ONNX_LOGE("Args recognizer : %s",ToString().c_str());
```
csukuangfj: please remove it once it is ready.

sherpa-onnx/csrc/provider-config.cc
```diff
@@ -0,0 +1,161 @@
...
+
+void CudaConfig::Register(ParseOptions *po) {
+  po->Register("cuda-cudnn-conv-algo-search", &cudnn_conv_algo_search,
+          "CuDNN convolution algrorithm search");
+}
+
+bool CudaConfig::Validate() const {
+
+  if(cudnn_conv_algo_search > 0 && cudnn_conv_algo_search < 4) {
+    SHERPA_ONNX_LOGE("cudnn_conv_algo_search: '%d' is not valid option."
```
csukuangfj:
```suggestion
    SHERPA_ONNX_LOGE("cudnn_conv_algo_search: '%d' is not a valid option."
```

sherpa-onnx/csrc/provider-config.h
```diff
@@ -0,0 +1,92 @@
...
+#include "sherpa-onnx/csrc/macros.h"
+#include "onnxruntime_cxx_api.h"  // NOLINT
+
+namespace sherpa_onnx {
+
+struct CudaConfig {
+  uint32_t cudnn_conv_algo_search = OrtCudnnConvAlgoSearchHeuristic;
+
+  CudaConfig() = default;
+  CudaConfig(const uint32_t &cudnn_conv_algo_search)
```
csukuangfj: Please remove `const` and `&` for primitive types.

sherpa-onnx/csrc/session.cc
```diff
@@ -64,26 +66,44 @@ static Ort::SessionOptions GetSessionOptionsImpl(int32_t num_threads,
...
-          {"trt_max_partition_iterations", "10"},
-          {"trt_min_subgraph_size", "5"},
-          {"trt_fp16_enable", "0"},
-          {"trt_detailed_build_log", "0"},
-          {"trt_engine_cache_enable", "1"},
-          {"trt_engine_cache_path", "."},
-          {"trt_timing_cache_enable", "1"},
-          {"trt_timing_cache_path", "."}};
+        {"device_id",
+          std::to_string(provider_config->device).c_str()},
```
csukuangfj: Please save `std::to_string()` as a local variable.
Otherwise, you are getting the use-after-free error here.

`c_str()` is from a temp variable that is freed after this line.

Please fix other similar issues in this file. Please never do this again.

sherpa-onnx/csrc/session.cc
```diff
@@ -32,11 +32,13 @@ static void OrtStatusFailure(OrtStatus *status, const char *s) {
 }
 
 static Ort::SessionOptions GetSessionOptionsImpl(int32_t num_threads,
-                                                 std::string provider_str) {
+    std::string provider_str,
```
csukuangfj: please use `const std::string&`

csukuangfj (changes requested):

csukuangfj:
> Hi, I haven't tried with any Offlinemodel, so I am extending the config support to only for OnlineModel, but easily it can be extended for OfflineModel. Let me know if you have any better suggestions for maintaining the config.
> 
> session.cc files also has several other models (vad,etc), since these are minimal models, they are left as they are if the support for trt/cuda backends can be extended for them. let me know
> 
> for pybind interface, still adding in changes

Thanks! That sounds good to me.

csukuangfj:
Please have a look at errors reported by `clang-tidy` 
https://github.com/k2-fsa/sherpa-onnx/actions/runs/9665507110/job/26662771827

and the style check by `cpplint`.

https://github.com/k2-fsa/sherpa-onnx/actions/runs/9665507136

(Note: You can run these two tools locally.)

sherpa-onnx/csrc/provider-config.h
manickavela29: sure, replicate this for Tensorrtconfig objects too for all the primitives?

csukuangfj: arguments of primitive type can be passed directly without `&`. Also we don't use `const` for primitive argments.

sherpa-onnx/csrc/session.cc
manickavela29: sure alright

sherpa-onnx/csrc/provider-config.cc
```diff
@@ -19,7 +19,7 @@ void CudaConfig::Register(ParseOptions *po) {
 bool CudaConfig::Validate() const {
 
   if(cudnn_conv_algo_search > 0 && cudnn_conv_algo_search < 4) {
-    SHERPA_ONNX_LOGE("cudnn_conv_algo_search: '%d' is not valid option."
+    SHERPA_ONNX_LOGE("cudnn_conv_algo_search: '%d' is not a valid option."
```
csukuangfj, line 21: If `cudnn_conv_algo_search` is 1, this if statement is also executed, right?
1 is a valid value since the log says `[1, 3]`.

manickavela29: yes, I will update it

manickavela29:
I wanted to keep provider config unique for each models, this would give good control for CUDA and TRT configs,
But if developers give different devices(outside of cuda/trt) is a complex scenario and it must be handled, so I remove those patches midway.

Most of the fixes I had also went away with those patches thats why several bugs

manickavela29:
> Please have a look at errors reported by `clang-tidy` https://github.com/k2-fsa/sherpa-onnx/actions/runs/9665507110/job/26662771827
> 
> and the style check by `cpplint`.
> 
> https://github.com/k2-fsa/sherpa-onnx/actions/runs/9665507136
> 
> (Note: You can run these two tools locally.)

yes, have been using for cpplint, had a hard time with exit 🙂
will try with clangtiddy

csukuangfj:
By the way, you can use
```
pip install clang-tidy
```

sherpa-onnx/python/csrc/provider-config.cc
```diff
@@ -0,0 +1,91 @@
+// sherpa-onnx/python/csrc/provider-config.h
```
csukuangfj:
```suggestion
// sherpa-onnx/python/csrc/provider-config.cc
```

manickavela29:
Python api is having some issues, can you direct me to particular workflowfile to build and test with it

sherpa-onnx/csrc/provider-config.h
```diff
@@ -67,17 +67,17 @@ struct TensorrtConfig {
 };
 
 struct ProviderConfig {
+  TensorrtConfig trt_config;
+  CudaConfig cuda_config;
+  std::string provider = "cpu";
   uint32_t device = 0;
```
csukuangfj: By the way, is there a reason to use `uint32_t` instead of `int32_t`?

manickavela29: Nothing special, just went with it,
want to keep it consistent with repo, as int32_t?
did think about compatibility with other devices.

csukuangfj: If there is no special reason, please use `int32_t`.

unsigned integers are error-prone.

sherpa-onnx/csrc/session.cc
manickavela29: Didn't know that to_string() can behave like this,
wanted to know more about this behaviour, but was not able to find and doc/resource regarding this.
Could you please share resource about this?

csukuangfj:
> Python api is having some issues, can you direct me to particular workflowfile to build and test with it

Have you fixed it? Sorry for the late reply.

manickavela29:
> By the way, you can use
> 
> ```
> pip install clang-tidy
> ```

what is the cmd to run this?


> > Python api is having some issues, can you direct me to particular workflowfile to build and test with it
> 
> Have you fixed it? Sorry for the late reply.

yes this is fixed, waiting for some build tests to see

csukuangfj:
> what is the cmd to run this?


Just start your terminal, activate your python virtual environment, and run
```bash
pip install clang-tidy
```
and then you can run
```bash
cd /path/to/sherpa-onnx
mkdir build
cd build
cmake -DSHERPA_ONNX_ENABLE_PYTHON=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..
make check
```

manickavela29:
> > what is the cmd to run this?
> 
> 
> Just start your terminal, activate your python virtual environment, and run
> ```bash
> pip install clang-tidy
> ```
> and then you can run
> ```bash
> cd /path/to/sherpa-onnx
> mkdir build
> cd build
> cmake -DSHERPA_ONNX_ENABLE_PYTHON=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..
> make check
> ```
> 

Yes got it, took it from the workflow file 🙌

manickavela29:
Any suggestion for datatype

> /home/runner/work/sherpa-onnx/sherpa-onnx/sherpa-onnx/csrc/provider-config.h:30:36: warning: implicit conversion from 'long' to 'int32_t' (aka 'int') changes value from 2147483648 to -2147483648 [-Wconstant-conversion]
>   int32_t trt_max_workspace_size = 2147483648;
>                                    ^~~~~~~~~~

csukuangfj:
> Any suggestion for datatype
> 
> > /home/runner/work/sherpa-onnx/sherpa-onnx/sherpa-onnx/csrc/provider-config.h:30:36: warning: implicit conversion from 'long' to 'int32_t' (aka 'int') changes value from 2147483648 to -2147483648 [-Wconstant-conversion]
> > int32_t trt_max_workspace_size = 2147483648;
> > ^~~~~~~~~~

For tensorrt related variables, you can use uint32_t.

manickavela29:
Most of the build tests are failing before build step itself, I think there is a pip dependency issue or something
Anyway it is good for review as of now

for jni, I have updated at one place, but I am not sure if it will fix it will completely
I am not very familiar with other interfaces, maybe you can update them accordingly.

I think it is good for review

sherpa-onnx/python/csrc/provider-config.h
csukuangfj, line 1: Please change
https://github.com/k2-fsa/sherpa-onnx/blob/40119487eeef8c4a82d640051982b0533a86c6c6/sherpa-onnx/python/csrc/sherpa-onnx.cc#L46
to
```c++
 PybindFeatures(&m); 
PybindProviderConfig(&m);
```

Note you need to include the header file.

csukuangfj:
By the way, you can ignore the failed websocket server tests.

sherpa-onnx/python/csrc/provider-config.h
manickavela-uni: I will update for this, 
I also noticed files under sherpa-onnx/sherpa-onnx/python/sherpa-onnx
Would be great if someone can update these

sherpa-onnx/csrc/provider-config.cc
```diff
@@ -0,0 +1,155 @@
...
+
+namespace sherpa_onnx {
+
+void CudaConfig::Register(ParseOptions *po) {
+  po->Register("cuda-cudnn-conv-algo-search", &cudnn_conv_algo_search,
+          "CuDNN convolution algrorithm search");
+}
+
+bool CudaConfig::Validate() const {
+  if (cudnn_conv_algo_search < 1 && cudnn_conv_algo_search > 3) {
```
csukuangfj:
```suggestion
  if (cudnn_conv_algo_search < 1 || cudnn_conv_algo_search > 3) {
```

sherpa-onnx/csrc/provider-config.cc
```diff
@@ -0,0 +1,155 @@
...
+    return false;
+  }
+  return true;
+}
+
+std::string CudaConfig::ToString() const {
+  std::ostringstream os;
+
+  os << "CudaConfig(";
+  os << "cudnn_conv_algo_search=\"" << cudnn_conv_algo_search << ")";
```
csukuangfj:
```suggestion
  os << "cudnn_conv_algo_search=" << cudnn_conv_algo_search << ")";
```

sherpa-onnx/csrc/provider-config.cc
```diff
@@ -0,0 +1,155 @@
...
+              "true to enable timing cache");
+  po->Register("trt-timing-cache-path", &trt_timing_cache_path,
+              "");
+  po->Register("trt-dump-subgraphs", &trt_dump_subgraphs,
+              "true to dump subgraphs");
+}
+
+bool TensorrtConfig::Validate() const {
+  if (trt_max_workspace_size > 0) {
+    SHERPA_ONNX_LOGE("trt_max_workspace_size: '%d' is not valid.",
```
csukuangfj:
```suggestion
    SHERPA_ONNX_LOGE("trt_max_workspace_size: '%u' is not valid.",
```

Use `%u` for unsigned integers.

sherpa-onnx/csrc/provider-config.cc
```diff
@@ -0,0 +1,155 @@
...
+    SHERPA_ONNX_LOGE("trt_max_partition_iterations: '%d' is not valid.",
+        trt_max_partition_iterations);
+    return false;
+  }
+  if (trt_min_subgraph_size > 0) {
+    SHERPA_ONNX_LOGE("trt_min_subgraph_size: '%d' is not valid.",
+        trt_min_subgraph_size);
+    return false;
+  }
+  if (trt_fp16_enable != true || trt_fp16_enable != false) {
```
csukuangfj: What is the point of the test here?
A bool is either true or false. There is not a third possibility.

This test always fails. Have you tested the code locally with tensorrt?

csukuangfj (changes requested):

sherpa-onnx/csrc/provider-config.cc
manickavela29: I havn't tested the params recently, I will check and get back

sherpa-onnx/csrc/provider-config.cc
```diff
@@ -60,7 +60,7 @@ void TensorrtConfig::Register(ParseOptions *po) {
 
 bool TensorrtConfig::Validate() const {
   if (trt_max_workspace_size > 0) {
-    SHERPA_ONNX_LOGE("trt_max_workspace_size: '%d' is not valid.",
+    SHERPA_ONNX_LOGE("trt_max_workspace_size: '%u' is not valid.",
```
csukuangfj: Please fix the remaining similar errors, not only this one.

sherpa-onnx/csrc/provider-config.cc
manickavela29: Apologies, Seemed to missed few commits from local.
Now it is sorted,
But the intention of _if(true||false)_ here is to validate bool or some other integer

manickavela29: > What is the point of the test here? A bool is either true or false. There is not a third possibility.
> 
> This test always fails. Have you tested the code locally with tensorrt?

It was always failing because trt_config.Register(po) was missing, I think the error would have been for _not a valida option_ , if that wasn't the case, I will debug further

manickavela29:
Hi @csukuangfj,
Let me know if there is any more suggestions

sherpa-onnx/csrc/provider-config.cc
```diff
@@ -0,0 +1,158 @@
...
+  po->Register("trt-timing-cache-enable", &trt_timing_cache_enable,
+              "Enable use of timing cache to speed up builds.");
+  po->Register("trt-timing-cache-path", &trt_timing_cache_path,
+              "Set path for storing timing cache.");
+  po->Register("trt-dump-subgraphs", &trt_dump_subgraphs,
+              "Dump optimized subgraphs for debugging.");
+}
+
+bool TensorrtConfig::Validate() const {
+  if (trt_max_workspace_size < 0) {
```
csukuangfj, line 62: An unsigned integer is never less than 0.

Please fix it.

csukuangfj: I think now you see why unsigned integers are error-prone.

csukuangfj (changes requested):

sherpa-onnx/csrc/provider-config.cc
```diff
@@ -0,0 +1,158 @@
...
+    SHERPA_ONNX_LOGE("trt_max_partition_iterations: '%u' is not valid.",
+        trt_max_partition_iterations);
+    return false;
+  }
+  if (trt_min_subgraph_size < 0) {
+    SHERPA_ONNX_LOGE("trt_min_subgraph_size: '%u' is not valid.",
+        trt_min_subgraph_size);
+    return false;
+  }
+  if (trt_fp16_enable != true || trt_fp16_enable != false) {
```
csukuangfj: This check does not make sense for a `bool` variable. Please remove this check.

csukuangfj: Also, please remove all checks about bool variables.

sherpa-onnx/csrc/provider-config.cc
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  return true;
+}
+
+std::string TensorrtConfig::ToString() const {
+  std::ostringstream os;
+
+  os << "TensorrtConfig(";
+  os << "trt_max_workspace_size=\"" << trt_max_workspace_size << "\", ";
```
csukuangfj:
```suggestion
  os << "trt_max_workspace_size=" << trt_max_workspace_size << ", ";
```

You don't need to use `""` to enclose an integer. Please fix similar issues below.

sherpa-onnx/csrc/provider-config.cc
```diff
@@ -0,0 +1,158 @@
...
+  return true;
+}
+
+std::string ProviderConfig::ToString() const {
+  std::ostringstream os;
+
+  os << "ProviderConfig(";
+  os << "device=\"" << device << "\", ";
+  os << "provider=\"" << provider << "\", ";
+  os << "cuda_config=\"" << cuda_config.ToString() << "\", ";
```
csukuangfj:
```suggestion
  os << "cuda_config=" << cuda_config.ToString() << ", ";
```

Note that `""` is needed only for string variables. Please fix similar issues in your PR.

sherpa-onnx/csrc/provider-config.h
```diff
@@ -0,0 +1,93 @@
...
+  std::string provider = "cpu";
+  int32_t device = 0;
+  // device only used for cuda and trt
+
+  ProviderConfig() = default;
+  ProviderConfig(const TensorrtConfig &trt_config,
+                const CudaConfig &cuda_config,
+                const std::string &provider,
+                int32_t device)
+      : device(device), provider(provider),
```
csukuangfj: Please initialize member variables in order by following the definition order.

That is, initialize `trt_config` first, and then `cuda_config`, etc.

sherpa-onnx/csrc/provider-config.h
```diff
@@ -0,0 +1,93 @@
...
+  bool Validate() const;
+
+  std::string ToString() const;
+};
+
+struct TensorrtConfig {
+  uint32_t trt_max_workspace_size = 2147483648;
+  uint32_t trt_max_partition_iterations = 10;
+  uint32_t trt_min_subgraph_size = 5;
+  bool trt_fp16_enable = 1;
```
csukuangfj: Please use `true` or `false` for `bool` variables.

sherpa-onnx/csrc/provider-config.cc
manickavela29: Shall I remove the conditional check?
I agree that this check is useless for now.

But I am planning to run some tests later and workspace values might go even beyond uint32, I will make it long/long long maybe in sometime

sherpa-onnx/csrc/session.cc
```diff
@@ -64,26 +66,51 @@ static Ort::SessionOptions GetSessionOptionsImpl(int32_t num_threads,
...
+      }
+      auto trt_config = provider_config->trt_config;
       struct TrtPairs {
         const char *op_keys;
         const char *op_values;
       };
 
+      auto device_id = std::to_string(provider_config->device);
+      auto trt_max_workspace_size =
+          std::to_string(trt_config.trt_max_workspace_size);
```
csukuangfj, line 82: Since you are converting integers to strings, please replace all
`uint32_t` with `int64_t`. Remember unsigned integers are error-prone. Please avoid using them if possible.

csukuangfj (changes requested):

csukuangfj:
By the way, could you fix the python tests. Please see

https://github.com/k2-fsa/sherpa-onnx/actions/runs/9689759305/job/26738528427?pr=992

sherpa-onnx/csrc/session.cc
manickavela29: I will make all the uint32 to int32, but for int64_t, SHERPA_ONNX_LOGE() format will be like below

#include <inttypes.h>
int64_t t;
printf("%" PRId64 "\n", t);

ref : https://stackoverflow.com/questions/9225567/how-to-portably-print-a-int64-t-type-in-c
let me know if there is any better suggestions

will raise the patch in sometime

manickavela29:
I ran the python tests and got this error

> + python3 ./python-api-examples/offline-telespeech-ctc-decode-files.py
> Traceback (most recent call last):
>   File "/mnt/efs/manickavela/asr_work/trt/sherpa-onnx/./python-api-examples/offline-telespeech-ctc-decode-files.py", line 16, in <module>
>     import sherpa_onnx
>   File "/mnt/efs/manickavela/miniconda3/envs/icefallmk/lib/python3.9/site-packages/sherpa_onnx/__init__.py", line 1, in <module>
>     from _sherpa_onnx import (
> ImportError: arg(): could not convert default argument 'trt_config: sherpa_onnx::TensorrtConfig' in method '<class '_sherpa_onnx.ProviderConfig'>.__init__' in
> to a Python object (type not registered yet?)

it would be great to have some hints for fixing this

manickavela29:
Hi @csukuangfj, would be great to have some leads on python fix

sherpa-onnx/python/csrc/sherpa-onnx.cc
```diff
@@ -51,7 +54,7 @@ PYBIND11_MODULE(_sherpa_onnx, m) {
   PybindEndpoint(&m);
   PybindOnlineRecognizer(&m);
   PybindKeywordSpotter(&m);
-
+  // PybindProviderConfig(&m);
```
csukuangfj: please change it to
```c++

PybindCudaConfig(&m);
PybindTensorrtConfig(&m);
PybindProviderConfig(&m);
```

You need to add the 3 corresponding header files.

sherpa-onnx/python/csrc/provider-config.cc
```diff
@@ -0,0 +1,35 @@
...
+void PybindProviderConfig(py::module *m) {
+  // PybindCudaConfig(m);
+  // PybindTensorrtConfig(m);
+
+  using PyClass = ProviderConfig;
+  py::class_<PyClass>(*m, "ProviderConfig")
+      .def(py::init<>())
+      .def(py::init<const TensorrtConfig &, const CudaConfig &,
+          const std::string &, int32_t>(),
+           py::arg("trt_config"), py::arg("cuda_config"),
```
csukuangfj:
```suggestion
           py::arg("trt_config") = TensorrtConfig{}, py::arg("cuda_config") = {},
```

csukuangfj (changes requested):

csukuangfj:
please also change
https://github.com/k2-fsa/sherpa-onnx/blob/master/sherpa-onnx/python/sherpa_onnx/online_recognizer.py#L174

I give you an example below. You need to change other `def from_xxx()` accordingly.

Change
https://github.com/k2-fsa/sherpa-onnx/blob/125bb9ff999aeed25924fc4e269ac02450bd7afa/sherpa-onnx/python/sherpa_onnx/online_recognizer.py#L174-L183
to
```python3
        provider_config = ProviderConfig(provider=provider)

        model_config = OnlineModelConfig(
            transducer=transducer_config,
            tokens=tokens,
            num_threads=num_threads,
            provider_config=provider_config,
            model_type=model_type,
            modeling_unit=modeling_unit,
            bpe_vocab=bpe_vocab,
            debug=debug,
        )
```

Remember to change

https://github.com/k2-fsa/sherpa-onnx/blob/125bb9ff999aeed25924fc4e269ac02450bd7afa/sherpa-onnx/python/sherpa_onnx/online_recognizer.py#L9

```python3
 ProviderConfig,
 OnlineModelConfig, 
```

sherpa-onnx/python/csrc/tensorrt-config.cc
```diff
@@ -0,0 +1,72 @@
...
+            ans->trt_detailed_build_log = trt_detailed_build_log;
+            ans->trt_engine_cache_enable = trt_engine_cache_enable;
+            ans->trt_timing_cache_enable = trt_timing_cache_enable;
+            ans->trt_engine_cache_path = trt_engine_cache_path;
+            ans->trt_timing_cache_path = trt_timing_cache_path;
+            ans->trt_dump_subgraphs = trt_dump_subgraphs;
+
+            return ans;
+          }),
+           py::arg("trt_max_workspace_size") = 2147483648,
```
csukuangfj: Please use
2147483647
or
0x7fffffff

sherpa-onnx/python/sherpa_onnx/online_recognizer.py
```diff
@@ -66,6 +68,19 @@ def from_transducer(
         debug: bool = False,
         rule_fsts: str = "",
         rule_fars: str = "",
+        provider: str = "cpu",
+        device: int = 0,
+        cudnn_conv_algo_search: int = 1,
+        trt_max_workspace_size: int = 2147483648,
```
csukuangfj:
```suggestion
        trt_max_workspace_size: int = 2147483647,
```

csukuangfj (changes requested):

sherpa-onnx/python/csrc/tensorrt-config.cc
manickavela29: sure alright!

manickavela29: Yes the issue is fixed. Thanks!

sherpa-onnx/python/sherpa_onnx/keyword_spotter.py
```diff
@@ -41,6 +42,7 @@ def __init__(
         keywords_threshold: float = 0.25,
         num_trailing_blanks: int = 1,
         provider: str = "cpu",
+        device: int = 1,
```
csukuangfj: Could you use 0 as the default device?

manickavela29:
Hi @csukuangfj,
would be great if you can have one more review by today. 🙂

sherpa-onnx/csrc/provider-config.cc
```diff
@@ -0,0 +1,134 @@
...
+  cuda_config.Register(po);
+  trt_config.Register(po);
+
+  po->Register("device", &device, "GPU device index for CUDA and Trt EP");
+  po->Register("provider", &provider,
+               "Specify a provider to use: cpu, cuda, coreml");
+}
+
+bool ProviderConfig::Validate() const {
+  if (device < 0) {
```
csukuangfj, line 116: Do you need to validate tensorrt config or cuda config when the provider is not CPU?

sherpa-onnx/python/csrc/sherpa-onnx.h
```diff
@@ -5,6 +5,8 @@
 #ifndef SHERPA_ONNX_PYTHON_CSRC_SHERPA_ONNX_H_
 #define SHERPA_ONNX_PYTHON_CSRC_SHERPA_ONNX_H_
 
+#define PYBIND11_DETAILED_ERROR_MESSAGES
```
csukuangfj: Could you remove it after debugging since it would increase the library file size?

sherpa-onnx/python/sherpa_onnx/__init__.py
```diff
@@ -16,6 +16,7 @@
     OfflineTtsVitsModelConfig,
     OfflineZipformerAudioTaggingModelConfig,
     OnlineStream,
+    ProviderConfig,
```
csukuangfj: Could you remove it?

I think it is not used outside of online_recognizer.py and keyword_spotter.py.

csukuangfj (commented):
Thank you!

Left some minor comments. Otherwise, it looks good to me.

sherpa-onnx/python/csrc/sherpa-onnx.h
manickavela29: yes, I will remove this. Thanks

sherpa-onnx/python/sherpa_onnx/__init__.py
manickavela29: yes I think keyword spotter was not able to import it, without adding providerconfig at __init__ i think,
Not very sure though, I will remove it, I the tests fails later, I will add it back

sherpa-onnx/csrc/provider-config.cc
manickavela29: will add a if around it to check cuda/trt

sherpa-onnx/csrc/provider-config.cc
```diff
@@ -113,9 +113,11 @@ void ProviderConfig::Register(ParseOptions *po) {
 }
 
 bool ProviderConfig::Validate() const {
-  if (device < 0) {
-    SHERPA_ONNX_LOGE("device: '%d' is invalid.", device);
-    return false;
+  if(provider == "cuda" || provider == "trt") {
```
csukuangfj: I mean, do you need to implement:
```
if provider == "cuda" and not cuda_config.Validate():
  return false

if provider == 'trt' and not trt_config.Validate():
  return false
```

(Note it's just pseudo code. I think you get the ideas.)

manickavela29: oh yes sorry,
literally I am not thinking through anymore 😄 , please don't mind
pybind took a good score on me 😮‍💨

manickavela29:
Hi @csukuangfj,

I think it is all good now, expected tests are also passing through,
let me know if there are any more reviews

csukuangfj (approved):

manickavela29:
uffff, Thank you so much

PR #1005, opened 2024-06-14
offline transducer: treat unk as blank

Chung-I:
Hi there, thanks for implementing this great project!

I saw that treating unk as blank is implemented in online transducer, but not in offline transducer, so I made some changes to offline transducer to match the behavior of online transducer.

Please let me know if there's anything else missing.

Thanks in advance!


csukuangfj:

Will merge once the CI passes.

PR #1006, opened 2024-06-14
scale value fix

lovemefan:
replace 372768 with 32768


csukuangfj:
Thank you  for your contribution!

PR #1058, opened 2024-06-25
Resolve issues with using pre-installed onnxruntime

hantengc:
There is an issue with the directory, ${location_onnxruntime_header_dir} may be NOT-FOUND


csukuangfj:
Could you append the two paths rather than to replace the existing two?

hantengc:
> Could you append the two paths rather than to replace the existing two?

额我知道你是为了保守起见，避免某些情况下是在/usr/include或者/usr/local/include目录下存在的正确的文件，但是我看了onnxruntime的文件一般都是保存在/usr/include/onnxruntime或者/usr/local/include/onnxruntime。

我们是不是直接按照规范来而不是附加一个路径？

csukuangfj:

PR #1089, opened 2024-07-08
parse option int64_t

manickavela29:
Adding parse option for int64_t
Will be useful in increasing the trt_workspace_size config


csukuangfj:

PR #1094, opened 2024-07-08
updating trt workspace int64

manickavela29:
trt_workspace range can be extended further with this,
and will not be limited to just 2GB default value


csukuangfj (approved):

Issue #1096, opened 2024-07-09
max speakers for speaker embedding manager


thewh1teagle:
It would be useful if I there was an option in  embedding manager to find the closest one from existing speakers.
This way, I can handle myself the case where I know how many speakers there are, and there's already enough speakers detected.

csukuangfj:
That sounds reasonable.

Could you follow
https://github.com/k2-fsa/sherpa-onnx/blob/3e4307e2fb88d4b1b648211c14f2fff6db11bca4/sherpa-onnx/csrc/speaker-embedding-manager.cc#L126
to add a `TopK` to return the name and scores for the topK match?

thewh1teagle:
> to add a `TopK` to return the name and scores for the topK match?

I assume we just need to sort the scores, then iterate through them, collect, and return. Could you share any specific IntelliSense settings used in the repository so I can feel more comfortable working in VSCode?

csukuangfj:
We are using `clang-format`, which can be installed with
```
pip install clang-format
```
I don't use VSCode. Maybe you can find a way to integrate clang-format with it.

You don't need to care about the style issues. We can reformat the file later.

PR #1102, opened 2024-07-10
feat: find best embedding matches

thewh1teagle:
Resolve https://github.com/k2-fsa/sherpa-onnx/issues/1096

### DIFF
--- a/sherpa-onnx/c-api/c-api.h
+++ b/sherpa-onnx/c-api/c-api.h
@@ -1116,7 +1116,7 @@ SHERPA_ONNX_API typedef struct SherpaOnnxSpeakerEmbeddingManagerSpeakerMatch {
 
 SHERPA_ONNX_API typedef struct
     SherpaOnnxSpeakerEmbeddingManagerBestMatchesResult {
-  SherpaOnnxSpeakerEmbeddingManagerSpeakerMatch *matches;
+  const SherpaOnnxSpeakerEmbeddingManagerSpeakerMatch *matches;
   int32_t count;
 } SherpaOnnxSpeakerEmbeddingManagerBestMatchesResult;
 
--- a/sherpa-onnx/c-api/c-api.h
+++ b/sherpa-onnx/c-api/c-api.h
@@ -1109,6 +1109,39 @@ SHERPA_ONNX_API const char *SherpaOnnxSpeakerEmbeddingManagerSearch(
 SHERPA_ONNX_API void SherpaOnnxSpeakerEmbeddingManagerFreeSearch(
     const char *name);
 
+SHERPA_ONNX_API typedef struct SherpaOnnxSpeakerEmbeddingManagerSpeakerMatch {
+  float score;
+  char *name;
+} SherpaOnnxSpeakerEmbeddingManagerSpeakerMatch;
+
+SHERPA_ONNX_API typedef struct
+    SherpaOnnxSpeakerEmbeddingManagerBestMatchesResult {
+  SherpaOnnxSpeakerEmbeddingManagerSpeakerMatch *matches;
+  int32_t count;
+} SherpaOnnxSpeakerEmbeddingManagerBestMatchesResult;
+
+// Get the best matching speakers whose embeddings match the given
+// embedding.
+//
+// @param p Pointer to the SherpaOnnxSpeakerEmbeddingManager instance.
+// @param v Pointer to an array containing the embedding vector.
+// @param threshold Minimum similarity score required for a match (between 0 and
+// 1).
+// @param n Number of best matches to retrieve.
+// @return Returns a pointer to
+// SherpaOnnxSpeakerEmbeddingManagerBestMatchesResult
+//         containing the best matches found. Returns NULL if no matches are
+//         found. The caller is responsible for freeing the returned pointer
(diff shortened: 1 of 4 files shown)


sherpa-onnx/c-api/c-api.h
```diff
@@ -1109,6 +1109,39 @@ SHERPA_ONNX_API const char *SherpaOnnxSpeakerEmbeddingManagerSearch(
 SHERPA_ONNX_API void SherpaOnnxSpeakerEmbeddingManagerFreeSearch(
     const char *name);
 
+SHERPA_ONNX_API typedef struct SherpaOnnxSpeakerEmbeddingManagerSpeakerMatch {
+  float score;
+  char *name;
```
csukuangfj:
```suggestion
  const char *name;
```

Add `const` to indicate that users should not modify it.

sherpa-onnx/c-api/c-api.h
```diff
@@ -1109,6 +1109,39 @@ SHERPA_ONNX_API const char *SherpaOnnxSpeakerEmbeddingManagerSearch(
...
+} SherpaOnnxSpeakerEmbeddingManagerSpeakerMatch;
+
+SHERPA_ONNX_API typedef struct
+    SherpaOnnxSpeakerEmbeddingManagerBestMatchesResult {
+  SherpaOnnxSpeakerEmbeddingManagerSpeakerMatch *matches;
```
csukuangfj:
```suggestion
  const SherpaOnnxSpeakerEmbeddingManagerSpeakerMatch *matches;
```

sherpa-onnx/csrc/speaker-embedding-manager.cc
```diff
@@ -131,6 +131,39 @@ class SpeakerEmbeddingManager::Impl {
...
+    for (int i = 0; i < scores.size(); ++i) {
+      if (scores[i] >= threshold) {
+        score_indices.emplace_back(scores[i], i);
+      }
+    }
+
+    std::sort(score_indices.rbegin(), score_indices.rend(),
+              [](const auto &a, const auto &b) { return a.first < b.first; });
+
+    for (int i = 0; i < std::min(n, static_cast<int32_t>(score_indices.size()));
```
csukuangfj, line 159:
```suggestion
    matches.reserve(score_indices.size());
    for (int i = 0; i < std::min(n, static_cast<int32_t>(score_indices.size()));
```

It is good to pre-allocate space.

csukuangfj (commented):
Thanks! Left some minor comments. Otherwise, it looks good to me.

sherpa-onnx/c-api/c-api.h
thewh1teagle: When using const there I get the following compile error

```console
/Users/user/Documents/sherpa-rs/sys/sherpa-onnx/sherpa-onnx/c-api/c-api.cc:1275:30: error: read-only variable is not assignable
    result->matches[i].score = matches[i].score;
    ~~~~~~~~~~~~~~~~~~~~~~~~ ^
/Users/user/Documents/sherpa-rs/sys/sherpa-onnx/sherpa-onnx/c-api/c-api.cc:1281:29: error: read-only variable is not assignable
    result->matches[i].name = name;
    ~~~~~~~~~~~~~~~~~~~~~~~ ^
```

I think we'll need to use internal non const in the function and case it to const before return

sherpa-onnx/csrc/speaker-embedding-manager.cc
thewh1teagle: looks better! Thanks :)

sherpa-onnx/c-api/c-api.h
thewh1teagle: How about


```cpp
const SherpaOnnxSpeakerEmbeddingManagerBestMatchesResult *
SherpaOnnxSpeakerEmbeddingManagerGetBestMatches(
    const SherpaOnnxSpeakerEmbeddingManager *p, const float *v, float threshold,
    int32_t n) {
  auto matches = p->impl->GetBestMatches(v, threshold, n);

  if (matches.empty()) {
    return nullptr;
  }

  auto *result =
      static_cast<SherpaOnnxSpeakerEmbeddingManagerBestMatchesResult *>(
          operator new(
              sizeof(SherpaOnnxSpeakerEmbeddingManagerBestMatchesResult)));
  auto *matchesArray =
      new SherpaOnnxSpeakerEmbeddingManagerSpeakerMatch[matches.size()];

  for (int i = 0; i < matches.size(); ++i) {
    matchesArray[i].score = matches[i].score;

    char *name = new char[matches[i].name.size() + 1];
    std::copy(matches[i].name.begin(), matches[i].name.end(), name);
    name[matches[i].name.size()] = '\0';

    matchesArray[i].name = name;
  }
  new (result) SherpaOnnxSpeakerEmbeddingManagerBestMatchesResult{
      matchesArray, static_cast<int32_t>(matches.size())};

  return result;
}
```

thewh1teagle: Fixed in https://github.com/k2-fsa/sherpa-onnx/pull/1102/commits/7d2341d1fe1ca45a2890a554fe2dc5360d6f15da

csukuangfj:

PR #1156, opened 2024-07-20
feat: add stt c api example

thewh1teagle:


c-api-examples/offline-stt-c-api.c
```diff
@@ -0,0 +1,98 @@
...
+  // Whisper config
+  SherpaOnnxOfflineWhisperModelConfig whisper_config;
+  whisper_config.decoder = decoder_filename;
+  whisper_config.encoder = encoder_filename;
+  whisper_config.language = language;
+  whisper_config.tail_paddings = 0;
+  whisper_config.task = "transcribe";
+
+  // Offline model config
+  SherpaOnnxOfflineModelConfig offline_model_config;
```
csukuangfj, line 49: Could you use `memset` to zero-intiialize the struct?

You can have a look at how we initialize struct in other C API examples.

thewh1teagle: Added

csukuangfj:

PR #1179, opened 2024-07-28
Add more Python examples for SenseVoice

csukuangfj:


csukuangfj:
The doc is available at 
https://k2-fsa.github.io/sherpa/onnx/sense-voice/python-api.html#

PR #133, opened 2023-04-28
Add lm rescore to online-modified-beam-search

pingfengluo:
**Usage of this PR**

*get RNNT model and RNN LM*
```
GIT_LFS_SKIP_SMUDGE=1 git clone https://huggingface.co/csukuangfj/sherpa-onnx-streaming-zipformer-bilingual-zh-en-2023-02-20
cd sherpa-onnx-streaming-zipformer-bilingual-zh-en-2023-02-20
git lfs pull --include "*.onnx"
cd -
```
*use modified_beam_search with RNN LM*
```
cd sherpa-onnx/build && cmake .. && make  && ln -sf ../../sherpa-onnx-streaming-zipformer-bilingual-zh-en-2023-02-20 exp
./bin/sherpa-onnx exp/tokens.txt exp/encoder-epoch-99-avg-1.int8.onnx exp/decoder-epoch-99-avg-1.onnx exp/joiner-epoch-99-avg-1.int8.onnx exp/test_wavs/8k.wav 2 "modified_beam_search" exp/with-state-epoch-99-avg-1.int8.onnx 
```

PR #1231, opened 2024-08-07
Re-implement LM rescore for online transducer

SilverSulfide:
Shallow fusion can be too slow for online cpu inference. Added an option to use classical LM rescore instead.

- Rescore implementation based on https://github.com/k2-fsa/sherpa-onnx/pull/133
- Shallow fusion enabled by default
- Pass _lm-shallow-fusion=false_ to enable rescore instead
- Updated online Python API

CPU runtime comparison for ~1 min wav file using the default _--lm-num-threads=1_

| Method              | Runtime (s) | RTF |
| :---------------- | :------: | :----: |
| No LM | 5.7  | 0.098 |
| Rescore | 40 | 0.68 |
| Shallow Fusion | 69 | 1.2 |

### DIFF
--- a/sherpa-onnx/csrc/online-lm-config.cc
+++ b/sherpa-onnx/csrc/online-lm-config.cc
@@ -18,6 +18,8 @@ void OnlineLMConfig::Register(ParseOptions *po) {
                "Number of threads to run the neural network of LM model");
   po->Register("lm-provider", &lm_provider,
                "Specify a provider to LM model use: cpu, cuda, coreml");
+  po->Register("lm-shallow-fusion", &shallow_fusion,
+               "Boolean whether to use shallow fusion or rescore.");
 }
 
 bool OnlineLMConfig::Validate() const {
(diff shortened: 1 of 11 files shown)


sherpa-onnx/csrc/online-lm-config.cc
```diff
@@ -34,7 +36,8 @@ std::string OnlineLMConfig::ToString() const {
 
   os << "OnlineLMConfig(";
   os << "model=\"" << model << "\", ";
-  os << "scale=" << scale << ")";
+  os << "scale=" << scale << ", ";
+  os << "shallow_fusion=\"" << (shallow_fusion ? "True" : "False") << "\")";
```
csukuangfj:
```suggestion
  os << "shallow_fusion=" << (shallow_fusion ? "True" : "False") << ")";
```

We use `""` only for string variables.

sherpa-onnx/python/sherpa_onnx/online_recognizer.py
```diff
@@ -274,6 +275,7 @@ def from_transducer(
         lm_config = OnlineLMConfig(
             model=lm,
             scale=lm_scale,
```
csukuangfj, line 277: Please also update
https://github.com/k2-fsa/sherpa-onnx/blob/5f9441bfdb2cc4d2a6076ac8be5e37a59f85f9fe/sherpa-onnx/python/csrc/online-lm-config.cc#L18

csukuangfj (changes requested):

SilverSulfide:
@csukuangfj hi, are there any more changes that should be written?

csukuangfj:
Could you fix the CI errors for Python?

SilverSulfide:
@csukuangfj should be resolved now!

PR #1232, opened 2024-08-07
addition of blank_penalty config in online_recognizer

parth-khiera:
Added support for the blank_penalty config parameter in the  Go module to match the existing Python implementation.


csukuangfj:

I will take care of the remaining work and will release a new version today.

PR #1253, opened 2024-08-13
Update offline-recognizer.cc

iprovalo:
Adding setConfig method to JNI to support setting a config on the previously initialized offline-recognizer.


csukuangfj (approved):

PR #1275, opened 2024-08-22
update kotlin api for better release native object and add user-frien…

fbzhong:
update kotlin api for better release native object and add user-friendly apis


csukuangfj:
Looks great! Thank you for your contribution!

PR #1287, opened 2024-08-24
update generate-asset-list.py

hantrungkien:
Update the `generate-asset-list.py` script example to include all assets under a directory.


csukuangfj:

Issue #1227, opened 2024-08-06
[Help wanted] Add various language bindings for online punctuation models


csukuangfj:
https://github.com/k2-fsa/sherpa-onnx/pull/1224
added support for online punctuation models. However, only C++ is supported.

Help from the community is appreciated to support the following languages:
- [ ] 1. C
- [ ] 2. Python
- [ ] 3. Java
- [ ] 4. JavaScript
- [ ] 5. C#
- [ ] 6. Go
- [ ] 7. Kotlin
- [ ] 8. Dart
- [ ] 9. Swift
- [ ] 10. Rust

We also need to create examples for the above languages.

Note you can create a separate PR for each language. 

This is a pretty good exercise for learning sherpa-onnx. We can give you any help you need to get you started.

Caet-pip:
I am interested to help with Python, what needs to be done?

csukuangfj:
> I am interested to help with Python, what needs to be done?

Thanks! The Python part has been contributed by the community.

PR #1312, opened 2024-08-30
Add Python binding for online punctuation models

yaochie:
For #1227.

I renamed the parameter for `OnlinePunctuationConfig` to `model_config`, let me know if that's ok, or whether I should follow #762 and keep it as `model`.


csukuangfj:

PR #1323, opened 2024-09-05
Throw error instead exit on fail to read wav in java

RGdevz:
I think it's better to throw a error instead of exit to let the developer handle it in the app


csukuangfj:

PR #1324, opened 2024-09-06
Fixed the C api calls and created the TTS project file

twodawg:
The dot net examples contained an orphaned project that seemed to use an older version of the C api. I added a project file for TTS and fixed the C api calls.


csukuangfj:

PR #1325, opened 2024-09-07
    Added an alternative way to initialize/load some models (for non-…

shawl336:
Added an alternative way to initialize/load some models (for non-Android codes), tokens, hotwrods and keywords, that initializing/loading from memory buffers. The original usage of model initialization unchanged.
Basically, I just added additional "const char* $_buf_begin, const char* $_buf_end" vars ($ == encoder/decoder/joiner/model), if they are non-nullptr, sherpa will attempt to init the models from the memory starting from "$_buf_start" to "$_buf_end" in priority to the original filename strings.
Also, I added "tokens_buf_str, hotwords_buf_str and keywords_buf_str", if they are non-empty, sherpa will attempt to treat them as the content directly, and using istringstream rather than iftream.
the supported models are:
    online models:
      OnlineTransducerModel
      OnlineParaformerModel
      OnlineWenetCtcModel
      OnlineZipformer2CtcModel
      OnlineNeMoCtcModel

    offline models:
      OfflineTransducerModel
      OfflineParaformerModel
      OfflineNemoEncDecCtcModel
      OfflineWhisperModel
      OfflineTdnnModel
      OfflineZipformerCtcModel
      OfflineWenetCtcModel
      OfflineSenseVoiceModel

    keywordspotter models:
      KeywordSpotterTransducer

### DIFF
--- a/sherpa-onnx/csrc/hypothesis.h
+++ b/sherpa-onnx/csrc/hypothesis.h
@@ -51,8 +51,13 @@ struct Hypothesis {
   // LM log prob if any.
   double lm_log_prob = 0;
 
-  // the nn lm score for next token given the current ys
+  // the nn lm score for next token given the current ys,
+  // when using shallow fusion
   CopyableOrtValue nn_lm_scores;
+
+  // cur scored tokens by RNN LM, when rescoring
+  int32_t cur_scored_pos = 0;
+
   // the nn lm states
   std::vector<CopyableOrtValue> nn_lm_states;
 
--- a/sherpa-onnx/csrc/online-lm-config.cc
+++ b/sherpa-onnx/csrc/online-lm-config.cc
@@ -18,6 +18,8 @@ void OnlineLMConfig::Register(ParseOptions *po) {
                "Number of threads to run the neural network of LM model");
   po->Register("lm-provider", &lm_provider,
                "Specify a provider to LM model use: cpu, cuda, coreml");
+  po->Register("lm-shallow-fusion", &shallow_fusion,
+               "Boolean whether to use shallow fusion or rescore.");
 }
 
 bool OnlineLMConfig::Validate() const {
@@ -34,7 +36,8 @@ std::string OnlineLMConfig::ToString() const {
 
   os << "OnlineLMConfig(";
   os << "model=\"" << model << "\", ";
-  os << "scale=" << scale << ")";
+  os << "scale=" << scale << ", ";
+  os << "shallow_fusion=" << (shallow_fusion ? "True" : "False") << ")";
 
   return os.str();
 }
--- a/sherpa-onnx/csrc/online-lm-config.h
+++ b/sherpa-onnx/csrc/online-lm-config.h
@@ -18,15 +18,18 @@ struct OnlineLMConfig {
   float scale = 0.5;
   int32_t lm_num_threads = 1;
   std::string lm_provider = "cpu";
+  // enable shallow fusion
+  bool shallow_fusion = true;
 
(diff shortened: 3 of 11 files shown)


csukuangfj:
Could you also add tests for your changes?

sherpa-onnx/c-api/c-api.h
```diff
@@ -54,30 +54,37 @@ extern "C" {
 /// from there.
 SHERPA_ONNX_API typedef struct SherpaOnnxOnlineTransducerModelConfig {
   const char *encoder;
+  const char *encoder_buf_begin, *encoder_buf_end;  // if non-null, loading the encoder from the buffer in prioriy
   const char *decoder;
+  const char *decoder_buf_begin, *decoder_buf_end;  // if non-null, loading the decoder from the buffer in prioriy
   const char *joiner;
+  const char *joiner_buf_begin, *joiner_buf_end;  // if non-null, loading the joiner from the buffer in prioriy
```
csukuangfj: Please add new fields to the end of existing fields.

Please don't change the order of exsiting fields.

csukuangfj (changes requested):

sherpa-onnx/csrc/online-transducer-model-config.h
```diff
@@ -12,15 +12,25 @@ namespace sherpa_onnx {
 
 struct OnlineTransducerModelConfig {
   std::string encoder;
+  const char *encoder_buf_begin, *encoder_buf_end;
```
csukuangfj:
```suggestion
  const char *encoder_buf_begin = nullptr;
  const char *encoder_buf_end = nullptr;
```

1. Please follow our existing code style to define one variable per line.
2. Please initialize it to `nullptr`.

csukuangfj:
By the way, I suggest that you change one model config per pull-request and add test for the changes.

Otherwise, the pull-request is very large and you need to write lots of tests to cover your changes.

sherpa-onnx/c-api/c-api.cc
```diff
@@ -58,21 +58,51 @@ SherpaOnnxOnlineRecognizer *SherpaOnnxCreateOnlineRecognizer(
 
   recognizer_config.model_config.transducer.encoder =
       SHERPA_ONNX_OR(config->model_config.transducer.encoder, "");
+  recognizer_config.model_config.transducer.encoder_buf_begin =
+      SHERPA_ONNX_OR(config->model_config.transducer.encoder_buf_begin, nullptr);
```
csukuangfj:
```suggestion
  recognizer_config.model_config.transducer.encoder_buf_begin =
      config->model_config.transducer.encoder_buf_begin;
```

And please change other places.

sherpa-onnx/c-api/c-api.h
csukuangfj: I suggest that you use
```
const char *encoder_buf;
int32_t encoder_buf_len;
```

sherpa-onnx/csrc/offline-nemo-enc-dec-ctc-model-config.h
```diff
@@ -12,11 +12,15 @@ namespace sherpa_onnx {
 
 struct OfflineNemoEncDecCtcModelConfig {
   std::string model;
+  const char *model_buf_begin, *model_buf_end;
```
csukuangfj: Please update the `Validate()` method to check that
when `model_buf_begin` is not nullptr, then model must be empty and vice versa.

csukuangfj (changes requested):

PR #1335, opened 2024-09-10
Preserve previous result as context for next segment

vsd-vector:
In online transducer recognizer during reset() sherpa-onnx preserves the output of the decoder network (decoder_output), but resets model context to sequence of blanks.

```cpp
decoder_->UpdateDecoderOut(&s->GetResult());
Ort::Value decoder_out = std::move(s->GetResult().decoder_out);

auto r = decoder_->GetEmptyResult(); # initialized with empty hyp containing blanks
...
s->SetResult(r);
s->GetResult().decoder_out = std::move(decoder_out);
```

After the reset, at t0 of the new segment the beam-search-decoder reuses decoder_output. 
```cpp
if (t == 0) {
    UseCachedDecoderOut(hyps_row_splits, *result, &decoder_out);
}
```

Let's assume it outputs some token Z, because it's the most probable considering cached decoder output (calculated for previous context before reset - "X Y").
Then on step t1, the beam-search-decoder calculates new decoder_output using current reset context which now is "<blank> Z".

It may happen that Z is no longer most probable hypothesis and so the beam search switches to another path. User sees this as "Z" flickering and getting deleted. Sometimes this switch can also happen after outputting 2-3 and probably even more tokens. Besides user discomfort, deleted words frequently contain correct transcript (at least in my subjective experience). 

This PR fixes this by using previous result tokens as "context" for next segment instead of caching the decoder output for one timestep.

### DIFF
--- a/sherpa-onnx/csrc/online-recognizer-transducer-impl.h
+++ b/sherpa-onnx/csrc/online-recognizer-transducer-impl.h
@@ -360,21 +360,19 @@ class OnlineRecognizerTransducerImpl : public OnlineRecognizerImpl {
   }
 
   void Reset(OnlineStream *s) const override {
+    int32_t context_size = model_->ContextSize();
+
     {
       // segment is incremented only when the last
-      // result is not empty
+      // result is not empty, contains non-blanks and longer than context_size)
       const auto &r = s->GetResult();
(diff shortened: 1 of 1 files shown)


csukuangfj:

PR #1401, opened 2024-10-09
docs(nodejs-addon-examples): add guide for pnpm user

YogiLiu:
For example, pnpm will install the `sherpa-onnx-linux-x64` at `node_modules/.pnpm/sherpa-onnx-linux-x64@1.10.27`, so the addon.js can not find `sherpa-onnx.node`.

### DIFF
--- a/nodejs-addon-examples/README.md
+++ b/nodejs-addon-examples/README.md
@@ -12,41 +12,32 @@ Note: [../nodejs-examples](../nodejs-examples) uses WebAssembly to wrap
 Before you continue, please first run
 
 ```bash
-npm install
+npm install # or pnpm install
 
 # For macOS x64
+## With npm
 export DYLD_LIBRARY_PATH=$PWD/node_modules/sherpa-onnx-darwin-x64:$DYLD_LIBRARY_PATH
+## With pnpm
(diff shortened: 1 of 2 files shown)


csukuangfj:
> pnpm will install the sherpa-onnx-linux-x64 at node_modules/.pnpm/sherpa-onnx-linux-x64@1.10.27

In that case, could you modify
https://github.com/k2-fsa/sherpa-onnx/blob/8535b1d3bbbf4bfedf3924abaf563e3e346ad042/scripts/node-addon-api/lib/addon.js#L10

Does it work if you add
```
  `./node_modules/.pnpm/sherpa-onnx-${platform_arch}/sherpa-onnx.node`,

```

YogiLiu:
No, because the folder name with the version like `sherpa-onnx-linux-x64@1.10.27`.

So alternative way to resolve this issue is parse the installed version of `sherpa-onnx-linux-x64`, but it is a complicated way.

YogiLiu:
Actually the folder structure like this:

image

YogiLiu:
Update: It is a issue about environment variable `LD_LIBRARY_PATH`.

csukuangfj (approved):

PR #1436, opened 2024-10-16
updated onnxruntime-linux-aarch64.cmake so that libonnxruntime.so can…

shawl336:
updated onnxruntime-linux-aarch64.cmake so that libonnxruntime.so can be found for specific aarch64 cross-compilation environments


csukuangfj (approved):

PR #1438, opened 2024-10-17
fix "log10" compile error by import CMATH lib

Zazzle516:
在进行 RISCV 运行的 sherpa-onnx 交叉编译的过程中，执行 ./build-riscv64-linux-gnu.sh 脚本的时候出现了
error: 'log10' is not a member of 'std'
在 sherpa-onnx/sherpa-onnx/csrc/offline-whisper-model.cc 文件中导入 <cmath> 库解决


csukuangfj (commented):
Could you put it after line 7?

We need to keep the `includes` sorted alphabetically.

PR #1452, opened 2024-10-21
JNI Exception Handling

iprovalo:
JNI Exception Handling.  Throw Java RuntimeException and let the client deal with it.  The advantage of this approach is that a client may decide to show a different message to the user.


sherpa-onnx/jni/voice-activity-detector.cc
```diff
@@ -173,11 +180,18 @@ JNIEXPORT bool JNICALL Java_com_k2fsa_sherpa_onnx_Vad_isSpeechDetected(
...
 SHERPA_ONNX_EXTERN_C
-JNIEXPORT void JNICALL Java_com_k2fsa_sherpa_onnx_Vad_reset(JNIEnv * /*env*/,
+JNIEXPORT void JNICALL Java_com_k2fsa_sherpa_onnx_Vad_reset(JNIEnv *env,
                                                             jobject /*obj*/,
                                                             jlong ptr) {
-  auto model = reinterpret_cast<sherpa_onnx::VoiceActivityDetector *>(ptr);
-  model->Reset();
+  try {
+    auto model = reinterpret_cast<sherpa_onnx::VoiceActivityDetector *>(ptr);
+    model->Reset();
```
csukuangfj, line 191: It is hard to imagine that this function can throw.

Could you post the exception messages?

iprovalo:
@csukuangfj would you consider merging this PR?  This is related to [JNI discussion](https://github.com/k2-fsa/sherpa-onnx/discussions/1426).

The justification is that there could be some unforeseen conditions causing the data to become corrupt, or a device type compatibility issue with any of the underlying native libraries, causing a crash.  JNI layer exception handling allows to delegate the exception handling to the UI and add some useful messages for the user to react to it by reinstalling the app in case if this behavior continues.  

I have noticed a significant drop in crashes when I rolled these out to a few hundred devices.

sherpa-onnx/jni/offline-recognizer.cc
```diff
@@ -333,11 +333,40 @@ Java_com_k2fsa_sherpa_onnx_OfflineRecognizer_createStream(JNIEnv * /*env*/,
...
-  auto stream = reinterpret_cast<sherpa_onnx::OfflineStream *>(streamPtr);
-
-  recognizer->DecodeStream(stream);
+    JNIEnv *env, jobject /*obj*/, jlong ptr, jlong streamPtr) {
+  try {
+    if (ptr == 0) {  // Check if the pointer is null
+      jclass exClass = env->FindClass("java/lang/NullPointerException");
+      if (exClass != nullptr) {
+        env->ThrowNew(exClass, "OfflineRecognizer_decode: "
+                      "OfflineRecognizer pointer is null.");
```
csukuangfj: Can you write a function to throw an exception so that a lot of duplicate code is reduced to call that function.

csukuangfj (changes requested):

sherpa-onnx/jni/voice-activity-detector.cc
iprovalo: There is not much in the exception log right now, beside this:

#00  pc 0x0000000000090480  /data/app/~~OPsl1ZjlmW-J2qp-_4TVtA==/com.lingofonex.android-VkSxvfpHKGKaQ2dvrt-dEw==/split_config.arm64_v8a.apk!libsherpa-onnx-jni.so (Java_com_k2fsa_sherpa_onnx_Vad_reset+16) (BuildId: df851e8489be509b8a09d60eab3ba57ea3ff5786)

While these exceptions are below 2%, they are still hurting the metrics significantly.  I deployed debug symbols in my latest release, if I get any more data pointing to specific code area, I will share.  For now, I am trying to handle these in the JNI, and try to recover the UI without a crash.

sherpa-onnx/jni/offline-recognizer.cc
iprovalo: Yes, good point, let me do that!

iprovalo: Updated.

csukuangfj:
Is it ready for merge?

iprovalo:
> Is it ready for merge?

Yes, this is ready to merge.

csukuangfj:

PR #1461, opened 2024-10-24
Handle NaN embeddings in speaker diarization.

csukuangfj:
It should fix https://github.com/thewh1teagle/sherpa-rs/issues/33

CC @WilliamVenner


jamescarter-le:
I encounter this NaN issue with most models (3dspeaker_speech_eres2net_sv_en_voxceleb_16k works), like the NEMO models returning NaN for most (but not all) audio clips.

I'm using .NET, `org.k2fsa.sherpa.onnx` `1.10.30`

PR #1463, opened 2024-10-24
Add speaker identification with VAD and non-streaming ASR using ALSA

Peakyxh:
This PR adds a file speaker-identification-with-vad-non-streaming-asr-alsa.py.


csukuangfj (approved):

PR #1510, opened 2024-11-05
Reduce vad-sense-voice example code.

whyb:
```c
while(..) {
}
while(..) {
}
```
So I reduce sherpa-onnx/c-api-examples/vad-sense-voice-c-api.c some codes.


csukuangfj (approved):

PR #1524, opened 2024-11-08
Revert: [#1521] No need to reset sample-buffer

iteamvep:
Revert commit #1521


csukuangfj:
Thank you for you contribution!

PR #1588, opened 2024-12-03
fix: support both old and new websockets request headers format

JiayuXu0:
Add compatibility check for request_headers format in websocket handling.
New version of websockets has a 'headers' attribute in request_headers,
while old version uses request_headers directly.
This change ensures the server can properly handle requests from different
versions of the websockets library.


python-api-examples/non_streaming_server.py
```diff
@@ -641,7 +641,11 @@ async def process_request(
         path: str,
         request_headers: websockets.Headers,
     ) -> Optional[Tuple[http.HTTPStatus, websockets.Headers, bytes]]:
-        if "sec-websocket-key" not in request_headers:
+        if "sec-websocket-key" not in (
```
csukuangfj, line 644: Could you also fix `streaming_server.py`?

JiayuXu0: It has been also fixed now.

csukuangfj:

PR #1638, opened 2024-12-21
SherpaOnnxVadAsr: Offload runSecondPass to background thread for improved real-time audio processing

rominf:
This change ensures that the main audio processing loop is not blocked by long-running operations in `runSecondPass`, improving responsiveness and reducing the risk of missing parts of input speech.


csukuangfj:

PR #1680, opened 2025-01-04
Upgraded to .NET 8 and made code style a little more internally consistent.

Lamothe:
The existing example projects target .NET 6 which is no longer supported.  This PR upgrades them to .NET 8, which is the current LTS version.


csukuangfj (approved):

PR #1681, opened 2025-01-04
Update workflows to use .NET 8.0 also.

Lamothe:
Oops, I broke the build.  The tests are failing because they aren't using .NET 8.


csukuangfj (commented):

csukuangfj (approved):

PR #1732, opened 2025-01-17
Add cache mechanism to sherpa tts

mah92:
Delay is a major drawback of sherpa tts on android phones, which makes it unusable for average blind people using average phones. By caching the most frequent tts requests, sherpa is made as fast as lightning on most of the time, using much less cpu and battery. Some blind individials have tested this upgrade and have been happy :)

Considered guide-lines:
- App does not cache user texts due to security and privacy concerns. Just a hash is saved from the texts.
- Usage statistics of texts(hashes) are saved to a list, and stored periodically(every 10 minutes). If cache space is filled, wav files that are less used are removed.
- Setting amount of cache and the ability to remove cached wav files are added in the MainActivity
- Present jni functions are tried to remain intact due to possible compatibility considerations.
- When changing speed, cache could be cleared automatically to update the speed of cached media as well, not implemented yet as there is a clear button on MainActivity.kt.
- Code is scanned for possible bugs with deepseek AI.


csukuangfj:

Will review it.

android/SherpaOnnxTtsEngine/app/src/main/java/com/k2fsa/sherpa/onnx/tts/engine/MainActivity.kt
```diff
@@ -88,6 +94,17 @@ class MainActivity : ComponentActivity() {
                     }) {
                         Box(modifier = Modifier.padding(it)) {
                             Column(modifier = Modifier.padding(16.dp)) {
+                                // Track used cache size in a mutable state
+                                var usedCacheSizeMB by remember { mutableStateOf(0) }
+
+                                // LaunchedEffect to periodically update the used cache size
+                                LaunchedEffect(Unit) {
```
csukuangfj, line 103: Can you use a coroutine to execute it periodically?

Otherwise, the `delay()` here will block one thread.
(Please correct me if I am wrong.)

sherpa-onnx/csrc/offline-tts-cache-mechanism.h
```diff
@@ -0,0 +1,82 @@
+// sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
```
csukuangfj:
```suggestion
// sherpa-onnx/csrc/offline-tts-cache-mechanism.h
```

sherpa-onnx/csrc/offline-tts-cache-mechanism.h
```diff
@@ -0,0 +1,82 @@
...
+class CacheMechanism {
+ public:
+  CacheMechanism(const std::string &cache_dir, int32_t cache_size);
+  ~CacheMechanism();
+
+  // Add a new wav file to the cache
+  void AddWavFile(const std::string &text_hash, const std::vector<float> &samples, int32_t sample_rate);
+
+  // Get the cached wav file if it exists
+  std::vector<float> GetWavFile(const std::string &text_hash, int32_t &sample_rate);
```
csukuangfj:
```suggestion
  std::vector<float> GetWavFile(const std::string &text_hash, int32_t &sample_rate) const;
```

For `Get` methods, please mark them as `const`.

sherpa-onnx/csrc/offline-tts-cache-mechanism.h
```diff
@@ -0,0 +1,82 @@
...
+  int32_t GetCacheSize() const;
+
+  // Set the cache size in bytes
+  void SetCacheSize(int32_t cache_size);
+
+  // Remove all the wav files in the cache 
+  void ClearCache();
+
+  // To get total used cache size(for wav files) in bytes
+  int64_t GetTotalUsedCacheSize();
```
csukuangfj:
```suggestion
  int32_t GetTotalUsedCacheSize() const;
```

sherpa-onnx/csrc/offline-tts-cache-mechanism.h
```diff
@@ -0,0 +1,82 @@
...
+  // Reduce used cache size if needed
+  void EnsureCacheLimit();
+
+  // Get the least repeated file in the cache
+  std::string GetLeastRepeatedFile();
+
+  // Data directory where the cache folder is located
+  std::string cache_dir_;
+
+  // Maximum number of files in the cache
```
csukuangfj:
```suggestion
  // Maximum number of bytes in the cache
```

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
```diff
@@ -0,0 +1,272 @@
...
+#include "sherpa-onnx/csrc/wave-reader.h"
+#include "sherpa-onnx/csrc/wave-writer.h"
+
+namespace sherpa_onnx {
+
+CacheMechanism::CacheMechanism(const std::string &cache_dir, int32_t cache_size)
+    : cache_dir_(cache_dir), cache_size_bytes_(cache_size), used_cache_size_bytes_(0) {
+  // Create the cache directory if it doesn't exist
+  if (!std::filesystem::exists(cache_dir_)) {
+    std::filesystem::create_directory(cache_dir_);
```
csukuangfj: What if it fails to create the directory?

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
```diff
@@ -0,0 +1,272 @@
...
+    // SHERPA_ONNX_LOGE("Created cache directory: %s", cache_dir_.c_str());
```
csukuangfj: Please remove unused code.

sherpa-onnx/csrc/offline-tts-cache-mechanism.h
```diff
@@ -0,0 +1,82 @@
...
+  int32_t cache_size_bytes_;
+
+  // Total used cache size for wav files in bytes
+  int64_t used_cache_size_bytes_;
```
csukuangfj:
```suggestion
  int32_t used_cache_size_bytes_;
```

Please use `int64_t` only when needed.

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
```diff
@@ -0,0 +1,272 @@
...
+      if (repeat_counts_.find(text_hash) == repeat_counts_.end()) {
+        // Remove the file if it's not in the repeat count file
+        std::filesystem::remove(entry.path());
+        // SHERPA_ONNX_LOGE("Removed orphaned wav file: %s", entry.path().c_str());
+      } else {
+        // Add the size of the WAV file to the total cache size
+        std::ifstream file(entry.path(), std::ios::binary | std::ios::ate);
+        if (file.is_open()) {
+          used_cache_size_bytes_ += file.tellg();
+          file.close();
```
csukuangfj:
```suggestion
```

It is called in the destructor of `file`.

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
```diff
@@ -0,0 +1,272 @@
...
+        }
+        cache_vector_.push_back(text_hash);
```
csukuangfj, line 289:
```suggestion
        cache_vector_.push_back(std::move(text_hash));
```

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
```diff
@@ -0,0 +1,272 @@
...
+      }
+    }
+  }
+}
+
+void CacheMechanism::EnsureCacheLimit() {
+  if(used_cache_size_bytes_ > cache_size_bytes_) {
+    auto target_cache_size = std::max((int)(cache_size_bytes_*0.95), 0); //Remove more to prevent deleting every step
```
csukuangfj:
```suggestion
    auto target_cache_size = std::max(static_cast<int>(cache_size_bytes_*0.95), 0); //Remove more to prevent deleting every step
```

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
```diff
@@ -0,0 +1,272 @@
...
+}
+
+void CacheMechanism::RemoveWavFile(const std::string &text_hash) {
+  std::string file_path = cache_dir_ + "/" + text_hash + ".wav";
+  if (std::filesystem::exists(file_path)) {
+    // Subtract the size of the removed WAV file from the total cache size
+    std::ifstream file(file_path, std::ios::binary | std::ios::ate);
+    if (file.is_open()) {
+      used_cache_size_bytes_ -= file.tellg();
+      file.close();
```
csukuangfj, line 259:
```suggestion
```

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
```diff
@@ -0,0 +1,272 @@
...
+    EnsureCacheLimit();
+
+    // Write the audio samples to a WAV file
+    bool success = WriteWave(file_path, sample_rate, samples.data(), samples.size());
+    if (success) {
+      // Calculate the size of the new WAV file and add it to the total cache size
+      std::ifstream file(file_path, std::ios::binary | std::ios::ate);
+      if (file.is_open()) {
+        used_cache_size_bytes_ += file.tellg();
+        file.close();
```
csukuangfj:
```suggestion
```

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
```diff
@@ -0,0 +1,272 @@
...
+    bool is_ok = false;
+    samples = ReadWave(file_path, &sample_rate, &is_ok);
+
+    if (is_ok) {
+      // SHERPA_ONNX_LOGE("Retrieved cached audio for text hash: %s, sample count: %d, freq:%dHz", text_hash.c_str(), samples.size(), sample_rate);
+    } else {
+      SHERPA_ONNX_LOGE("Failed to read cached file: %s", file_path.c_str());
+    }
+  } else {
+    // SHERPA_ONNX_LOGE("Cached audio not found for text hash: %s", text_hash.c_str());
```
csukuangfj: Please remove unused code and `else`.

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  // Ensure the text_hash exists in the map before incrementing the count
+  if (repeat_counts_.find(text_hash) == repeat_counts_.end()) {
+    repeat_counts_[text_hash] = 0; // Initialize if it doesn't exist
+  }
+  repeat_counts_[text_hash]++; // Increment the repeat count
```
csukuangfj: If the file does not exist, shall we skip updating the count?

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
```diff
@@ -0,0 +1,272 @@
...
+    if (entry.path().extension() == ".wav") {
+      std::filesystem::remove(entry.path());
+      // SHERPA_ONNX_LOGE("Removed wav file: %s", entry.path().c_str());
+    }
+  }
+
+  // SHERPA_ONNX_LOGE("Removed all wav files!");
+
+  // Reset the total cache size to 0
+  used_cache_size_bytes_ = 0;
```
csukuangfj, line 169: Do you also need to update `repeat_counts_` and `cache_vector_`?

sherpa-onnx/csrc/CMakeLists.txt
```diff
@@ -167,6 +167,7 @@ if(SHERPA_ONNX_ENABLE_TTS)
     offline-tts-vits-model.cc
     offline-tts.cc
     piper-phonemize-lexicon.cc
+    offline-tts-cache-mechanism.cc
```
csukuangfj: please sort the filenames alphabetically.

sherpa-onnx/csrc/offline-tts.h
```diff
@@ -12,6 +12,7 @@
 
 #include "sherpa-onnx/csrc/offline-tts-model-config.h"
 #include "sherpa-onnx/csrc/parse-options.h"
+#include "sherpa-onnx/csrc/offline-tts-cache-mechanism.h"
```
csukuangfj: Please sort the includes alphabetically.

sherpa-onnx/csrc/offline-tts.h
```diff
@@ -32,6 +33,9 @@ struct OfflineTtsConfig {
   // If you set it to -1, then we process all sentences in a single batch.
   int32_t max_num_sentences = 1;
 
+  // Path to cache_directory
+  std::string cache_dir;
```
csukuangfj: Please create a new config for cache.

sherpa-onnx/csrc/offline-tts.h
```diff
@@ -87,12 +91,26 @@ class OfflineTts {
...
+  int32_t CacheSize() const;
+
+  // Set the maximum number of cached audio files size
+  void SetCacheSize(const int32_t cache_size);
+
+  // Remove all cache data
+  void ClearCache();
+
+  // To get total used cache size(for wav files) in bytes
+  int64_t GetTotalUsedCacheSize();
```
csukuangfj: Please remove these methods from OfflineTtts.

You can pass a smart pointer of `CacheMechanism` to the constructor of OffineTts.
(You can add an overload constructor for OfflineTts)

sherpa-onnx/csrc/offline-tts.h
```diff
@@ -87,12 +91,26 @@ class OfflineTts {
...
+  
   // Number of supported speakers.
   // If it supports only a single speaker, then it return 0 or 1.
   int32_t NumSpeakers() const;
 
  private:
+  OfflineTtsConfig config_;
   std::unique_ptr<OfflineTtsImpl> impl_;
+  std::unique_ptr<CacheMechanism> cache_mechanism_;
```
csukuangfj:
```suggestion
  std::unique_ptr<OfflineTtsImpl> impl_;
```

csukuangfj (changes requested):

csukuangfj:
Please use
https://github.com/k2-fsa/sherpa-onnx/blob/master/scripts/check_style_cpplint.sh
to check the style of your C++ code.

You can refer to
https://github.com/k2-fsa/sherpa-onnx/blob/66e02d83815c4f4010e81c479bf690e44b865cae/scripts/check_style_cpplint.sh#L19-L26

---

You can use
```
pip install clang-format==12.0.1
```
to install clang-format and use it to auto-format your code.

mah92:
Thank you, I will correct tomorrow. 
I wonder if it is better to remove the clear button and remove the cache automatically when the speed is changed. By the way the clear button seems to be displaced after merging to the newer commit.

csukuangfj:
> I wonder if it is better to remove the clear button and remove the cache automatically when the speed is changed

Yes, I agree.

---

By the way, you can even set a default cache size. Users don't need to set the cache size through the UI.

mah92:
Another proposal: whatever we do in the model, there still remains some words that are not spelled correctly(like keyboard letters). Add them permanently in the cache so that they be read fron the cache instead of the model?

mah92:
About the cache slider, I think it is good to keep it for a while. Then rethink based on user feedback.

csukuangfj:
> the cache instead of the model?

Ok, it is fine with me.

csukuangfj:
> Another proposal: whatever we do in the model, there still remains some words that are not spelled correctly(like keyboard letters). Good to add them permanently in the cache so that they be read fron the cache instead of the model?

Please make the cache do only 1 thing.

You can create a separate PR to add it.

android/SherpaOnnxTtsEngine/app/src/main/java/com/k2fsa/sherpa/onnx/tts/engine/MainActivity.kt
mah92: I searched, LaunchedEffect is indeed a coroutine, and delay is non blocking any thread.

sherpa-onnx/csrc/offline-tts-cache-mechanism.h
mah92: corrected.

sherpa-onnx/csrc/offline-tts-cache-mechanism.h
mah92: Impossible as repeat_counts_ is increased in this function. It is to keep the statistics of hashes.

sherpa-onnx/csrc/offline-tts-cache-mechanism.h
mah92: corrected

sherpa-onnx/csrc/offline-tts-cache-mechanism.h
mah92: corrected

mah92: Corrected. At the beginning, I tought that the cache might go over 2GB, but now I know that this is not the case.

sherpa-onnx/csrc/offline-tts.h
mah92: corrected

sherpa-onnx/csrc/CMakeLists.txt
mah92: corrected

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
mah92: Added cache_mechanism_inited_ variable so that the library can be disabled. Added checking this variable through all public functions.

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
mah92: Done

sherpa-onnx/csrc/offline-tts-cache-mechanism.h
mah92: done

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
mah92: done

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
mah92: done

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
mah92: done

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
mah92: done

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
mah92: Kept file.close() here as I want to remove the file before desctructor of the file is called.

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
mah92: done

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
mah92: You are correct. I wrongly incremented the count in both in the GetWavFile and AddWavFile functions. Now incrementing is done only in the GetWavFile.

sherpa-onnx/csrc/offline-tts-cache-mechanism.cc
mah92: You are correct. I need to reset them too. Thank you for your attention to the details.

sherpa-onnx/csrc/offline-tts.h
mah92: What do you mean exactly?

sherpa-onnx/csrc/offline-tts.h
mah92: I did not get how.
You mean I directly call functions in CacheMechanism from jni functions themselves?

sherpa-onnx/csrc/offline-tts.h
mah92: I did not get how.

sherpa-onnx/csrc/offline-tts.h
mah92: I can remove cache_dir by calling the constructor from within GetOfflineTtsConfig. Is it acceptable? 
If not, please give more details.

sherpa-onnx/csrc/offline-tts.h
mah92: I do not get it. If CacheMechanism is to be added to the constructor of OffineTts, aren't these lines still needed?

mah92:
> Please use https://github.com/k2-fsa/sherpa-onnx/blob/master/scripts/check_style_cpplint.sh to check the style of your C++ code.
> 
> You can refer to
> 
> https://github.com/k2-fsa/sherpa-onnx/blob/66e02d83815c4f4010e81c479bf690e44b865cae/scripts/check_style_cpplint.sh#L19-L26
> 
> You can use
> 
> ```
> pip install clang-format==12.0.1
> ```
> 
> to install clang-format and use it to auto-format your code.

done.

sherpa-onnx/csrc/offline-tts.h
csukuangfj: For example, OfflineTts has a config called OfflineTtsConfig.

You can create a struct called CacheMechanismConfig and put cache_dir in it. In this way, you don't need to add a new field to OfflineTtsConfig. You also don't need to add new methods to OfflineTts. (Note you need to wrap CacheMechanism and its config to Kotlin via jni)

Remember that sherpa-onnx currently supports 12 programming languages. If we change the config in C++, we also need to update APIs of other programming languages.

You can create a smart pointer of CacheMechanism and pass it to the constructor of OfflineTts. 

If later we need to extend the CacheMechanism, we can leave OfflineTts and OfflineTtsConfig untouched. We only need to change CacheMechanism or/and CacheMechanismConfig.

By the way, since the CacheMechanism is related to tts, can you rename it to OfflineTtsCacheMechanism?

In general, the filename of a class contains the class name, though there are dashes in the filename but not in the class name

Methods of CacheMechanism  don't need to be repeated in OfflineTts.

sherpa-onnx/csrc/offline-tts.h
csukuangfj: Yes. CacheMechanism is constructed outside of OfflineTts.

OfflineTts is only for converting text to speech. Anything related to cache should be put in a separate class, i.e., please don't repeat the methods of CacheMechanism in OfflineTts.

sherpa-onnx/csrc/offline-tts.h
csukuangfj: please update OfflineTtsImpl.

You don't need to add any new fields or any new methods to OfflineTts.

You can add a new field
```
CacheMechanism* cache_; // not owned here
```
to `OfflineTtsImpl`.

sherpa-onnx/csrc/offline-tts.h
csukuangfj: We use the `pimpl` implementation here.

sherpa-onnx/csrc/offline-tts.h
mah92: Shall I add OfflineTtsCacheMechanismConfig to OfflineTtsConfig or keep it separate?

mah92:
Shall I add OfflineTtsCacheMechanismConfig to OfflineTtsConfig or keep it separate?

sherpa-onnx/csrc/offline-tts.h
csukuangfj: Please keep it separate.

csukuangfj:
You don't need to wrap.impl_ to.jni.


You need to.wrap. CacheMechanism to jni.


Not sure why you.need to wrap impl_ to jni.

mah92:
You told me ealier to update OfflineTtsImpl and add CacheMechansim pointer there. It was impossible due to it's private nature. So I should cache the pointer in the  jni/offline-tts.cc?

csukuangfj:
Yes, I said that.

Let me give.you the pseudo code later.

mah92:
This is what I have done in the last commit:

csukuangfj:
So here is the pseudo code.

1. Create a cache mechanism outside of OfflineTts.
```cpp
OfflineTtsCacheMechanismConfig config;
// set config

auto cache = new OfflineTtsCacheMechanism(config)
```

2. Now pass the pointer to `cache` to the OfflineTts
```cpp
OfflineTtsConfig config;

auto tts = new OfflineTts(config, cache);
```

3. Now Change OfflineTtsImpl
```cpp
class OfflineTtsImpl {
public:
  OfflineTtsImpl(const OfflineTtsConfig& config, OfflineTtsCacheMechanism* cache)
   : ...
     cache_(cache) {
   ...
  }

private:
  OfflineTtsCacheMechanism *cache_; // not owned here
};
```

Note again in the above code, you don't need to add any new methods (except the constructor) of OfflineTts.

You also don't need to care about that `impl_` is private in `OfflineTts`.

---

You just need to wrap `OfflineTtsCacheMechanism`  and `OfflineTtsCacheMechanismConfig`. to JNI.

---

Please don't add any method related to OfflineTtsCacheMechanism to `OfflineTts`. You don't need to do that.

sherpa-onnx/csrc/offline-tts.h
```diff
@@ -91,6 +99,8 @@ class OfflineTts {
   // If it supports only a single speaker, then it return 0 or 1.
   int32_t NumSpeakers() const;
 
+  std::unique_ptr<OfflineTtsCacheMechanism> cache_mechanism_; // not owned here
```
csukuangfj:
```suggestion
```

Please don't add any new fields to OfflineTts.

As said before, we use `pimpl` here.

If you don't know what `pimpl` is, please google it first.

The only member `OfflineTts` has is `impl_`.

Everything else is put in`OfflineTtsImpl`.

csukuangfj (changes requested):

mah92:
I will catch up soon. By the way, What files needs change in the 1. and 2. part of pseudo code change comment?

csukuangfj:
> What files needs change in the 1. and 2. part of pseudo code change comment?

> Create a cache mechanism outside of OfflineTts.

You can first wrap the cache mechanism class to kotlin via jni. Then you can create it in kotlin, like how you create offline tts in kotlin.

> Now pass the pointer to cache to the OfflineTts

You need to create offline tts in kotlin, which means you need to pass the instance of cache mechanism from kotlin.

mah92:
Hi again. I have progressed, but need some help.
1) "You just need to wrap OfflineTtsCacheMechanism and OfflineTtsCacheMechanismConfig. to JNI."
 What do you mean to wrap them? Would you give me s pseudo code?
Should I cache "cache" as static in jni/offline-tts.cc?
2) I see the warning "This branch must not contain merge commits.". Should I rebase the whole branch?

mah92:
By the way, I have made a working thread for cleanup based on user feedback.

mah92:
I think the remaining work is not sth that I can do without your elaborate help. maybe It takes less time from you...

mah92:
After thinking thoroughly and implementing two android apps using sherpa-onnx framework, I think this feature is something the developer should implement in the upper layer, not needed in the sherpa-onnx itself. It makes sherpa too complex and I didn't use this for my apps. I propose to close this. Agree? @csukuangfj

csukuangfj:
Yes, please close it.

PR #1739, opened 2025-01-20
Fix `dither` binding in Pybind11 to ensure independence from `high_freq` in `FeatureExtractorConfig`

jacklynblack:
This PR fixes the incorrect binding of the dither attribute in the FeatureExtractorConfig Pybind11 bindings. Previously, dither was mistakenly bound to the high_freq member variable, causing unexpected behavior where changes to one attribute affected the other.

The fix updates the Pybind11 binding to correctly map dither to its own member variable, ensuring that dither and high_freq are independent properties in Python


csukuangfj (approved):

PR #1742, opened 2025-01-21
Reduce vad-moonshine-c-api example code.

whyb:
Reduce vad-moonshine-c-api example code.

```cpp
while(..) {
}
while(..) {
}
```

So I reduce sherpa-onnx/c-api-examples/vad-moonshine-c-api.c some codes.

The PR like this: https://github.com/k2-fsa/sherpa-onnx/pull/1510

The screenshot is test results after modification in my pc.
圖片


csukuangfj (approved):

Issue #1777, opened 2025-01-30
Skipping words


Violetadhdgd:
The tts has a problem where it skips some words at the start of a sentence or after a comma. Not everytime but it happens with certain commonly used words

csukuangfj:
Can you.be.more specific?

For instance, it is not.clear which model.you are using.

Violetadhdgd:
Kokoro tts

csukuangfj:
Can you be more specific?

  Would be great if you can post command to reproduce.it.

ahadjawaid:
I have the same issue. You can try it with a lot of sentences. It usually happens at the start of a sentence. Here are some example sentences: 
"Hi I’m ahad and like to have fun." Here Hi I'm is skipped
"At YC event last week Brian Chesky gave a talk." the phrase At is skipped. 
There are some more random ones I tried. Let me know if you need more.

eolasd:
I have found the same issue with the sample code.
Using the C#/Dotnet sample code straight from here: 
https://github.com/k2-fsa/sherpa-onnx/blob/master/dotnet-examples/kokoro-tts/Program.cs

It misses the first word on every inference.
Using Kokoro TTS 0.19 and latest sherpa v1.10.41

ahadjawaid:
I think the issue may stem from the phonemizer. I'm uncertain where to check to test this issue. Does anyone know where the phonemizer code is?

ahadjawaid:
I found the bug. You need to prepend a 0 token when tokenizing the phonemes. I'll make pull request to close this issue.

csukuangfj:
Fixed in #1787

ahadjawaid:
I fixed it in the C source code. So you can find what is the source code for flutter and fix it there if it exists there.

csukuangfj:
We will have a new release this week, hopefully tomorrow.

csukuangfj:
Yes, today.

PR #1787, opened 2025-02-03
Fix: Prepend 0 to tokenization to prevent word skipping for Kokoro.

ahadjawaid:
Addressed issue Skipping words #1777

In the original code they have a 0 in front and behind. Currently it only has it behind so I added it to the front also. Here is the original code:
https://huggingface.co/hexgrad/kLegacy/blob/d0acca3b0748cae52e0a14a28ad03515002ff22f/v0.19/kokoro.py#L117


csukuangfj:

csukuangfj (approved):

PR #1817, opened 2025-02-07
Update sherpa-onnx-tts.js  VitsModelConfig.model can be none

cgisky1980:


csukuangfj (approved):

PR #1837, opened 2025-02-10
Disable strict hotword matching mode for offline transducer

vsd-vector:
I noticed that for offline transducer strict_mode is True when traversing ContextGraph, but for online non-strict mode is used (see #638 ).

What is the rationale behind this?

In my testing, when using strict_mode the decoder sometimes "gets stuck" in the final state of the hotword graph, penalizing all tokens after the hotword. In other words, if we have sentence "he is my friend" and hotwords list contains "he", then sometimes decoder output is just "he" and remainder of the sentence is truncated. This is typically happens if weight is too large, but I would still consider such behaviour undesirable for ASR (not keyword spotter). 

This PR proposes to disable strict hotword matching mode for offline transducer. 
Also it introduces new variable, so that later this mode can be made configurable.

### DIFF
--- a/sherpa-onnx/csrc/offline-transducer-modified-beam-search-decoder.cc
+++ b/sherpa-onnx/csrc/offline-transducer-modified-beam-search-decoder.cc
@@ -138,7 +138,9 @@ OfflineTransducerModifiedBeamSearchDecoder::Decode(
           new_hyp.timestamps.push_back(t);
           if (context_graphs[i] != nullptr) {
             auto context_res =
-                context_graphs[i]->ForwardOneStep(context_state, new_token, strict_hotword_mode_);
+                context_graphs[i]->ForwardOneStep(context_state,
+                  new_token,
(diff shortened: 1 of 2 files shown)


csukuangfj:
@pkufool  Please have a review.

pkufool:
@vsd-vector  I would suggest to use non_strict for hotwords. So I think make strict mode false by default is enough, don't necessary to make it configurable.

vsd-vector:
@pkufool I removed the variable and replaced it with "false" constant.

pkufool (approved):
LGTM

vsd-vector:
@csukuangfj any chance to merge this into next release?

csukuangfj:

vsd-vector:
@csukuangfj 

When it can be merged? Is something missing?

csukuangfj:
Sorry, missed it.

PR #1838, opened 2025-02-10
Fix: made print sherpa_onnx_loge when it is in debug mode

ahadjawaid:
Currently, during normal use you may get a lot of print statements such as: `Use espeak-ng to handle the OOV: 'ipsum'` which may not be relevant unless you are debugging.


csukuangfj (approved):

Issue #1779, opened 2025-01-31
Adding fa/en matcha-tts model for android


mah92:
Hi
I have trained a matcha tts model for fa/en, It sounds very good. After reading the docs, I needed to get your opinion on some of the points before trying to merge it. First of Al, I saw the matcha model sample in the TtsEngine.kt file.

1) Should I add the metadata to the model as in vits/piper?

2) I have a 24KHz dataset which degrades when converted to 22050Hz. So I switched to 24KHz. Doing so, I also trained a vocoder. As 22050Hz was hardcoded in the matcha-tts itself, I wondered if any change would be needed in sherpa.

3) In the matcha-tts repo, there is an option to embed the vocoder in the onnx model (to perform like an end to end model). I saw in the documentation that you gave the vocoder explicitly. Is it possible (and isn't it better) to embed the vocoder and treat the model like and end to end model (like vits)?

4) Which hifi-gan version do you propose (v1, v2, or v3)? You used v2 in  the examples, but the matcha-tts defaults to v1... I trained v1, and v3 but could not hear any difference.

5) I found two repositories training hifigan:
rhasspy/hifi-gan-train (containing onnx export)
jik876/hifi-gan (which I currently use, but I did not found onnx export, so I wrote one and I am working on it)
Is there any difference in sherpa to use which?

P.n: It is not directly relevant to sherpa but do you know any standard way to convert jik876/hifi-gan to onnx usable in sherpa? I did not find any.

csukuangfj:
> Should I add the metadata to the model as in vits/piper?

Yes, you have to do that. Please refer to
https://github.com/k2-fsa/icefall/blob/master/egs/ljspeech/TTS/matcha/export_onnx.py#L174
for how to do that. 

> As 22050Hz was hardcoded in the matcha-tts itself, I wondered if any change would be needed in sherpa.

You can use an arbitrary sample rate, as long as it matches the one used in the training. Please follow the above link
to specify the exact sample rate in the meta data. You don't need to do anything in sherpa-onnx for using a sample rate other than 22.050 kHz.

> Is it possible (and isn't it better) to embed the vocoder and treat the model like and end to end model (like vits)?

Yes, it is possible. But we choose the other way. The current C++ code in sherpa-onnx assumes you split it.

The pro is that you can replace the vocoder as you like and keep the acoustic model unchanged.

> Which hifi-gan version do you propose (v1, v2, or v3)?

[...]

PR #1841, opened 2025-02-11
Fix matcha tts model names for Persian and English

csukuangfj:
See also https://github.com/k2-fsa/sherpa-onnx/issues/1779#issuecomment-2648948128

### DIFF
--- a/scripts/apk/generate-tts-apk-script.py
+++ b/scripts/apk/generate-tts-apk-script.py
@@ -404,12 +404,12 @@ def get_matcha_models() -> List[TtsModel]:
             lang="en",
         ),
         TtsModel(
-            model_dir="matcha-tts-fa_en-male",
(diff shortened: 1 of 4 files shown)


scripts/apk/generate-tts-apk-script.py
```diff
@@ -404,12 +404,12 @@ def get_matcha_models() -> List[TtsModel]:
             lang="en",
         ),
         TtsModel(
-            model_dir="matcha-tts-fa_en-male",
+            model_dir="matcha-tts-fa_en-musa",
             acoustic_model_name="model.onnx",
             lang="fa",
```
mah92, line 409: I do not know if "fa" is appropriate here or "fas". Please make sure @csukuangfj

csukuangfj: it is correct to use fa

PR #1888, opened 2025-02-18
Fix generating Chinese lexicon for Kokoro TTS 1.0

csukuangfj:


csukuangfj:
Generated sample with this fix

Text: 银行行长说快点还钱

https://github.com/user-attachments/assets/81e2d381-bb3d-4c4a-a05c-b4b6631c5a78

PR #1891, opened 2025-02-18
Reduce vad-whisper-c-api example code.

whyb:
Reduce vad-whisper-c-api.c example code.

```cpp
while(..) {
}
while(..) {
}
```

So I reduce sherpa-onnx/c-api-examples/vad-whisper-c-api.c some codes.

This change is exactly the same as these PRs: https://github.com/k2-fsa/sherpa-onnx/pull/1510 and https://github.com/k2-fsa/sherpa-onnx/pull/1742


csukuangfj (approved):

PR #1893, opened 2025-02-18
Flutter Config toJson/fromJson

Dokotela:
I was testing different models, and realized it would be convenient to be able to pass in the Config parameters as Json. 

- Created fromJson constructors and toJson methods to all pure Dart/Flutter classes. 
- Any classes involving pointers or FFI I left alone.


csukuangfj (approved):

PR #1904, opened 2025-02-21
Fixing Whisper Model Token Normalization

iprovalo:
Fixing Whisper Model Token Normalization UTF8 characters issues.  
Adding more unit tests for the RemoveInvalidUtf8Sequences.

Issue reported [here](https://github.com/k2-fsa/sherpa-onnx/pull/1452)

The summary of the issue: all normalization works for the text in the result, however, it was not applied to tokens, causing JNI exception.

### DIFF
--- a/sherpa-onnx/csrc/offline-recognizer-whisper-impl.h
+++ b/sherpa-onnx/csrc/offline-recognizer-whisper-impl.h
@@ -24,7 +24,6 @@
 namespace sherpa_onnx {
 
 class OfflineRecognizerWhisperImpl : public OfflineRecognizerImpl {
-
  private:
   OfflineRecognitionResult Convert(const OfflineWhisperDecoderResult &src,
                                    const SymbolTable &sym_table) const {
--- a/sherpa-onnx/csrc/offline-recognizer-whisper-impl.h
+++ b/sherpa-onnx/csrc/offline-recognizer-whisper-impl.h
@@ -23,29 +23,33 @@
 
 namespace sherpa_onnx {
 
-static OfflineRecognitionResult Convert(const OfflineWhisperDecoderResult &src,
-                                        const SymbolTable &sym_table) {
-  OfflineRecognitionResult r;
-  r.tokens.reserve(src.tokens.size());
+class OfflineRecognizerWhisperImpl : public OfflineRecognizerImpl {
 
-  std::string text;
-  for (auto i : src.tokens) {
-    if (!sym_table.Contains(i)) {
-      continue;
+ private:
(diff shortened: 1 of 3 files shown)


sherpa-onnx/csrc/offline-recognizer-whisper-impl.h
```diff
@@ -23,29 +23,32 @@
...
-                                        const SymbolTable &sym_table) {
-  OfflineRecognitionResult r;
-  r.tokens.reserve(src.tokens.size());
-
-  std::string text;
-  for (auto i : src.tokens) {
-    if (!sym_table.Contains(i)) {
-      continue;
+class OfflineRecognizerWhisperImpl : public OfflineRecognizerImpl {
+ private:
```
iprovalo: @csukuangfj this is a private method, I will move it below after you take a look at the diffs.  I just converted the existing static method into this instance function.  The reason is that it needs access to `ApplyInverseTextNormalization()`.

sherpa-onnx/csrc/offline-recognizer-whisper-impl.h
```diff
@@ -156,7 +159,6 @@ class OfflineRecognizerWhisperImpl : public OfflineRecognizerImpl {
                                       std::move(cross_kv.second));
 
       auto r = Convert(results[0], symbol_table_);
-      r.text = ApplyInverseTextNormalization(std::move(r.text));
```
iprovalo, line 159: Moved to Convert()

sherpa-onnx/csrc/text-utils-test.cc
```diff
@@ -55,4 +55,77 @@ TEST(RemoveInvalidUtf8Sequences, Case1) {
   EXPECT_EQ(s.size() + 4, v.size());
 }
 
+
```
iprovalo, line 58: Added a few tests, verified the characters are being removed correctly.

sherpa-onnx/csrc/offline-recognizer-whisper-impl.h
csukuangfj: Can you move it to the end of the private section for methods in this class?

We usually put the public constructor at the beginning of a class in sherpa-onnx.

(It is a code style issue)

csukuangfj (approved):
Thanks!

Just left a minor comment.

sherpa-onnx/csrc/offline-recognizer-whisper-impl.h
iprovalo: Done.  I initially left it at the top so that the diffs are easier to read.

csukuangfj (approved):

PR #1917, opened 2025-02-25
go.mod set to use go 1.17, and use unsafe.Slice to optimize the code

franck-li:
include scripts/go/_internal go-api-examples

### DIFF
--- a/flutter/sherpa_onnx/lib/src/audio_tagging.dart
+++ b/flutter/sherpa_onnx/lib/src/audio_tagging.dart
@@ -8,9 +8,22 @@ import './sherpa_onnx_bindings.dart';
 class OfflineZipformerAudioTaggingModelConfig {
   const OfflineZipformerAudioTaggingModelConfig({this.model = ''});
 
+  factory OfflineZipformerAudioTaggingModelConfig.fromJson(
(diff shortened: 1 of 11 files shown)


scripts/go/sherpa_onnx.go
```diff
@@ -659,15 +659,15 @@ func (s *OfflineStream) GetResult() *OfflineRecognizerResult {
 	result.Emotion = C.GoString(p.emotion)
 	result.Event = C.GoString(p.event)
 	result.Tokens = make([]string, n)
-	tokens := (*[1 << 10]*C.char)(unsafe.Pointer(p.tokens_arr))[:n:n]
+	tokens := unsafe.Slice(p.tokens_arr, n) //(*[1 << 28]*C.char)(unsafe.Pointer(p.tokens_arr))[:n:n]
```
csukuangfj:
```suggestion
	tokens := unsafe.Slice(p.tokens_arr, n)
```

Can you remove unused code?

We don't need to keep
```go
// (*[1 << 28]*C.char)(unsafe.Pointer(p.tokens_arr))[:n:n]
```

franck-li: yes, I can

PR #1920, opened 2025-02-25
go.mod set to use go 1.17, and use unsafe.Slice to optimize the code

franck-li:
include scripts/go go-api-examples


csukuangfj (approved):

PR #1962, opened 2025-03-06
c-api add wave write to buffer.

cjsdurj:
in same case , the api can be useful :
- a tts server which  sent  wave  output to its client  and  not  do not need to save a tmp file to filesystem。

### DIFF
--- a/sherpa-onnx/csrc/wave-reader.cc
+++ b/sherpa-onnx/csrc/wave-reader.cc
@@ -315,7 +315,9 @@ std::vector<float> ReadWaveImpl(std::istream &is, int32_t *sampling_rate,
 std::vector<float> ReadWave(const std::string &filename, int32_t *sampling_rate,
                             bool *is_ok) {
   std::ifstream is(filename, std::ifstream::binary);
-  return ReadWave(is, sampling_rate, is_ok);
+  auto samples = ReadWave(is, sampling_rate, is_ok);
+  is.close();
+  return samples;
 }
 
 std::vector<float> ReadWave(std::istream &is, int32_t *sampling_rate,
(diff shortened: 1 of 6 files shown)


sherpa-onnx/csrc/wave-reader.cc
```diff
@@ -315,7 +315,9 @@ std::vector<float> ReadWaveImpl(std::istream &is, int32_t *sampling_rate,
 std::vector<float> ReadWave(const std::string &filename, int32_t *sampling_rate,
                             bool *is_ok) {
   std::ifstream is(filename, std::ifstream::binary);
-  return ReadWave(is, sampling_rate, is_ok);
+  auto samples = ReadWave(is, sampling_rate, is_ok);
+  is.close();
```
csukuangfj:
```suggestion
```

No need to call it explicitly.

sherpa-onnx/csrc/wave-writer.cc
```diff
@@ -61,21 +66,28 @@ bool WriteWave(const std::string &filename, int32_t sampling_rate,
...
   }
-
-  os.write(reinterpret_cast<const char *>(&header), sizeof(header));
-  os.write(reinterpret_cast<const char *>(samples_int16.data()),
-           samples_int16.size() * sizeof(int16_t));
-
+  os << buffer;
   if (!os) {
     SHERPA_ONNX_LOGE("Write %s failed", filename.c_str());
+    os.close();
```
csukuangfj:
```suggestion
```

sherpa-onnx/csrc/wave-writer.cc
```diff
@@ -61,21 +66,28 @@ bool WriteWave(const std::string &filename, int32_t sampling_rate,
...
     return false;
   }
-
+  os.close();
```
csukuangfj:
```suggestion
```

Note: This is C++, not C. You don't need to close it by yourself.
It is closed in the destructor.

csukuangfj (changes requested):
Thanks! Left some minor comments. Otherwise, it looks good to me.

Issue #1541, opened 2024-11-15
Linking issue


silvio3punkt5:
Hi,

I've experienced a linking issue on Ubuntu 22.04 when I install the onnxruntime libs (same as used in cmake: "onnxruntime-linux-x64-static_lib-1.17.1-glibc2_17.zip") upfront into /usr/local.
Steps to reproduce:
```
# install onnxruntime libs
mkdir -p /tmp/onnx && cd /tmp/onnx && wget https://github.com/csukuangfj/onnxruntime-libs/releases/download/v1.17.1/onnxruntime-linux-x64-static_lib-1.17.1-glibc2_17.zip  &&  unzip onnxruntime-linux-x64-static_lib-1.17.1-glibc2_17.zip &&  mv onnxruntime-linux-x64-static_lib-1.17.1-glibc2_17 onnxruntime && cp -a onnxruntime/* /usr/local/ && cd /tmp && rm -fr onnx*

cmake -DCMAKE_BUILD_TYPE=Release
make -j
```
It is detected during configuration:
```
-- location_onnxruntime_header_dir: /usr/local/include
-- location_onnxruntime_lib: /usr/local/lib/libonnxruntime.a
```

This results in this errors during linking:
```
/usr/bin/ld: ../../lib/libsherpa-onnx-core.a(offline-punctuation-impl.cc.o): in function `_GLOBAL__sub_I_offline_punctuation_impl.cc':
offline-punctuation-impl.cc:(.text.startup+0x17): undefined reference to `OrtGetApiBase'
/usr/bin/ld: ../../lib/libsherpa-onnx-core.a(offline-ct-transformer-model.cc.o): in function `_GLOBAL__sub_I_offline_ct_transformer_model.cc':
offline-ct-transformer-model.cc:(.text.startup+0x17): undefined reference to `OrtGetApiBase'
/usr/bin/ld: ../../lib/libsherpa-onnx-core.a(onnx-utils.cc.o): in function `_GLOBAL__sub_I_onnx_utils.cc':
onnx-utils.cc:(.text.startup+0x17): undefined reference to `OrtGetApiBase'
/usr/bin/ld: ../../lib/libsherpa-onnx-core.a(session.cc.o): in function `_GLOBAL__sub_I_session.cc':
session.cc:(.text.startup+0x17): undefined reference to `OrtGetApiBase'
[...]

PR #2032, opened 2025-03-20
fix static linking

sangeet2020:
Hi,

This PR is in reference to the issue: https://github.com/k2-fsa/sherpa-onnx/issues/1541

The current cmake logic does not handle the case where ONNX runtime is pre-installed and static linking is used.

The proposed solution modifies the `sherpa-onnx/csrc/CMakeLists.txt` file.
```cmake
if((BUILD_SHARED_LIBS OR SHERPA_ONNX_USE_PRE_INSTALLED_ONNXRUNTIME_IF_AVAILABLE) AND NOT DEFINED onnxruntime_lib_files)
  target_link_libraries(sherpa-onnx-core onnxruntime)
else()
  target_link_libraries(sherpa-onnx-core ${onnxruntime_lib_files})
```
The modified conditioning allows linking to the pre-installed ONNX Runtime when `BUILD_SHARED_LIBS` is True or when `SHERPA_ONNX_USE_PRE_INSTALLED_ONNXRUNTIME_IF_AVAILABLE` is set. This prevents build failures caused by missing onnxruntime_lib_files in static builds.

Thanks

### DIFF
--- a/sherpa-onnx/csrc/CMakeLists.txt
+++ b/sherpa-onnx/csrc/CMakeLists.txt
@@ -260,7 +260,7 @@ if(SHERPA_ONNX_ENABLE_RKNN)
   endif()
 endif()
 
-if(BUILD_SHARED_LIBS AND NOT DEFINED onnxruntime_lib_files)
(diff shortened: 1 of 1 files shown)


csukuangfj (approved):

PR #2036, opened 2025-03-21
fix vits dict dir config

amutu:
Or will report:
Please provide  --vits-dict-dir for Chinese TTS models using jieba


csukuangfj (approved):

PR #2129, opened 2025-04-17
online-transducer: reset the encoder toghter with 2 previous output symbols (non-blank)

KarelVesely84:
I've noticed a potential issue in streaming-transducer code. 

In the case of an endpoint following and empty segment, 
the prev. 2 tokens in predictor are reset to 2x blank symbol.
But, the encoder states are kept.

It seems more logical to reset BOTH the encoder states and the prev. 
blank symbols (i.e. to effectively start the decoding from scratch).

If the prev. 2 tokens are kept, the encoder states are also kept.

Would you agree too ?
K.

// I've tested it with custom encoder, recognizing a 16min long czech fairy tale from youtube.
// It has behaved normally, but I am not sure, how often there were 2 endpoints in a row,
// with no speech in between.

### DIFF
--- a/sherpa-onnx/python/csrc/online-recognizer.cc
+++ b/sherpa-onnx/python/csrc/online-recognizer.cc
@@ -67,7 +67,7 @@ static void PybindOnlineRecognizerConfig(py::module *m) {
            py::arg("max_active_paths") = 4, py::arg("hotwords_file") = "",
            py::arg("hotwords_score") = 0, py::arg("blank_penalty") = 0.0,
            py::arg("temperature_scale") = 2.0, py::arg("rule_fsts") = "",
-           py::arg("rule_fars") = "", py::arg("reset_encoder"))
+           py::arg("rule_fars") = "", py::arg("reset_encoder") = false)
       .def_readwrite("feat_config", &PyClass::feat_config)
       .def_readwrite("model_config", &PyClass::model_config)
       .def_readwrite("lm_config", &PyClass::lm_config)
--- a/sherpa-onnx/python/csrc/online-recognizer.cc
+++ b/sherpa-onnx/python/csrc/online-recognizer.cc
@@ -58,7 +58,7 @@ static void PybindOnlineRecognizerConfig(py::module *m) {
                     const OnlineLMConfig &, const EndpointConfig &,
                     const OnlineCtcFstDecoderConfig &, bool,
                     const std::string &, int32_t, const std::string &, float,
-                    float, float, const std::string &, const std::string &>(),
+                    float, float, const std::string &, const std::string &, bool>(),
            py::arg("feat_config"), py::arg("model_config"),
            py::arg("lm_config") = OnlineLMConfig(),
            py::arg("endpoint_config") = EndpointConfig(),
@@ -67,7 +67,7 @@ static void PybindOnlineRecognizerConfig(py::module *m) {
            py::arg("max_active_paths") = 4, py::arg("hotwords_file") = "",
            py::arg("hotwords_score") = 0, py::arg("blank_penalty") = 0.0,
            py::arg("temperature_scale") = 2.0, py::arg("rule_fsts") = "",
-           py::arg("rule_fars") = "")
+           py::arg("rule_fars") = "", py::arg("reset_encoder"))
       .def_readwrite("feat_config", &PyClass::feat_config)
       .def_readwrite("model_config", &PyClass::model_config)
       .def_readwrite("lm_config", &PyClass::lm_config)
@@ -82,6 +82,7 @@ static void PybindOnlineRecognizerConfig(py::module *m) {
       .def_readwrite("temperature_scale", &PyClass::temperature_scale)
       .def_readwrite("rule_fsts", &PyClass::rule_fsts)
       .def_readwrite("rule_fars", &PyClass::rule_fars)
+      .def_readwrite("reset_encoder", &PyClass::reset_encoder)
       .def("__str__", &PyClass::ToString);
 }
 
--- a/sherpa-onnx/csrc/online-recognizer-transducer-impl.h
+++ b/sherpa-onnx/csrc/online-recognizer-transducer-impl.h
@@ -382,14 +382,13 @@ class OnlineRecognizerTransducerImpl : public OnlineRecognizerImpl {
       }
     }
 
-    // reset encoder states
-    // s->SetStates(model_->GetEncoderInitStates());
-
     auto r = decoder_->GetEmptyResult();
     auto last_result = s->GetResult();
-    // if last result is not empty, then
-    // truncate all last hyps and save as the context for next result
+
     if (static_cast<int32_t>(last_result.tokens.size()) > context_size) {
+      // if last result is not empty, then
+      // truncate all last hyps and save as the 'ys' context for next result
+      // (the encoder state buffers are kept)
       for (const auto &it : last_result.hyps) {
         auto h = it.second;
         r.hyps.Add({std::vector<int64_t>(h.ys.end() - context_size,
@@ -399,6 +398,11 @@ class OnlineRecognizerTransducerImpl : public OnlineRecognizerImpl {
 
       r.tokens = std::vector<int64_t> (last_result.tokens.end() - context_size,
                                        last_result.tokens.end());
+    } else {
+      if(config_.reset_encoder) {
+        // reset encoder states, use blanks as 'ys' context
+        s->SetStates(model_->GetEncoderInitStates());
+      }
     }
 
     // but reset all contextual biasing graph states to root
--- a/sherpa-onnx/csrc/online-recognizer.cc
+++ b/sherpa-onnx/csrc/online-recognizer.cc
@@ -121,6 +121,10 @@ void OnlineRecognizerConfig::Register(ParseOptions *po) {
       "rule-fars", &rule_fars,
       "If not empty, it specifies fst archives for inverse text normalization. "
       "If there are multiple archives, they are separated by a comma.");
+
+  po->Register("reset-encoder", &reset_encoder,
+               "True to reset encoder_state on an endpoint after empty segment."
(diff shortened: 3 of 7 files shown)


csukuangfj:
I remember that for some streaming zipformer transducer model, if we also reset the encoder states, then it has trouble with recognizing the first word after reset.

KarelVesely84:
> I remember that for some streaming zipformer transducer model, if we also reset the encoder states, then it has trouble with recognizing the first word after reset.

aha, that could be an issue with masking-out the initial left context, initialized to 0.0 matices. in my model this is done inside the encoder model based on the `processed_lens` in streaming state. with zipformer, i think this was done externally in train.py , but i did not investigate into this in zipformer...

csukuangfj:
Can we add a config parameter to let the user decide whether the encode states should be reset on a detected endpoint？
We can set it to false by default。

KarelVesely84:
yes, that would be a good option fitting both ways, what would be the name ? `reset_encoder=T/F` ?

KarelVesely84:
Okay, the `reset_encoder` option in `OnlineRecognizerConfig` is now added and tested locally.

I found also an issue with `__init__.py` when some modules are disabled in cmake.
Those symbols are then missing in the `_sherpa_onnx*.so` file and an 
`import sherpa_onnx` in python fails.

csukuangfj:
> I am not sure how to solve this well. Either by tracking what was build.

How about something like
```cpp
#if SHERPA_ONNX_ENABLE_TTS == 1
  PybindOfflineTts(&m);
#else
  m.attr("OfflineTts") = py::none();
  m.attr("OfflineTtsConfig") = py::none();
  m.attr("OfflineTtsModelConfig") = py::none();
#endif
```

or we can add something like
```cpp
#define SHERPA_ONNX_DUMMY_IMPL(x) m.attr(#x) = py::none()

#if SHERPA_ONNX_ENABLE_TTS == 1
  PybindOfflineTts(&m);
#else
  SHERPA_ONNX_DUMMY_IMPL(OfflineTts);
  SHERPA_ONNX_DUMMY_IMPL(OfflineTtsConfig);
  SHERPA_ONNX_DUMMY_IMPL(OfflineTtsModelConfig);
#endif
```

KarelVesely84:
I have a question. Where should this be ? In the pybind11 code ?
Or directly in the `__init__.py` of the installed sherpa-onnx python module ?

Currently, if let's say TTS is deactivated, also the pybind11 wrapper is deactivated:
https://github.com/k2-fsa/sherpa-onnx/blob/921c4370e65d639e9200c9433472c0312eb62a3a/sherpa-onnx/python/csrc/CMakeLists.txt#L60

So, you suggest to activate all the pybind11 wrappers, and some symbols will be strings like "Deactivated in cmake." or the None symbols. And it will be driven by conditioning according to the `SHERPA_ONNX_ENABLE_*` macros.

Do I understand it correctly ?

K.

csukuangfj:
> I have a question. Where should this be ? In the pybind11 code ?


You can put it here
https://github.com/k2-fsa/sherpa-onnx/blob/921c4370e65d639e9200c9433472c0312eb62a3a/sherpa-onnx/python/csrc/sherpa-onnx.cc#L76-L79

or anywhere you think is appropriate, as long as it is in the C++ code.

---

> And the __init__.py will stay as it is now.



Yes, no need to change `__init__.py`.

The main idea is to set the attribute of  the `_sherpa` module so that when you use
```
from _sherpa import xxx
```
it will not throw.

---

in c++, if we use
```
m.attr("xxx") = 'yyy'
```
it is equivalent to
```
_sherpa.xxx = 'yyy'
```

We don't care what is assigned to `_sherpa.xxx`. What matters is `_sherpa` has an attribute called `xxx` so when we use
```
from _sherpa import xxx
```
it won't throw.

KarelVesely84:
I tried something in that direction.
There are 2 new files: 
[sherpa-onnx/python/csrc/faked-tts.cc](https://github.com/k2-fsa/sherpa-onnx/pull/2129/files#diff-390382f056af0b0ce0030b6b083e4874faa99abb5c507620a383cbd164bf89dd) 
[sherpa-onnx/python/csrc/faked-diarization.cc](https://github.com/k2-fsa/sherpa-onnx/pull/2129/files#diff-2b0efc167f543e01e47c98cad43e4dbfafc36244c9073456585afbef92e3fd85)

And they are enabled here in [python/csrc/CMakeLists.txt](https://github.com/k2-fsa/sherpa-onnx/pull/2129/files#diff-26d2cc2d2291892d8dcd3aaf7904481cb5cb6fd2dc0d3c9b8636586fefb59b43)

However, if I compile the project, it finished correctly and installed the package.
But, then when sourcing in python `import sherpa_onnx` i still get the error message:
```
ImportError: cannot import name 'FastClustering' from '_sherpa_onnx' (/mnt/matylda5/iveselyk/CNECT_TENDER/SHERPA_ONNX/CONDA_ENVIRONMENT/lib/python3.9/site-packages/_sherpa_onnx.cpython-39-x86_64-linux-gnu.so)
```

Is there some other place where the "faked" wrappers sholud be enabled besides the [python/csrc/CMakeLists.txt](https://github.com/k2-fsa/sherpa-onnx/pull/2129/files#diff-26d2cc2d2291892d8dcd3aaf7904481cb5cb6fd2dc0d3c9b8636586fefb59b43) ?

My belief is that the `sherpa-onnx-core` is first built independently of the pybind11 wrappers.
Next all the pybind11 wrappers are compiled, and these two components are forming the library `_sherpa_onnx.cpython-39-x86_64-linux-gnu.so`.

I tried to inspect the symbols inside with `nm -gDC _sherpa_onnx.cpython-39-x86_64-linux-gnu.so`, but I could not grep even those that are clearly there... (no `OnlineRecognizer`, no `py::` in the grep's)

KarelVesely84:
This is the example build log after changing the `faked-diarization.cc`:
```

**Comment by nshmyrev** [comment]
Related issue https://github.com/k2-fsa/sherpa-onnx/issues/1700

Btw, I might be wrong but two blanks is not what model expects, zipformer models take -1 and blank:

https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/pruned_transducer_stateless2/beam_search.py#L590

Two blanks are significantly worse for offline zipformer, not sure about streaming one.

**Comment by KarelVesely84** [comment]
sorry, the "close" was purely accidental

**Comment by KarelVesely84** [comment]
> Related issue #1700
> 
> Btw, I might be wrong but two blanks is not what model expects, zipformer models take -1 and blank:
> 
> https://github.com/k2-fsa/icefall/blob/master/egs/librispeech/ASR/pruned_transducer_stateless2/beam_search.py#L590
> 
> Two blanks are significantly worse for offline zipformer, not sure about streaming one.

aha, this is worth checking too, a comparison with the `decode.py` / `streaming_decode.py` behavior of the model...

File: sherpa-onnx/python/csrc/faked-tts.cc [hunk 1]
```diff
@@ -0,0 +1,47 @@
+// sherpa-onnx/python/csrc/faked-tts.cc
+//
+// Copyright (c)  2025  Brno University of Technology (author: Karel Vesely)
+
+#include "sherpa-onnx/python/csrc/offline-tts-kokoro-model-config.h"
+#include "sherpa-onnx/python/csrc/offline-tts-matcha-model-config.h"
+#include "sherpa-onnx/python/csrc/offline-tts-model-config.h"
+#include "sherpa-onnx/python/csrc/offline-tts-vits-model-config.h"
```
csukuangfj: can we remove these lines?

KarelVesely84: I am not sure. I tried to keep the same `.h` files and replace the internals of `.cc` files.
So these are irrelevant for the pybind11 API for python ?

KarelVesely84: I tried it, and rebuilt with a new `buid` dir. The error is still the same:
```
ImportError: cannot import name 'FastClustering' from '_sherpa_onnx' (/mnt/matylda5/iveselyk/CNECT_TENDER/SHERPA_ONNX/CONDA_ENVIRONMENT/lib/python3.9/site-packages/_sherpa_onnx.cpython-39-x86_64-linux-gnu.so)
```

csukuangfj:
You need to call the functions you add in sherpa-onnx.cc

Please have a look at my comment before.

It shows you where to put the code.

Put the  .cc files in CMakeLists.txt is far not enough.

KarelVesely84:
okay, now it is fixed, the empty symbols are hard-coded in `sherpa-onnx/python/csrc/sherpa-onnx.cc`
the `import sherpa_onnx` in python is now okay,
the decoding of 15min long recording with OnlineRecognizer was also okay.

(I should not try to program while being too tired after lunch ;-) )

Thank you for all the suggestions and patience,
K.

sherpa-onnx/python/csrc/online-recognizer.cc
```diff
@@ -67,7 +67,7 @@ static void PybindOnlineRecognizerConfig(py::module *m) {
            py::arg("max_active_paths") = 4, py::arg("hotwords_file") = "",
            py::arg("hotwords_score") = 0, py::arg("blank_penalty") = 0.0,
            py::arg("temperature_scale") = 2.0, py::arg("rule_fsts") = "",
-           py::arg("rule_fars") = "")
+           py::arg("rule_fars") = "", py::arg("reset_encoder"))
```
csukuangfj:
```suggestion
           py::arg("rule_fars") = "", py::arg("reset_encoder") = false)
```

Can you give it a default value?

csukuangfj (commented):
Thanks! Left a minor comment. Otherwise, it looks great to me.

sherpa-onnx/python/csrc/online-recognizer.cc
KarelVesely84: okay, default value is in place now

csukuangfj:

PR #2215, opened 2025-05-14
Expose dither for JNI

esavin:
This patch add dither to FeatureConfig java and expose it to JNI code

### DIFF
--- a/sherpa-onnx/java-api/src/com/k2fsa/sherpa/onnx/FeatureConfig.java
+++ b/sherpa-onnx/java-api/src/com/k2fsa/sherpa/onnx/FeatureConfig.java
@@ -6,10 +6,12 @@ package com.k2fsa.sherpa.onnx;
 public class FeatureConfig {
     private final int sampleRate;
     private final int featureDim;
+    private final float dither;
 
     private FeatureConfig(Builder builder) {
         this.sampleRate = builder.sampleRate;
         this.featureDim = builder.featureDim;
+        this.dither = builder.dither;
     }
 
     public static Builder builder() {
@@ -24,9 +26,14 @@ public class FeatureConfig {
         return featureDim;
     }
 
+   public float getDither() {
+        return dither;
+   }
+
     public static class Builder {
         private int sampleRate = 16000;
         private int featureDim = 80;
+        private float dither = 0.0f;
 
         public FeatureConfig build() {
             return new FeatureConfig(this);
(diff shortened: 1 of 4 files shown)


sherpa-onnx/java-api/src/com/k2fsa/sherpa/onnx/FeatureConfig.java
```diff
@@ -6,10 +6,12 @@
 public class FeatureConfig {
     private final int sampleRate;
     private final int featureDim;
+    private final float dither;
```
csukuangfj, line 9: Can you also update
https://github.com/k2-fsa/sherpa-onnx/blob/2e9e0b4e9e18aba667d7ca050ad83872af2c5740/sherpa-onnx/kotlin-api/FeatureConfig.kt#L3-L6

You can add a line
```kotlin
val dither: Float = 0.0f

**Review by csukuangfj — CHANGES_REQUESTED** [review]

File: sherpa-onnx/java-api/src/com/k2fsa/sherpa/onnx/FeatureConfig.java (hunk 1 above)
Comment (esavin) [reply sherpa-onnx/java-api/src/com/k2fsa/sherpa/onnx/FeatureConfig.java:9]: Done. 
May be it is necessary to add dither to fun getFeatureConfig(sampleRate: Int, featureDim: Int): FeatureConfig. But in this case it will need to add dither to all android examples.

Comment (csukuangfj) [reply sherpa-onnx/java-api/src/com/k2fsa/sherpa/onnx/FeatureConfig.java:9]: I think you can leave it as it is.

File: sherpa-onnx/kotlin-api/FeatureConfig.kt [hunk 2]
```diff
@@ -3,6 +3,7 @@ package com.k2fsa.sherpa.onnx
 data class FeatureConfig(
     var sampleRate: Int = 16000,
     var featureDim: Int = 80,
+    val dither: Float = 0.0f
```
csukuangfj: sorry, can you use var to replace val?

sherpa-onnx/java-api/src/com/k2fsa/sherpa/onnx/FeatureConfig.java
esavin: ok

sherpa-onnx/kotlin-api/FeatureConfig.kt
esavin: done

Issue #2148, opened 2025-04-25
support canary-180m-flash


jc955:
hello 
The canary-180m is ranked 9th in the https://huggingface.co/spaces/hf-audio/open_asr_leaderboard ASR ranking. With a size of only 700MB and very good transcription accuracy, it is very suitable for local transcription.

Is it possible to support this model onnx format? thanks!

csukuangfj:
Yes, it is possible. Please wait until we have some time, possibly next month.

lucaelin:
There are still some issues with this model for text longer than about 8 words.
- transcription: "Sherpa is the deployment framework of the Nexgen Caldi Project. Sherpa supports. Sherpa supports. Sherpa supports. Serpa. Serpa. Serpa. Serpa. Serpa. Serpa. Serpa. Serpa. Serpa. Serpa. Serpa. Serpa. Serpa." (and so on)
- transcription int8: "Sherpa is the deployment framework of the nextgen call deprock. Sherpa supports. Sherpa supports deploy Sherds deploy Sherds deploys deploys deploys deploys deploys deploys deploys deployments depments depments depments depments depments depments depments depments depments depments depments" (and so on) 
- original: "sherpa is the deployment framework of the Next-gen Kaldi project. sherpa supports deploying speech related pre-trained models on various platforms with various language bindings. If you are interested in how to train your own model or fine tune a pre-trained model, please refer to icefall."

Surprisingly it did much better in german, but punctuation was a bit weird.

csukuangfj:
Please show.your.code.

lucaelin:
The code I am using is the one added in this PR* https://github.com/k2-fsa/sherpa-onnx/blob/master/scripts/nemo/canary/test_180m_flash.py with models and tokens loaded from https://github.com/k2-fsa/sherpa-onnx/releases/download/asr-models/sherpa-onnx-nemo-canary-180m-flash-en-es-de-fr.tar.bz2 no other custom code. Tell me if you need audio samples, but from my testing any longer audio seems to produce incorrect transcriptions after a few words.

Edit: The [PR mentioned above](https://github.com/k2-fsa/sherpa-onnx/pull/2272)

csukuangfj:
Please.show.the.command for testing and share info.about which language your test wav contains.

lucaelin:
The command I used was 
```sh
[...]

PR #2348, opened 2025-07-06
fix(canary): use dynamo export, single input_ids and avoid 0/1 specialization

lucaelin:
Using dynamo export fixes issues related to the decoder producing garbage output on long sequences.
In order to use dynamo, the example input needs to match the runtime inputs, therefore requiring fixed size input_ids at runtime and (to avoid 0/1 specialization) a larger decoder_mems_list in the example. opset_version 18 is required for dynamo export. 

Additionally, I enabled int8 exports for the decoder, it seems to work just fine in my testing, while being twice as fast. The fp16 exports seem to be broken, though, this has been the case before as well.

Fixes https://github.com/k2-fsa/sherpa-onnx/issues/2148

### DIFF
--- a/scripts/nemo/canary/export_onnx_180m_flash.py
+++ b/scripts/nemo/canary/export_onnx_180m_flash.py
@@ -197,12 +197,12 @@ def export_decoder(canary_model):
     decoder = DecoderWrapper(canary_model)
     decoder_input_ids = torch.tensor([[1, 0]], dtype=torch.int32)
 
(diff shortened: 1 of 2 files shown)


csukuangfj (approved):

csukuangfj:
Thanks again! I've tested it locally and it works perfectly with the wave from
https://github.com/k2-fsa/sherpa-onnx/issues/2148#issuecomment-2955651791

Will add C++ runtime for it in the coming week.

PR #2352, opened 2025-07-06
Add C++ runtime and Python API for NeMo Canary models

csukuangfj:
From https://hf.co/nvidia/canary-180m-flash

> It supports automatic speech-to-text recognition (ASR) in 4 languages (English, German, French, Spanish) and translation from English to German/French/Spanish and from German/French/Spanish to English with or without punctuation and capitalization (PnC).

Example output from the python example:

```
en_wav_en_result  Ask not what your country can do for you. Ask what you can do for your country.
en_wav_es_result  No preguntes qué puede hacer tu país por ti. Pregúntale qué puedes hacer por tu país.
en_wav_de_result  Fragen Sie nicht, was Ihr Land für Sie tun kann. Fragen Sie, was Sie für Ihr Land tun können.
en_wav_fr_result  Demandez-vous ce que votre pays peut faire pour vous. Demandez ce que vous pouvez faire pour votre pays.
----------
de_wav_en_result  Everything has an end. Only the sausage has two
de_wav_de_result  Alles hat ein Ende, nur die Wurst hat zwei.
```


csukuangfj:
CC @lucaelin @jc955 @eqikkwkp25-cyber @rohithkodali

csukuangfj:
Please either re-download the onnx model or use the latest export script to re-export the model.

lucaelin:
I did a few more tests yesterday and there is still a fair bit of accuracy loss in the onnx version, which from what I can tell comes from the spectogram preprocessing. I ran out of time yesterday to fully debug this, but it seems like moving the preprocessor step from native fbank into the encoder-wrapper and using the NeMo preprocessor, I am getting the exact same results as using the NeMo/torch version directly. The difference is particularly strong on short audio, like single word audio. 
I hope I can look into this a bit further this evening but I don't think I know enough about native fbank implementation to get to the bottom of this.
So I guess we have three options: 
- keep it as is and accept the difference
- debug native fbank to better align with the original
- bake the preprocessor into onnx

csukuangfj:
> from what I can tell comes from the spectogram preprocessing

I think it is possible to make kaldi-native-fbank output identical fbank features as NeMo.

Can you post the code for computing features for kaldi-native-fbank and NeMo? We can help debug it.

Would be great if you could also share the test audio.

lucaelin:
> Can you post the code for computing features for kaldi-native-fbank and NeMo? We can help debug it.

Will do later today (CEST), but the fbank code is the same as in the `scripts/nemo/canary/test_180m_flash.py`. As for moving the preprocessing into the EncoderWrapper all I needed to add was a call to `preprocessor.get_features(x, x_len)` and passing audio to the model instead of features.

csukuangfj:
> all I needed to add was a call to preprocessor.get_features(x, x_len) and passing audio to the model instead of features.

I am not sure if it is possible since the preprocesor contains `stft`.

---

You may need to add two lines after
https://github.com/k2-fsa/sherpa-onnx/blob/0e738c356c9e42dda8119a243874d125ef581b58/scripts/nemo/canary/test_180m_flash.py#L167

```
 opts.mel_opts.hig_freq = 8000
 opts.frame_opts.snip_edges = False
```

csukuangfj:
Doc of this model can be found at 
https://k2-fsa.github.io/sherpa/onnx/nemo/canary.html

PR #2423, opened 2025-07-25
Implement max_symbols_per_frame for GigaAM2 

nshmyrev:
Accurate decoding requires max_symbols_per_frame since the model uses char tokens instead of BPE.

See here:

https://github.com/salute-developers/GigaAM/blob/main/gigaam/decoding.py#L103

WER improves significantly, 16.9 -> 14.2%


csukuangfj (approved):

PR #2454, opened 2025-08-06
Add APIs for Online NeMo CTC models

### LINKED ISSUE #2452: C-api doesn't support nemo ctc for online recognizer
Can't initialise sherpa with nemo ctc  online models in ios.

 
- CSRC code seems to be able to accept online nemo ctc for online models.

https://github.com/k2-fsa/sherpa-onnx/blob/ee9bd2553304dfa692f792ff4c5f7bf57f6079f2/sherpa-onnx/csrc/online-model-config.h#L16

- But the C-api doesn't expose these to be used:

https://github.com/k2-fsa/sherpa-onnx/blob/ee9bd2553304dfa692f792ff4c5f7bf57f6079f2/sherpa-onnx/c-api/c-api.h#L103

https://github.com/k2-fsa/sherpa-onnx/blob/ee9bd2553304dfa692f792ff4c5f7bf57f6079f2/sherpa-onnx/c-api/c-api.cc#L68

csukuangfj:
Fixes #2452 

CC @yagarwal1307

### DIFF
--- a/harmony-os/SherpaOnnxHar/sherpa_onnx/Index.ets
+++ b/harmony-os/SherpaOnnxHar/sherpa_onnx/Index.ets
@@ -25,6 +25,7 @@ export { Samples,
 } from './src/main/ets/components/NonStreamingAsr';
 
 export { OnlineStream,
+  OnlineNemoCtcModelConfig,
   OnlineTransducerModelConfig,
   OnlineParaformerModelConfig,
   OnlineZipformer2CtcModelConfig,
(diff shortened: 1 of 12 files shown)


yagarwal1307:
One more thing

- I see we have a support for  OnlineWenetCtcModelConfig as well in csrc too. I think it would be great if we fix that right now only. Otherwise will have to raise another pr in the future

yagarwal1307:
@csukuangfj Also currently im using the xcframework file in releases for it. Any chance I can get that, otherwise Ill have to change my code a lot for testing. Or we can write some sample swift program to test it independently

csukuangfj:
You need to build it from source or wait for our next release.


If you can release a streaming wenet ctc model.with.reasonable accuracy, we can support it. Otherwise, you need to modify the code by yourself.

yagarwal1307:
That makes sense. Ill try to build for the source then.
(Currently don't have a wenet model and aren't using it as well. I thought the code for it was missing by mistake so suggested the change.)
Thanks a lot for the help 😄!

PR #2480, opened 2025-08-10
Add more piper tts models

csukuangfj:
See also https://github.com/OHF-Voice/piper1-gpl/discussions/27

CC @JarbasAl

You can try them at
- https://k2-fsa.github.io/sherpa/onnx/tts/all/
- https://huggingface.co/spaces/k2-fsa/text-to-speech
- https://k2-fsa.github.io/sherpa/onnx/tts/apk-engine.html


JarbasAl:
want to point out i did not decide on a license for my trained models yet, likely will be a creative commons non commercial

without a license you shouldn't really be distributing them as you do not have explicit authorization for that...

csukuangfj:
> want to point out i did not decide on a license for my trained models yet, likely will be a creative commons non commercial
> 
> without a license you shouldn't really be distributing them as you do not have explicit authorization for that...

Thanks for pointing it out. I will delete them then.

JarbasAl:
> > want to point out i did not decide on a license for my trained models yet, likely will be a creative commons non commercial
> > without a license you shouldn't really be distributing them as you do not have explicit authorization for that...
> 
> Thanks for pointing it out. I will delete them then.

it's ok, i want them to be freely available, but a license will need to be added and distributed together with them

I am still deciding if i want to allow commercial usage or not, but for regular users they will always be freely available

for now i guess you can just add Attribution-NonCommercial-ShareAlike - CC BY-NC-SA 

if anything i will make them less restrictive in the future, not more

csukuangfj:
@JarbasAl 

Thanks! Please see #2530

Please let me know if any change is needed.

Issue #2517, opened 2025-08-22
wheel blocked as virus


DarkJukeMastah:
Hey there,

Not sure if this is you guys: https://pypi.org/project/sherpa-onnx/
But Windows Defender blocks the sherpa_onnx-1.12.9-cp311-cp311-win_amd64.whl file as a virus: Python/Multiverze!rfn

I haven't seen any other issues about it so it must be something new, just wanted to let you know.

richard-stanton:
Same issue, newer version:
Detected file: C:\Users\xxxxxxxxxxxxxx\AppData\Local\Temp\pip-unpack-3q6z7a74\sherpa_onnx-1.12.9-cp313-cp313-win_amd64.whl
Detected as: Exploit:Python/Multiverze!rfn

csukuangfj:
I am fixing it. Please wait.


As an alternative, you.can use

pip install "sherpa-onnx<1"

csukuangfj:
I have split the package into 3 small packages. Will release it in the next version 1.12.10

csukuangfj:
Please see
https://github.com/k2-fsa/sherpa-onnx/pull/2521

csukuangfj:
Fixed. Please try the latest version v1.12.10

<img width="1373" height="755" alt="Image" src="https://github.com/user-attachments/assets/9c513682-c995-4e41-9ae7-c8ab74c7402c" />

PR #2521, opened 2025-08-24
Split sherpa-onnx Python package

csukuangfj:
Issue #2517 reports that our Python package has been flagged as malware. I suspect this is because the wheel includes a large number of binary executables.

Another drawback of our current approach is that every wheel bundles the same binaries, even though they only depend on the operating system and not the Python ABI. This leads to significant duplication, wastes storage space, and makes it easy to hit PyPI’s free storage limit of 10 GB.

This PR splits the current package into three smaller ones. If you only use the Python API, you can continue installing sherpa-onnx in the same way as before—the change is completely transparent to you.

The following details are for those curious about the changes.

Now we have 3 packages:
  - sherpa-onnx-core
  - sherpa-onnx-bin
  - sherpa-onnx

# sherpa-onnx-core

It contains only the shared library files and the header files. You can install it with
```
pip install sherpa-onnx-core
```

Note that it is a dependency of `sherpa-onnx-bin` and `sherpa-onnx`. It is automagically installed if you install `sherpa-onnx-bin` or `sherpa-onnx`.

Its whl filename looks like the following: 
```
sherpa_onnx_core-x.y.z-py3-none-macosx_10_15_x86_64.whl
```

You can see it contains `py3-none`, which means it does not depend on the Python ABI.

After installing, you can run
```
python3 -m sherpa_onnx --cflags

python3 -m sherpa_onnx --c-api-libs
python3 -m sherpa_onnx --c-api-libs-only-L
python3 -m sherpa_onnx --c-api-libs-only-l

python3 -m sherpa_onnx --cxx-api-libs
python3 -m sherpa_onnx --cxx-api-libs-only-L
python3 -m sherpa_onnx --cxx-api-libs-only-l
```

# sherpa-onnx-bin
It contains pre-compiled binaries. You can use

```
pip install sherpa-onnx-bin
```

to install it.

After installing, you can run the following in your terminal

```
sherpa-onnx-version
sherpa-onnx --help
sherpa-onnx-offline --help
... ...
```

Note that the commands `sherpa-onnx-version`, `sherpa-onnx`, `sherpa-onnx-offline`, etc, are automagically available without changing your `PATH` environment variable.

# sherpa-onnx
It contains our Python API.

Install it with
```
pip install sherpa-onnx
```

## Summary by CodeRabbit

* **New Features**
  * Split packaging into sherpa-onnx-core (libs/headers) and sherpa-onnx-bin (CLI tools); added python -m sherpa_onnx helper for build/link flags.

* **Improvements**
  * Prebuilt wheels broadened (Linux x86/aarch64, macOS x86/arm64/universal2, Windows x86/x64) with updated naming and wider Python support (including 3.14).
  * CI publishing: Hugging Face by default and optional gated PyPI publishing.

* **Other**
  * Added four Polish Piper voice models to Hugging Face.

* **Breaking Changes**
  * Minimum Python raised to 3.7.


csukuangfj:
To try it before we release a new version, you can run
```
pip install "sherpa-onnx<1" sherpa-onnx-bin
```

scripts/wheel/sherpa-onnx-core/sherpa_onnx/__main__.py
```diff
@@ -0,0 +1,35 @@
+import sys
```
csukuangfj, line 1: Note that `__main__.py` and `_info.py` are generated by chatgpt. I did only some minor changes.

Issue #2435, opened 2025-07-29
No online / streaming models with ru language?


MrBzik:
Hi, I am stuck with the docs, it seems as if all the models that support Russian language are located in offline category. I need real-time speech-to-text utility. Am I missing something? 
Trying to convert those models to OnlineModelConfig gives no good.

nshmyrev:
https://huggingface.co/alphacep/vosk-model-small-streaming-ru/tree/main

MrBzik:
Oh, thanks. Yes, that works. For some reason I limited myself to the models present in the docs.

csukuangfj:
@MrBzik 
We have added a new Russian streaming ASR model.

Please see
https://k2-fsa.github.io/sherpa/onnx/pretrained_models/online-ctc/t-one-ctc-models.html#sherpa-onnx-streaming-t-one-russian-2025-09-08-russian

It is from
https://github.com/voicekit-team/T-one

MrBzik:
@csukuangfj 

I tried out prebuilt apk with this model on 11's Android (30 SDK). 
[sherpa-onnx-1.12.12-arm64-v8a-asr-ru-t_one_ctc_2025_09_08.apk](https://huggingface.co/csukuangfj/sherpa-onnx-apk/resolve/main/asr/1.12.12/sherpa-onnx-1.12.12-arm64-v8a-asr-ru-t_one_ctc_2025_09_08.apk)

It is not working. It runs infinite loop of attempts to start native process I guess. The logs are: 

`
Start to initialize model
Select model type 27
OnlineRecognizerConfig...
We expect that tokens.txt contains the symbol <blk> or <eps> or <blank> and its ID.`

csukuangfj:
Can you post complete logs? @MrBzik

MrBzik:
I started with replacing old apk with the new one, which has no described issue. Sorry

PR #2590, opened 2025-09-11
Fix initializing symbol table for OnlineRecognizer.

csukuangfj:
See also https://github.com/k2-fsa/sherpa-onnx/issues/2435#issuecomment-3280606973

@MrBzik Please re-download the APK once 
https://github.com/csukuangfj/sherpa-onnx/actions/runs/17649596976
finishes.  It may take 30 miniutes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected tokenizer configuration precedence: when an in-memory token buffer is provided, it is now reliably used; otherwise the token list from configuration is applied. This ensures consistent symbol table initialization and avoids unexpected overrides.
  * No public API changes.

### DIFF
--- a/sherpa-onnx/csrc/online-recognizer-ctc-impl.h
+++ b/sherpa-onnx/csrc/online-recognizer-ctc-impl.h
@@ -79,6 +79,7 @@ class OnlineRecognizerCtcImpl : public OnlineRecognizerImpl {
       : OnlineRecognizerImpl(config),
         config_(config),
         model_(OnlineCtcModel::Create(config.model_config)),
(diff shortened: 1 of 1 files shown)


MrBzik:
Hi, I tried new apk. Initialization of the model is successful. Thank you. 
Though the model itself works too slow for streaming recognition. Output precision is low as well.

csukuangfj:
> Hi, I tried new apk. Initialization of the model is successful. Thank you. Though the model itself works too slow for streaming recognition. Output precision is low as well.

Thank you for testing it.

PR #2604, opened 2025-09-17
Build shell script for node addon

colourmebrad:
`build-node-addon.sh` builds into
* `./build/` the shared library
* `./scripts/node-addon-api/lib/` the js wrappers
* `./scripts/node-addon-api/build/Release/sherpa-onnx.node` the addon

I also had to update .gitignore to not ignore build-*.sh files


csukuangfj:

We have already provided a script at
https://github.com/k2-fsa/sherpa-onnx/blob/master/scripts/node-addon-api/run.sh

Is it possible to use that without creating a new script?

colourmebrad:
> Thank you for your contribution!
> 
> We have already provided a script at https://github.com/k2-fsa/sherpa-onnx/blob/master/scripts/node-addon-api/run.sh
> 
> Is it possible to use that without creating a new script?

Totally missed that one, my bad! I'll play around and, assuming it works, I'll close the PR. Thanks 👍

PR #2627, opened 2025-09-24
fix: hot update language for sencevoice

o-u-p:
since the parent class OfflineRecognizerImpl already have a config, I think OfflineRecognizerSenseVoiceImpl can resue it, if not so, the recognizer.setConfig will not work.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined internal configuration handling within offline recognition components to improve extensibility and maintainability.
  * Adjusted access levels to better support derived implementations without changing external behavior.
  * No impact on features, performance, or UI.
  * No API changes; existing integrations continue to work as before.
  * No user action required.


csukuangfj (approved):

PR #2645, opened 2025-10-04
Fix ios build script

vovasty:
# Problem
Swift Packages which use `sherpa-onnx` can't find headers and emit errors about missing prefix `lib` in library names.

# Root cause
iOS build script copies headers into `sherpa-onnx.xcframework/Headers`, but it should copy them for each sdk.

# Solution
* Use parameter `-headers` for `xcodebuild -create-xcframework` command so headers are copied for each arch.
* rename `sherpa-onnx.a` into `libsherpa-onnx.a`
* Remove `HEADER_SEARCH_PATHS` setting from Xcode projects cause it not necessary anymore.

## Summary by CodeRabbit

- New Features
  - Simplified iOS integration: headers are now bundled with the framework, reducing manual setup.
- Bug Fixes
  - Fewer build/link issues on iOS by standardizing library naming and header handling.
- Chores
  - Streamlined iOS build process and artifact generation.
  - Removed custom header search paths from example projects to rely on framework-provided headers.

### DIFF
--- a/build-ios.sh
+++ b/build-ios.sh
@@ -138,7 +138,7 @@ done
 
 # Merge archive first, because the following xcodebuild create xcframework
 # cannot accept multi archive with the same architecture.
-libtool -static -o build/simulator/sherpa-onnx.a \
+libtool -static -o build/simulator/libsherpa-onnx.a \
   build/simulator/lib/libcppinyin_core.a \
   build/simulator/lib/libkaldi-native-fbank-core.a \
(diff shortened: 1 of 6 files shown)


csukuangfj:

Have you tested it and does it work for you?

vovasty:
yes, it works perfectly. Now you can use it in swift packages.

csukuangfj:
> yes, it works perfectly. Now you can use it in swift packages.

Great! Thanks!

PR #2487, opened 2025-08-13
[WIP] Add Zipvoice

pkufool:

PR #2646, opened 2025-10-04
Fix iOS example apps build

vovasty:
#2487 has added `zipvoice` parameter to `SherpaOnnxOfflineTtsModelConfig` which breaks iOS example apps building.
Fix the issue by adding default parameter to `sherpaOnnxOfflineTtsModelConfig`.

### DIFF
--- a/swift-api-examples/SherpaOnnx.swift
+++ b/swift-api-examples/SherpaOnnx.swift
@@ -920,6 +920,7 @@ func sherpaOnnxOfflineTtsModelConfig(
   vits: SherpaOnnxOfflineTtsVitsModelConfig = sherpaOnnxOfflineTtsVitsModelConfig(),
   matcha: SherpaOnnxOfflineTtsMatchaModelConfig = sherpaOnnxOfflineTtsMatchaModelConfig(),
   kokoro: SherpaOnnxOfflineTtsKokoroModelConfig = sherpaOnnxOfflineTtsKokoroModelConfig(),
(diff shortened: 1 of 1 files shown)


swift-api-examples/SherpaOnnx.swift
```diff
@@ -920,6 +920,7 @@ func sherpaOnnxOfflineTtsModelConfig(
   vits: SherpaOnnxOfflineTtsVitsModelConfig = sherpaOnnxOfflineTtsVitsModelConfig(),
   matcha: SherpaOnnxOfflineTtsMatchaModelConfig = sherpaOnnxOfflineTtsMatchaModelConfig(),
   kokoro: SherpaOnnxOfflineTtsKokoroModelConfig = sherpaOnnxOfflineTtsKokoroModelConfig(),
+  zipvoice: SherpaOnnxOfflineTtsZipvoiceModelConfig = SherpaOnnxOfflineTtsZipvoiceModelConfig(),
```
csukuangfj, line 923: Can you follow our convention to add a function
```
sherpaOnnxOfflineTtsZipvoiceModelConfig
```

Note it starts with the lowercase `s`.

csukuangfj (changes requested):

swift-api-examples/SherpaOnnx.swift
csukuangfj: Fixed in the latest master.

PR #2834, opened 2025-11-28
[ZipVoice] Fix english tokenization error

pkufool:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected English language text processing in text-to-speech to consistently use the proper phoneme set for English segments during voice synthesis.

### DIFF
--- a/sherpa-onnx/csrc/offline-tts-zipvoice-frontend.cc
+++ b/sherpa-onnx/csrc/offline-tts-zipvoice-frontend.cc
@@ -342,7 +342,7 @@ std::vector<TokenIDs> OfflineTtsZipvoiceFrontend::ConvertTextToTokenIds(
       TokenizeZh(pt.first, pinyin_encoder_.get(), token2id_, &token_ids,
                  &tokens);
     } else if (pt.second == "en") {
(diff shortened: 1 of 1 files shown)


sherpa-onnx/csrc/offline-tts-zipvoice-frontend.cc
```diff
@@ -342,7 +342,7 @@ std::vector<TokenIDs> OfflineTtsZipvoiceFrontend::ConvertTextToTokenIds(
       TokenizeZh(pt.first, pinyin_encoder_.get(), token2id_, &token_ids,
                  &tokens);
     } else if (pt.second == "en") {
-      TokenizeEn(pt.first, token2id_, voice, &token_ids, &tokens);
+      TokenizeEn(pt.first, token2id_, "en-us", &token_ids, &tokens);
```
csukuangfj, line 345: What was `voice` before the change?

pkufool: It is passed in as an argument (default value is ""),  we know the text here is english, so can use "en-us" directly. The function is inherited from a base class, so I did not remove the "voice" arugment, we may use it when zipvoice support more languages.

PR #2862, opened 2025-12-04
Fix the discrepancy with the Silero VAD isSpeech logic

ming030890:
Source: https://github.com/snakers4/silero-vad/blob/63fe03add76a5f9a96d2fd339fc53b672c01c536/src/silero_vad/utils_vad.py#L342-L343

The issue is that when the threshold is < 0.15, threshold - 0.15 becomes negative, which means the VAD never exits the is_speech state.
In the original Silero implementation, the negative threshold is clamped to a minimum of 0.01 to prevent this problem.

## Summary by CodeRabbit

* **New Features**
  * Added configurable threshold parameter for speech detection adjustment.
  * Updated speech detection logic to support dynamic threshold configuration with built-in safety constraints.

### DIFF
--- a/sherpa-onnx/csrc/silero-vad-model.cc
+++ b/sherpa-onnx/csrc/silero-vad-model.cc
@@ -4,6 +4,7 @@
 
 #include "sherpa-onnx/csrc/silero-vad-model.h"
 
+#include <algorithm>
 #include <memory>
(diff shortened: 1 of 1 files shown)


ming030890:
sorry this is a build error. I will fix it before creating another PR.

PR #2936, opened 2025-12-25
Add funASR-Nano support

Wasser1462:
# FunASR-Nano Offline ASR Integration

## Summary

This PR adds/updates **FunASR-Nano** offline ASR integration in **sherpa-onnx**, including tokenizer integration, model config wiring, and example commands.

The goal is to make FunASR-Nano usable via `sherpa-onnx-offline` and `sherpa-onnx-vad-with-offline-asr` with clear guidance on which precision is recommended on GPU/CPU.

## Key Changes

- Add/Update FunASR-Nano offline model config and wiring for:
  - encoder adaptor
  - LLM prefill/decode (KV-cache style)
  - embedding
  - tokenizer integration

## Model Download (ModelScope)

Pre-exported models are uploaded to ModelScope and can be downloaded with:

```bash
modelscope download --model zengshuishui/FunASR-nano-onnx-V2 --output_dir ./models
```

## Recommended Runtime / Performance Notes

Based on local benchmarking:

### GPU (CUDA): Recommend FP32

- Tested on RTX 4090
- RTF ≈ 0.039
gpu

### CPU: Recommend INT8

- Tested on Intel(R) Xeon(R) Gold 6448Y
- RTF ≈ 0.068
cpu

### FP16

- CPU inference works
- GPU FP16 inference is not supported yet (time constraints and pending investigation)

### 1) FP32 on GPU (Recommended for CUDA)

```bash
./bin/sherpa-onnx-offline \
  --funasr-nano-encoder-adaptor=$MODEL_DIR/encoder_adaptor.onnx \
  --funasr-nano-llm-prefill=$MODEL_DIR/llm_prefill.onnx \
  --funasr-nano-llm-decode=$MODEL_DIR/llm_decode.onnx \
  --funasr-nano-embedding=$MODEL_DIR/embedding.onnx \
  --funasr-nano-tokenizer=$MODEL_DIR/Qwen3-0.6B \
  --provider=cuda \
  --num-threads=4 \
  $WAV_ZH
```

### 2) INT8 (Recommended for CPU)

```bash
./bin/sherpa-onnx-offline \
  --funasr-nano-encoder-adaptor=$MODEL_DIR/encoder_adaptor.int8.onnx \
  --funasr-nano-llm-prefill=$MODEL_DIR/llm_int8/llm_prefill.int8.onnx \
  --funasr-nano-llm-decode=$MODEL_DIR/llm_int8/llm_decode.int8.onnx \
  --funasr-nano-embedding=$MODEL_DIR/embedding.int8.onnx \
  --funasr-nano-tokenizer=$MODEL_DIR/Qwen3-0.6B \
  --provider=cpu \
  --num-threads=4 \
  $WAV_ZH
```

### 3) VAD + INT8 on CPU (Recommended)

```bash
./bin/sherpa-onnx-vad-with-offline-asr \
  --silero-vad-model=$VAD_MODEL \
  --funasr-nano-encoder-adaptor=$MODEL_DIR/encoder_adaptor.int8.onnx \
  --funasr-nano-llm-prefill=$MODEL_DIR/llm_int8/llm_prefill.int8.onnx \
  --funasr-nano-llm-decode=$MODEL_DIR/llm_int8/llm_decode.int8.onnx \
  --funasr-nano-embedding=$MODEL_DIR/embedding.int8.onnx \
  --funasr-nano-tokenizer=$MODEL_DIR/Qwen3-0.6B \
  --provider=cpu \
  --num-threads=4 \
  $WAV_SONG
```

### 4) VAD + FP32 on GPU (Example)

```bash
./bin/sherpa-onnx-vad-with-offline-asr \
  --silero-vad-model=$VAD_MODEL \
  --funasr-nano-encoder-adaptor=$MODEL_DIR/encoder_adaptor.onnx \
  --funasr-nano-llm-prefill=$MODEL_DIR/llm_fp32/llm_prefill.fp32.onnx \
  --funasr-nano-llm-decode=$MODEL_DIR/llm_fp32/llm_decode.fp32.onnx \
  --funasr-nano-embedding=$MODEL_DIR/embedding.onnx \
  --funasr-nano-tokenizer=$MODEL_DIR/Qwen3-0.6B \
  --provider=cuda \
  --num-threads=4 \
  $WAV_SONG
```

## Testing

Ran basic CLI sanity tests with:

- `./bin/sherpa-onnx-offline`
- `./bin/sherpa-onnx-vad-with-offline-asr`

Verified output correctness on CPU and CUDA environments as described above.

## Notes / Disclaimer

This work was completed with assistance from AI tools. If you spot any issues or have improvement suggestions, please let me know — I will follow up quickly.

## Summary by CodeRabbit

* **New Features**
  * FunASR‑nano offline speech recognition: KV‑cache, embedding support, integrated tokenizer, and half‑precision/float conversion helpers for better model handling.
  * New example tools: a C++ executable and a Python CLI script for offline decoding with configurable generation parameters, progress/timing output, and multi‑file decoding.

* **Documentation**
  * CLI usage/help updated with FunASR‑nano examples and configuration guidance.

### DIFF
--- a/python-api-examples/offline-funasr-nano-decode-files.py
+++ b/python-api-examples/offline-funasr-nano-decode-files.py
@@ -1,4 +1,7 @@
 #!/usr/bin/env python3
+#
+# Copyright (c)  2025  zengyw
+#
 """
 Decode audio files using FunASR-nano models with sherpa-onnx Python API.
 
@@ -19,6 +22,8 @@ Usage:
 import argparse
 import sys
 from pathlib import Path
+
+import soundfile as sf
 
 try:
     import sherpa_onnx
@@ -144,46 +149,35 @@ def get_args():
 
 
 def create_recognizer(args) -> sherpa_onnx.OfflineRecognizer:
-    """Create an offline recognizer for FunASR-nano models."""
-    config = sherpa_onnx.OfflineRecognizerConfig(
-        feat_config=sherpa_onnx.FeatureExtractorConfig(
-            sampling_rate=16000,
-            feature_dim=80,
-        ),
-        model_config=sherpa_onnx.OfflineModelConfig(
-            funasr_nano=sherpa_onnx.OfflineFunASRNanoModelConfig(
-                encoder_adaptor=args.encoder_adaptor,
-                llm_prefill=args.llm_prefill,
-                llm_decode=args.llm_decode,
-                embedding=args.embedding,
-                tokenizer=args.tokenizer,
-                system_prompt=args.system_prompt,
-                user_prompt=args.user_prompt,
-                max_new_tokens=args.max_new_tokens,
-                temperature=args.temperature,
-                top_p=args.top_p,
-                seed=args.seed,
-            ),
-            num_threads=args.num_threads,
-            debug=args.debug,
-            provider=args.provider,
-        ),
-        decoding_method="greedy_search",
+    return sherpa_onnx.OfflineRecognizer.from_funasr_nano(
+        encoder_adaptor=args.encoder_adaptor,
+        llm_prefill=args.llm_prefill,
+        llm_decode=args.llm_decode,
+        embedding=args.embedding,
+        tokenizer=args.tokenizer,
+        num_threads=args.num_threads,
+        provider=args.provider,
+        debug=args.debug,
+        system_prompt=args.system_prompt,
+        user_prompt=args.user_prompt,
+        max_new_tokens=args.max_new_tokens,
+        temperature=args.temperature,
+        top_p=args.top_p,
+        seed=args.seed,
     )
-
-    recognizer = sherpa_onnx.OfflineRecognizer(config)
-    return recognizer
 
 
 def decode_file(
     recognizer: sherpa_onnx.OfflineRecognizer,
     filename: str,
-) -> sherpa_onnx.OfflineRecognitionResult:
+):
     """Decode a single audio file."""
-    wave = sherpa_onnx.read_wave(filename)
+    audio, sample_rate = sf.read(filename, dtype="float32", always_2d=True)
+    audio = audio[:, 0]  # only use the first channel
+
     stream = recognizer.create_stream()
-    stream.accept_waveform(wave.sample_rate, wave.samples)
-    recognizer.decode(stream)
+    stream.accept_waveform(sample_rate, audio)
+    recognizer.decode_stream(stream)
     result = stream.result
     return result
 
--- a/sherpa-onnx/csrc/funasr-nano-tokenizer.cc
+++ b/sherpa-onnx/csrc/funasr-nano-tokenizer.cc
@@ -1,4 +1,6 @@
 // sherpa-onnx/csrc/funasr-nano-tokenizer.cc
+//
+// Copyright (c)  2025  zengyw
 
 #include "sherpa-onnx/csrc/funasr-nano-tokenizer.h"
 
@@ -785,7 +787,7 @@ static inline std::string MakeMergeKey(const std::string &a,
 }  // namespace
 
 // Parse tokenizer.json added_tokens: extract objects with {id, content, ...}
-bool ParseAddedTokensFromTokenizerJson(
+static bool ParseAddedTokensFromTokenizerJson(
     const std::string &blob, std::vector<FunASRNanoTokenizer::AddedToken> *out) {
   if (!out) return false;
   out->clear();
@@ -871,7 +873,7 @@ void BuildAddedTokensTrie(
   }
 }
 
-void MergeVocabAndAddedTokens(
+static void MergeVocabAndAddedTokens(
     std::unordered_map<std::string, int32_t> *vocab,
     const std::vector<FunASRNanoTokenizer::AddedToken> &added,
     std::unordered_set<std::string> *added_contents) {
--- a/sherpa-onnx/csrc/funasr-nano-tokenizer.h
+++ b/sherpa-onnx/csrc/funasr-nano-tokenizer.h
@@ -1,11 +1,12 @@
 // sherpa-onnx/csrc/funasr-nano-tokenizer.h
 //
+// Copyright (c)  2025  zengyw
+//
 // A self-contained Qwen3 ByteLevel-BPE tokenizer implementation.
 // - No dependency on tokenizers-cpp / HF tokenizers
 // - Loads vocab.json + merges.txt + tokenizer.json(added_tokens)
 // - Supports AddedTokens via Trie longest-match
 // - ByteLevel bytes_to_unicode encode/decode
-// Author: zengyongwang
 
 #ifndef SHERPA_ONNX_CSRC_FUNASR_NANO_TOKENIZER_H_
 #define SHERPA_ONNX_CSRC_FUNASR_NANO_TOKENIZER_H_
@@ -87,7 +88,7 @@ class FunASRNanoTokenizer {
   std::unordered_map<std::string, int32_t> token2id_;
   std::vector<std::string> id2token_;
 
-  // merges ranks: "left\t right" -> rank
+  // merges ranks: "left\tright" -> rank
   std::unordered_map<std::string, int32_t> merges_rank_;
 
   // BPE cache: bytelevel_word -> list of merged tokens
--- a/sherpa-onnx/csrc/offline-funasr-nano-model-config.h
+++ b/sherpa-onnx/csrc/offline-funasr-nano-model-config.h
@@ -1,4 +1,6 @@
 // sherpa-onnx/csrc/offline-funasr-nano-model-config.h
+//
+// Copyright (c)  2025  zengyw
 
 #ifndef SHERPA_ONNX_CSRC_OFFLINE_FUNASR_NANO_MODEL_CONFIG_H_
 #define SHERPA_ONNX_CSRC_OFFLINE_FUNASR_NANO_MODEL_CONFIG_H_
--- a/sherpa-onnx/csrc/offline-funasr-nano-model.cc
+++ b/sherpa-onnx/csrc/offline-funasr-nano-model.cc
@@ -1,4 +1,6 @@
 // sherpa-onnx/csrc/offline-funasr-nano-model.cc
+//
+// Copyright (c)  2025  zengyw
 
 #include "sherpa-onnx/csrc/offline-funasr-nano-model.h"
 
@@ -66,6 +68,12 @@ static inline bool IsCudaProvider(const std::string &provider) {
   // Keep it conservative. We only enable IO binding policy below when we
   // are on CUDA; other EPs keep the existing behavior.
   return p == "cuda" || (p.size() > 4 && p.find("cuda") == 0);
+}
+
+// Check if a tensor element type is FP16-IO (float16 or uint16).
+static inline bool IsFloat16IO(ONNXTensorElementDataType t) {
+  return t == ONNX_TENSOR_ELEMENT_DATA_TYPE_FLOAT16 ||
+         t == ONNX_TENSOR_ELEMENT_DATA_TYPE_UINT16;
 }
 
 // Get the element type of a session input tensor.
@@ -261,6 +269,22 @@ class OfflineFunASRNanoModel::Impl {
       // configures the CUDA EP device; binding here only affects output memory.
       cuda_mem_info_ = std::make_unique<Ort::MemoryInfo>(
           "Cuda", OrtDeviceAllocator, 0, OrtMemTypeDefault);
+
+      // Check if prefill/decode models have FP16-IO, which is not supported on CUDA yet.
+      ONNXTensorElementDataType prefill_in_type = prefill_embeds_in_type_;
+      ONNXTensorElementDataType prefill_out_type =
+          GetSessionOutputElemType(prefill_sess_.get(), 0);
+      ONNXTensorElementDataType decode_in_type = decode_embeds_in_type_;
+      ONNXTensorElementDataType decode_out_type =
+          GetSessionOutputElemType(decode_sess_.get(), 0);
+
+      if (IsFloat16IO(prefill_in_type) || IsFloat16IO(prefill_out_type) ||
+          IsFloat16IO(decode_in_type) || IsFloat16IO(decode_out_type)) {
+        SHERPA_ONNX_LOGE(
+            "fp16-IO LLM models are not supported on CUDA yet. Please use "
+            "fp32/int8 models.");
+        SHERPA_ONNX_EXIT(-1);
+      }
     }
   }
 
@@ -413,6 +437,22 @@ class OfflineFunASRNanoModel::Impl {
     if (use_cuda_iobinding_) {
       cuda_mem_info_ = std::make_unique<Ort::MemoryInfo>(
           "Cuda", OrtDeviceAllocator, 0, OrtMemTypeDefault);
+
+      // Check if prefill/decode models have FP16-IO, which is not supported on CUDA yet.
+      ONNXTensorElementDataType prefill_in_type = prefill_embeds_in_type_;
+      ONNXTensorElementDataType prefill_out_type =
+          GetSessionOutputElemType(prefill_sess_.get(), 0);
+      ONNXTensorElementDataType decode_in_type = decode_embeds_in_type_;
+      ONNXTensorElementDataType decode_out_type =
+          GetSessionOutputElemType(decode_sess_.get(), 0);
+
+      if (IsFloat16IO(prefill_in_type) || IsFloat16IO(prefill_out_type) ||
+          IsFloat16IO(decode_in_type) || IsFloat16IO(decode_out_type)) {
+        SHERPA_ONNX_LOGE(
+            "fp16-IO LLM models are not supported on CUDA yet. Please use "
+            "fp32/int8 models.");
+        SHERPA_ONNX_EXIT(-1);
+      }
     }
   }
 
--- a/sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.cc
+++ b/sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.cc
@@ -1,4 +1,6 @@
 // sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.cc
+//
+// Copyright (c)  2025  zengyw
 
 #include "sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.h"
 
@@ -51,7 +53,7 @@ OfflineRecognizerFunASRNanoImpl::CreateStream() const {
 // Sets normalization, window type, and disables edge snipping and dithering
 // to match the model's expected input format.
 void OfflineRecognizerFunASRNanoImpl::InitFeatConfig() {
-  config_.feat_config.normalize_samples = true;
+  config_.feat_config.normalize_samples = false;
   config_.feat_config.window_type = "hamming";
   config_.feat_config.snip_edges = false;
   config_.feat_config.dither = 0.0f;
@@ -163,9 +165,15 @@ OfflineRecognitionResult OfflineRecognizerFunASRNanoImpl::GenerateText(
   int32_t context_len = static_cast<int32_t>(source_ids.size());
   const int32_t max_seq_len = 2048;
   if (context_len > max_seq_len) {
+    SHERPA_ONNX_LOGE(
+        "Input sequence length (%d) exceeds maximum sequence length (%d). "
+        "Truncating to %d tokens. Recognition result may be incomplete due to "
+        "truncated input.",
+        context_len, max_seq_len, max_seq_len);
     source_ids.resize(max_seq_len);
     context_len = max_seq_len;
   }
+
   // Get text embeddings for the prompt tokens
   std::vector<int64_t> input_ids = source_ids;
   std::array<int64_t, 2> ids_shape{1, context_len};
@@ -236,8 +244,7 @@ OfflineRecognitionResult OfflineRecognizerFunASRNanoImpl::GenerateText(
   generated_ids.reserve(funasr_config.max_new_tokens);
   const int64_t eos_id = tokenizer_->GetEosTokenId();
   const int64_t im_end_id = tokenizer_->GetImEndTokenId();
-  const int32_t max_new_tokens =
-      funasr_config.max_new_tokens > 0 ? funasr_config.max_new_tokens : 256;
+  const int32_t max_new_tokens = funasr_config.max_new_tokens;
   std::vector<std::pair<Ort::Value, Ort::Value>> past_key_values;
   bool is_first_step = true;
   // Autoregressive generation loop
--- a/cxx-api-examples/funasr-nano-cxx-api.cc
+++ b/cxx-api-examples/funasr-nano-cxx-api.cc
@@ -1,5 +1,7 @@
 // cxx-api-examples/funasr-nano-cxx-api.cc
 //
+// Copyright (c)  2025  zengyw
+//
 // This file demonstrates how to use FunASR-nano with sherpa-onnx's C++ API.
 //
 // clang-format off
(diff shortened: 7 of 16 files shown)


csukuangfj:
Screenshot 2025-12-25 at 17 57 16


There is no `llm_prefill.int8.onnx` and  `llm_decode.int8.onnx` in the above modelscope repo for the following arguments:

```
  --funasr-nano-llm-prefill=$MODEL_DIR/llm_int8/llm_prefill.int8.onnx \
  --funasr-nano-llm-decode=$MODEL_DIR/llm_int8/llm_decode.int8.onnx \
```

Can you update the modelscope or the commandline?

csukuangfj:
I see. It is in
https://www.modelscope.cn/models/zengshuishui/FunASR-nano-onnx-V2/tree/master/llm_int8

Please ignore the previous comment.

csukuangfj:
Can you add
https://github.com/k2-fsa/sherpa-onnx/blob/1f5f9633cc71f0426cbcea591fba56ef8a646984/scripts/medasr/export_onnx.py#L100
when exporting models to ONNX?

It simplifies the deployment for Android using assetsif no external data is used.

Wasser1462:
Yes, I can do that. I’ll re-export the ONNX models with `external_data=False` and make the necessary adaptations so the models are packaged as a single `.onnx` file (no external `.data`). This should simplify Android deployment via assets.

sherpa-onnx/csrc/offline-funasr-nano-model.cc
```diff
@@ -0,0 +1,973 @@
...
+inline const char *ToCStr<const char *>(const char *const &x) {
+  return x;
+}
+
+// Check if a tensor is stored on CUDA device.
+static inline bool TensorIsCuda(const Ort::Value &v) {
+  if (!v.IsTensor()) return false;
+  auto mi = v.GetTensorMemoryInfo();
+  auto name_obj = mi.GetAllocatorName();
+  const char *name = ToCStr(name_obj);
```
csukuangfj: Do you need to free `name` to avoid memory leak?

Also, instead of comparing name, can you compare
```
return mi.GetDeviceType() == OrtMemoryInfoDeviceType_GPU && mi.GetDeviceId() >= 0;
```

You can remove the template `ToCStr` in this case.

sherpa-onnx/csrc/offline-funasr-nano-model.cc
```diff
@@ -0,0 +1,973 @@
...
+  if (!name || !name[0]) return false;
+  return std::strstr(name, "Cuda") || std::strstr(name, "CUDA");
+}
+
+// Copy tensor data from GPU to CPU if needed.
+// Uses dynamic loading of CUDA functions to avoid compile-time dependencies.
+static inline void CopyRawToCpu(void *dst, const Ort::Value &src, size_t bytes,
```
csukuangfj: Is this needed?
What is the disadvantage of following the same approach we use for other models—letting ONNX Runtime automatically move tensors between CPU and GPU?
The current implementation seems unnecessarily complicated.

sherpa-onnx/csrc/offline-funasr-nano-model.cc
Wasser1462: Thanks for the suggestions.

-No, I don’t need to free `name`. `GetAllocatorName()` returns a `std::string` (value type), so it’s managed automatically and won’t cause a leak.
-Agreed on the device-type/device-id check. It’s cleaner and more reliable than string matching. I’ll switch to that approach and remove the string-based helper logic accordingly.

sherpa-onnx/csrc/offline-funasr-nano-model.cc
Wasser1462: I originally added `CopyRawToCpu` because some of my debug/casting checks need to read tensor data on CPU, while under CUDA the tensor buffer may reside on GPU. However, for the normal inference path, it’s better to follow the existing sherpa-onnx pattern and let ONNX Runtime handle CPU/GPU transfers automatically.

I’ll simplify this part, keep any special handling only where it’s truly needed.

sherpa-onnx/csrc/offline-funasr-nano-model.cc
```diff
@@ -0,0 +1,1037 @@
...
+                   return static_cast<char>(std::tolower(c));
+                 });
+  return s;
+}
+
+static inline bool IsCudaProvider(const std::string &provider) {
+  auto p = ToLower(provider);
+  // Keep it conservative. We only enable IO binding policy below when we
+  // are on CUDA; other EPs keep the existing behavior.
+  return p == "cuda" || (p.size() > 4 && p.rfind("cuda", 0) == 0);
```
csukuangfj:
```suggestion
  return p == "cuda" || (p.size() > 4 && p.find("cuda") == 0);
```

It is more readable to use `find` instead of `rfind()` here.

sherpa-onnx/csrc/offline-funasr-nano-model.cc
```diff
@@ -0,0 +1,1037 @@
...
+// Convert tensor to float32, handling both float16 and float32 inputs.
+// NOTE: This helper assumes the input tensor is on CPU memory.
+// The caller must ensure the tensor is on CPU (e.g., via IO Binding).
+static Ort::Value CastToFloat32(Ort::Value in, OrtAllocator *alloc) {
+  if (!in.IsTensor()) return in;
+  auto info = in.GetTensorTypeAndShapeInfo();
+  auto shape = info.GetShape();
+  size_t n = NumelFromShape(shape);
+  if (n == 0) return in;
+  auto et = static_cast<ONNXTensorElementDataType>(info.GetElementType());
```
csukuangfj:
```suggestion
  auto et = info.GetElementType();
```

The return type is already `ONNXTensorElementDataType`, so there is no need to call `static_cast<>`.

sherpa-onnx/csrc/offline-funasr-nano-model.cc
csukuangfj: > I don’t need to free name

You are right. Thanks for the explanation.

sherpa-onnx/csrc/offline-funasr-nano-model.cc
```diff
@@ -0,0 +1,1037 @@
...
+        sess_opts_embedding_(GetSessionOptions(config)),
+        allocator_(),
+        cpu_mem_info_(Ort::MemoryInfo::CreateCpu(OrtDeviceAllocator,
+                                                 OrtMemTypeDefault)),
+        is_cpu_provider_(config.provider == "cpu" || config.provider.empty()) {
+    const auto &c = config_.funasr_nano;
+    auto buf_encoder = ReadFile(mgr, c.encoder_adaptor);
+    InitEncoderAdaptorFromMemory(buf_encoder.data(), buf_encoder.size());
+    auto buf_prefill = ReadFile(mgr, c.llm_prefill);
+    InitLLMPrefillFromMemory(buf_prefill.data(), buf_prefill.size());
```
csukuangfj:
```suggestion
{
    auto buf_prefill = ReadFile(mgr, c.llm_prefill);
    InitLLMPrefillFromMemory(buf_prefill.data(), buf_prefill.size());
 }
```
Use `{}` so that `buf_prefill` is freed right after the initialization. We can reduce the peak memory usage in this case.

sherpa-onnx/csrc/offline-funasr-nano-model.cc
```diff
@@ -0,0 +1,1037 @@
...
+    }
+
+    std::array<Ort::Value, 1> inputs = {std::move(features)};
+    auto outputs = encoder_sess_->Run(
+        {}, encoder_input_names_ptr_.data(), inputs.data(), inputs.size(),
+        encoder_output_names_ptr_.data(), encoder_output_names_ptr_.size());
+    return std::move(outputs[0]);
+  }
+
+  Ort::Value ForwardLLM(Ort::Value, Ort::Value) {
```
csukuangfj: Can you remove it if it is not supported?
And also please optimize the fun asr nano model config validation logic when use_kv_cache is always true.

sherpa-onnx/csrc/offline-funasr-nano-model.cc
```diff
@@ -0,0 +1,1037 @@
...
+      SHERPA_ONNX_READ_META_DATA(decode_hidden_size, "hidden_size");
+      hidden_size_ = decode_hidden_size;
+    }
+    int32_t decode_num_layers = 0;
+    SHERPA_ONNX_READ_META_DATA(decode_num_layers, "num_layers");
+    if (num_layers_ > 0 && decode_num_layers != num_layers_) {
+      SHERPA_ONNX_LOGE(
+          "Decode model num_layers (%d) != prefill num_layers (%d)",
+          decode_num_layers, num_layers_);
+    }
```
csukuangfj: Looks like num_layers_ is not used in the code. Can it be removed?

sherpa-onnx/csrc/offline-funasr-nano-model.cc
```diff
@@ -0,0 +1,1037 @@
...
+      }
+    }
+
+    // Build inputs: [inputs_embeds, attention_mask, past_key_0, past_value_0, ...]
+    // NOTE: We create non-owning Ort::Value views that reference existing buffers.
+    std::vector<Ort::Value> inputs;
+    inputs.reserve(2 + 2 * past_key_values.size());
+    inputs.push_back(std::move(inputs_embeds));
+    inputs.push_back(std::move(attention_mask));
+    for (const auto &kv : past_key_values) {
```
csukuangfj, line 593: Can you re-use
https://github.com/k2-fsa/sherpa-onnx/blob/1f5f9633cc71f0426cbcea591fba56ef8a646984/sherpa-onnx/csrc/onnx-utils.cc#L188

It can simplify your code.

You need to change
https://github.com/k2-fsa/sherpa-onnx/blob/1f5f9633cc71f0426cbcea591fba56ef8a646984/sherpa-onnx/csrc/onnx-utils.cc#L192-L193
to
```cpp
 auto memory_info = v->GetTensorMemoryInfo()
```

Wasser1462: Thanks a lot for the detailed review and guidance — I learned a lot from your comments!

I've updated the PR accordingly:
- Removed unused code/parameters (e.g., `num_layers_`)
- Simplified the config/validation logic since KV cache is always enabled
- Reused `View()` from `onnx-utils.cc`, and updated it to use `v->GetTensorMemoryInfo()` to preserve the original memory info as suggested

BTW, I re-exported and re-uploaded the ONNX models without external `.data` files. However, for the FP32 model, the single-file export fails because it exceeds the 2GB limit during ONNX export, so I kept the external `.data` file for FP32.

Could you please take another look when you have a chance? 🙏

sherpa-onnx/csrc/funasr-nano-tokenizer.h
```diff
@@ -0,0 +1,108 @@
...
+  int64_t pad_token_id_ = -1;
+  int64_t im_end_token_id_ = -1;
+
+  std::unordered_set<int32_t> special_ids_;
+
+  // Vocab: token <-> id
+  std::unordered_map<std::string, int32_t> token2id_;
+  std::vector<std::string> id2token_;
+
+  // merges ranks: "left\t right" -> rank
```
csukuangfj:
```suggestion
  // merges ranks: "left\tright" -> rank
```

There is no space after `\t` in the code.

sherpa-onnx/csrc/funasr-nano-tokenizer.cc
```diff
@@ -0,0 +1,1366 @@
...
+  std::string k = a;
+  k.push_back('\t');
+  k.append(b);
+  return k;
+}
+
+}  // namespace
+
+// Parse tokenizer.json added_tokens: extract objects with {id, content, ...}
+bool ParseAddedTokensFromTokenizerJson(
```
csukuangfj:
```suggestion
static bool ParseAddedTokensFromTokenizerJson(
```

or move it into the above anonymous namespace.

sherpa-onnx/csrc/funasr-nano-tokenizer.cc
```diff
@@ -0,0 +1,1366 @@
...
+        node = new_node;
+      } else {
+        node = it->second;
+      }
+    }
+    (*trie)[node].token_index = i;
+  }
+}
+
+void MergeVocabAndAddedTokens(
```
csukuangfj:
```suggestion
static void MergeVocabAndAddedTokens(
```

Please make functions used only in this file `static`.

csukuangfj:
i am still reading the code. Thank you for your contribution!

sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.cc
```diff
@@ -0,0 +1,492 @@
...
+std::unique_ptr<OfflineStream>
+OfflineRecognizerFunASRNanoImpl::CreateStream() const {
+  return std::make_unique<OfflineStream>(config_.feat_config);
+}
+
+// Initialize feature extraction configuration for FunASR-nano.
+// Sets normalization, window type, and disables edge snipping and dithering
+// to match the model's expected input format.
+void OfflineRecognizerFunASRNanoImpl::InitFeatConfig() {
+  config_.feat_config.normalize_samples = true;
```
csukuangfj:
```suggestion
  config_.feat_config.normalize_samples = false;
```

all funasr models, by default, use samples from in the range [-32768, 32767], so we need to set it to false.

See also
https://github.com/modelscope/FunASR/blob/main/funasr/frontends/wav_frontend.py#L96
```
        upsacle_samples: bool = True,
```

https://github.com/modelscope/FunASR/blob/main/funasr/frontends/wav_frontend.py#L130
```
            if self.upsacle_samples:
                waveform = waveform * (1 << 15)
```

Note: There is a typo in `upsacle_samples`.

Its default value is `True`.

sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.cc
```diff
@@ -0,0 +1,492 @@
...
+}
+
+// Generate text from encoder output using autoregressive decoding with KV cache.
+// Combines text embeddings (from prompts) and audio embeddings (from encoder)
+// to form the input sequence, then generates tokens autoregressively.
+OfflineRecognitionResult OfflineRecognizerFunASRNanoImpl::GenerateText(
+    Ort::Value encoder_out, const std::string &system_prompt,
+    const std::string &user_prompt) const {
+  OfflineRecognitionResult result;
+  if (!model_->HasEmbeddingModel()) {
```
csukuangfj: Can you check at the model config validation time that if the embedding model is not provided, it returns false to the user?

We can assume at this point the embedding model is always available.

sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.cc
```diff
@@ -0,0 +1,492 @@
...
+  int32_t hidden_size = static_cast<int32_t>(enc_shape[2]);
+  int32_t fbank_beg_idx = 0;
+  int32_t fake_token_len = 0;
+  std::vector<int64_t> source_ids = BuildSourceIds(
+      system_prompt, user_prompt, audio_token_len, fbank_beg_idx,
+      fake_token_len);
+  int32_t context_len = static_cast<int32_t>(source_ids.size());
+  const int32_t max_seq_len = 2048;
+  if (context_len > max_seq_len) {
+    source_ids.resize(max_seq_len);
```
csukuangfj, line 173: Will the recognition result be truncated as a result of this resize?
If so, could you print a log message noting that the result may be truncated due to excessive input length?

sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.cc
```diff
@@ -0,0 +1,492 @@
...
+  // Set attention mask for context tokens
+  for (int32_t i = 0; i < context_len; ++i) attention_mask[i] = 1;
+  int32_t valid_len = context_len;
+  std::vector<int64_t> generated_ids;
+  
+  generated_ids.reserve(funasr_config.max_new_tokens);
+  const int64_t eos_id = tokenizer_->GetEosTokenId();
+  const int64_t im_end_id = tokenizer_->GetImEndTokenId();
+  const int32_t max_new_tokens =
+      funasr_config.max_new_tokens > 0 ? funasr_config.max_new_tokens : 256;
```
csukuangfj: `funasr_config.max_new_tokens` is always `>=0` since you have validated it in the model config validation function.

sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.cc
```diff
@@ -0,0 +1,492 @@
...
+  // Autoregressive generation loop
+  for (int32_t step = 0; step < max_new_tokens; ++step) {
+    if (valid_len >= max_seq_len) break;
+    Ort::Value logits(nullptr);
+    
+    if (is_first_step) {
+      // First step: use prefill model with full context
+      std::array<int64_t, 3> embeds_shape{1, context_len, hidden_size};
+      Ort::Value inputs_embeds_tensor = Ort::Value::CreateTensor(
+          memory_info, inputs_embeds_fp16.data(),
```
csukuangfj, line 259: Could you explain how you obtain the fp16 model?

Do we need to handle fp16 inputs and outputs?

You might want to take a look at the following script:
https://github.com/k2-fsa/sherpa-onnx/blob/acd44bf9ab0ee83c2189c93b442038c0b4e66786/scripts/uvr_mdx/add_meta_data_and_quantize.py#L31

If you set keep_io_types=True, the model’s input and output will always remain float32.

csukuangfj:
The code looks great to me—thanks again!

Would it be possible to also open-source the model export code?

The repository below does not seem to include code for exporting LLMs with KV cache support or for exporting fp16 models:
https://github.com/Wasser1462/FunASR-nano-onnx

sherpa-onnx/csrc/funasr-nano-tokenizer.h
```diff
@@ -0,0 +1,108 @@
+// sherpa-onnx/csrc/funasr-nano-tokenizer.h
+//
+// A self-contained Qwen3 ByteLevel-BPE tokenizer implementation.
+// - No dependency on tokenizers-cpp / HF tokenizers
+// - Loads vocab.json + merges.txt + tokenizer.json(added_tokens)
+// - Supports AddedTokens via Trie longest-match
+// - ByteLevel bytes_to_unicode encode/decode
+// Author: zengyongwang
```
csukuangfj: Was this file authored by you? Additionally, could you add copyright information to the other files included in this pull request?

sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.cc
Wasser1462: Thanks, `offline-funasr-nano-model-config.cc::Validate()` already checks that `--funasr-nano-embedding` is provided and the file exists (returns `false` otherwise),
so we can assume the embedding model is always available by the time we enter `OfflineRecognizerFunASRNanoImpl::GenerateText()`.

sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.cc
Wasser1462: Yes, the result can be truncated when `context_len > max_seq_len` because we resize
the input to `max_seq_len`. I added a log message to warn users that the input is
being truncated and the recognition result may be incomplete.

sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.cc
Wasser1462: Thanks for the catch! `max_new_tokens` is already validated to be `>= 0` in config validation, so I simplified the code to:

```cpp
const int32_t max_new_tokens = funasr_config.max_new_tokens;
```

sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.cc
Wasser1462: Thanks for the suggestion!

At the moment I’ll **keep fp16-IO LLM models supported on CPU only**.  
For **CUDA**, fp16 inputs/outputs are still unstable on my side (e.g. logits can become NaN), so I’m adding a validation check to **reject fp16-IO LLM models when using the CUDA EP** with a clear error message.

Users can continue to use the **fp32/int8** models on CUDA, and I’ll revisit fp16 support later by re-exporting the fp16 model with **float32 IO** (e.g. `keep_io_types=True`) and adapting the runtime accordingly.

Wasser1462:
Thanks for the kind words!

1) About the export code: the current exporter is still in a rough state. I initially exported with opset 18, but to stay compatible with the ONNX Runtime version used in sherpa-onnx, I had to add a set of post-export fixes (opset/IR adjustments, graph patches, etc.), so it’s not clean enough to upstream yet. I’m cleaning it up now and will open-source it and submit a PR to sherpa-onnx once it’s polished and reproducible.

2) About the tokenizer/other files: yes, they are authored by me. I will add proper copyright headers to this file and the other new files in this PR.

python-api-examples/offline-funasr-nano-decode-files.py
```diff
@@ -0,0 +1,217 @@
...
+        nargs="+",
+        help="The input sound file(s) to decode. "
+        "Each file must be of single channel, 16-bit PCM encoded wav file. "
+        "Its sample rate can be arbitrary and does not need to be 16kHz.",
+    )
+
+    return parser.parse_args()
+
+
+def create_recognizer(args) -> sherpa_onnx.OfflineRecognizer:
```
csukuangfj, line 151: Can you follow
https://github.com/k2-fsa/sherpa-onnx/blob/1f124600f506b45426c3b05f7fcc588ed373d03e/sherpa-onnx/python/sherpa_onnx/offline_recognizer.py#L210
to add
```
def from_funasr_nano
```
?

Also, could you upload the files you  have changed inside
`sherpa-onnx/python/csrc`?

Wasser1462: The PR has been updated accordingly.

csukuangfj:
> Thanks for the kind words!
> 
> 1) About the export code: the current exporter is still in a rough state. I initially exported with opset 18, but to stay compatible with the ONNX Runtime version used in sherpa-onnx, I had to add a set of post-export fixes (opset/IR adjustments, graph patches, etc.), so it’s not clean enough to upstream yet. I’m cleaning it up now and will open-source it and submit a PR to sherpa-onnx once it’s polished and reproducible.
> 
> 2) About the tokenizer/other files: yes, they are authored by me. I will add proper copyright headers to this file and the other new files in this PR.
> 

Hope you can release the export script soon. I want to try it with Qualcomm NPU.

Wasser1462:
The export script has been pushed here for reference:
https://github.com/Wasser1462/FunASR-nano-onnx/tree/main/export_kv_cache_model

Please note it’s not fully cleaned up yet — I originally exported with opset 18, and later had to add a number of compatibility fixes (opset/IR downgrades, Reduce* axes fixes, etc.) to match the ONNX Runtime version used by sherpa-onnx. I’m currently tied up with year-end work tasks, so I haven’t had time to refactor and document it properly, but it should still be useful as a starting point for trying Qualcomm NPU.

sherpa-onnx/csrc/offline-funasr-nano-model-config.h
```diff
@@ -0,0 +1,59 @@
...
+  std::string user_prompt = "语音转写：";
+
+  // Maximum number of new tokens to generate
+  int32_t max_new_tokens = 512;
+
+  // Sampling temperature
+  float temperature = 0.3f;
+
+  // Top-p (nucleus) sampling threshold
+  float top_p = 0.8f;
```
csukuangfj, line 43: Looks like `temperature` and `top_p` are never used.

Do you plan to use them in a separate pull request?

csukuangfj:
Thanks again for your great contribution!

I will merge it first. Please address any issues, if any, in a separate pull request.

Remember2015:
...
`modelscope download --model zengshuishui/FunASR-nano-onnx-V2 --output_dir ./models`
...
arg `--output_dir` maybe `--local_dir` in newer version of modelscope

Wasser1462:
Heads up: this PR was built on the old main export code (the prefill + decode inference flow). Since main has been refactored and the export pipeline was restructured, the original working version is preserved here:

https://github.com/Wasser1462/FunASR-nano-onnx/tree/backup/main-2026-01-06

If you need the legacy prefill + decode export/inference code, please use that backup branch. I’ll open a new PR soon for the refactored version.

PR #3115, opened 2026-02-01
add TTS generation with GenerationConfig params C API

seven1240:
## Summary by CodeRabbit

* **New Features**
  * Pocket TTS: offline English TTS with expanded pocket-model configuration options.
  * Generation config API: create/use generation configs (silence/speed/sid/steps/extra), supply reference audio/text, and new config-driven generation entry points with progress callbacks accepting a user argument.
  * OHOS and non-enabled build behavior: generation APIs present with clear runtime handling when TTS is not enabled.

* **Documentation**
  * Added a C example demonstrating Pocket TTS usage, progress reporting, WAV output, and integrated build entry.

### DIFF
--- a/c-api-examples/pocket-tts-en-c-api.c
+++ b/c-api-examples/pocket-tts-en-c-api.c
@@ -1,0 +1,113 @@
+// c-api-examples/pocket-tts-en-c-api.c
+//
+// Copyright (c)  2026  Xiaoyingtao Corporation
+
+// This file shows how to use sherpa-onnx C API
+// for English TTS with Pocket TTS.
+//
+// clang-format off
+/*
+Usage
+
+
+wget https://github.com/k2-fsa/sherpa-onnx/releases/download/tts-models/sherpa-onnx-pocket-tts-int8-2026-01-26.tar.bz2
+tar xf sherpa-onnx-pocket-tts-int8-2026-01-26.tar.bz2
+rm sherpa-onnx-pocket-tts-int8-2026-01-26.tar.bz2
+
+./pocket-tts-en-c-api
+
+ */
+// clang-format on
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "sherpa-onnx/c-api/c-api.h"
+
+static int32_t ProgressCallback(const float *samples, int32_t num_samples,
+                                float progress, void *arg) {
+  fprintf(stderr, "Progress: %.3f%%\n", progress * 100);
+  // return 1 to continue generating
+  // return 0 to stop generating
+  return 1;
+}
+
+int32_t main(int32_t argc, char *argv[]) {
+  SherpaOnnxOfflineTtsConfig config;
+  memset(&config, 0, sizeof(config));
+  config.model.pocket.lm_flow =
+      "./sherpa-onnx-pocket-tts-int8-2026-01-26/lm_flow.int8.onnx";
+  config.model.pocket.lm_main =
+      "./sherpa-onnx-pocket-tts-int8-2026-01-26/lm_main.int8.onnx";
+  config.model.pocket.encoder =
+      "./sherpa-onnx-pocket-tts-int8-2026-01-26/encoder.onnx";
+  config.model.pocket.decoder =
+      "./sherpa-onnx-pocket-tts-int8-2026-01-26/decoder.int8.onnx";
+  config.model.pocket.text_conditioner =
+      "./sherpa-onnx-pocket-tts-int8-2026-01-26/text_conditioner.onnx";
+  config.model.pocket.vocab_json =
+      "./sherpa-onnx-pocket-tts-int8-2026-01-26/vocab.json";
+  config.model.pocket.token_scores_json =
+      "./sherpa-onnx-pocket-tts-int8-2026-01-26/token_scores.json";
+
+  config.model.num_threads = 2;
+
+  // If you don't want to see debug messages, please set it to 0
+  config.model.debug = 1;
+
+  const char *filename = "./generated-pocket-en.wav";
+  const char *text =
+      "Today as always, men fall into two groups: slaves and free men. Whoever "
+      "does not have two-thirds of his day for himself, is a slave, whatever "
+      "he may be: a statesman, a businessman, an official, or a scholar. "
+      "Friends fell out often because life was changing so fast. The easiest "
+      "thing in the world was to lose touch with someone.";
+
+  const SherpaOnnxOfflineTts *tts = SherpaOnnxCreateOfflineTts(&config);
+  if (!tts) {
+    fprintf(stderr, "Error create Offline TTS\n");
+    return -1;
+  }
+  int32_t sid = 0;
+  float speed = 1.0;  // larger -> faster in speech speed
+  GenerationConfig cfg = {0};
+  const char *reference_audio_file =
+      "./sherpa-onnx-pocket-tts-int8-2026-01-26/test_wavs/bria.wav";
+  const SherpaOnnxWave *wave = NULL;
+  wave = SherpaOnnxReadWave(reference_audio_file);
+  if (!wave) {
+    fprintf(stderr, "Failed to read %s\n", reference_audio_file);
+    SherpaOnnxDestroyOfflineTts(tts);
+    return -1;
+  }
+  cfg.reference_audio = wave->samples;
+  cfg.reference_audio_len = wave->num_samples;
+  cfg.reference_sample_rate = wave->sample_rate;
+  cfg.extra = "{\"max_reference_audio_len\": 10.0}";
+
+#if 0
+  // If you don't want to use a callback, then please enable this branch
+  const SherpaOnnxGeneratedAudio *audio =
+      SherpaOnnxOfflineTtsGenerateWithConfig(tts, text, &cfg, NULL, NULL);
+#else
+  const SherpaOnnxGeneratedAudio *audio =
+      SherpaOnnxOfflineTtsGenerateWithConfig(tts, text, &cfg, ProgressCallback,
+                                             NULL);
+#endif
+
+  if (wave) SherpaOnnxFreeWave(wave);
+
+  fprintf(stderr, "Input text is: %s\n", text);
+  fprintf(stderr, "Speaker ID is: %d\n", sid);
+
+  if (audio) {
+    SherpaOnnxWriteWave(audio->samples, audio->n, audio->sample_rate, filename);
+    fprintf(stderr, "Saved to: %s\n", filename);
+    SherpaOnnxDestroyOfflineTtsGeneratedAudio(audio);
+  }
+
+  SherpaOnnxDestroyOfflineTts(tts);
+
+  return 0;
+}
--- a/sherpa-onnx/c-api/c-api.cc
+++ b/sherpa-onnx/c-api/c-api.cc
@@ -12,6 +12,8 @@
 #include <utility>
 #include <vector>
 
+#include "nlohmann/json.hpp"
+
 #if __OHOS__
 #include "rawfile/raw_file_manager.h"
 #endif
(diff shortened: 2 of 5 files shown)


sherpa-onnx/c-api/c-api.h
```diff
@@ -1200,6 +1211,71 @@ SherpaOnnxOfflineTtsGenerateWithZipvoice(const SherpaOnnxOfflineTts *tts,
                                          int32_t n_prompt, int32_t prompt_sr,
                                          float speed, int32_t num_steps);
 
+SHERPA_ONNX_API typedef struct SherpaOnnxTtsGenerationConfig
```
csukuangfj: Please create a struct in c-api.h for `GenerationConfig`
to include the following fields

https://github.com/k2-fsa/sherpa-onnx/blob/72f77c62be77ec85f8ec64b9b43fc58d0d354266/sherpa-onnx/csrc/offline-tts.h#L68-L76


Note: Use
```
const float* reference_audio;
int32_t reference_audio_len;
```
to represent 
```cpp
std::vector<float> reference_audio; 
```
and use
```
const char*  reference_text;
```

---

Also, please don't wrap
https://github.com/k2-fsa/sherpa-onnx/blob/72f77c62be77ec85f8ec64b9b43fc58d0d354266/sherpa-onnx/csrc/offline-tts.h#L83-L90
to c-api.

Please delete
- SherpaOnnxTtsGenerationConfigCreate
- SherpaOnnxTtsGenerationConfigDestroy
- SherpaOnnxTtsGenerationConfigGetInt
- SherpaOnnxTtsGenerationConfigGetFloat
- SherpaOnnxTtsGenerationConfigGetStr
- SherpaOnnxTtsGenerationConfigSetStr
- SherpaOnnxTtsGenerationConfigSetInt
- SherpaOnnxTtsGenerationConfigSetFloat
- SherpaOnnxTtsGenerationConfigToString

sherpa-onnx/c-api/c-api.h
```diff
@@ -1200,6 +1211,71 @@ SherpaOnnxOfflineTtsGenerateWithZipvoice(const SherpaOnnxOfflineTts *tts,
...
+
+// The user has to free the returned pointer if it's not NULL
+SHERPA_ONNX_API char *SherpaOnnxTtsGenerationConfigToString(
+    SherpaOnnxTtsGenerationConfig *config);
+
+// Generate audio from the given text and speaker id (sid).
+// The user has to use SherpaOnnxDestroyOfflineTtsGeneratedAudio() to free the
+// returned pointer to avoid memory leak.
+SHERPA_ONNX_API const SherpaOnnxGeneratedAudio *
+SherpaOnnxOfflineTtsGenerateWithConfig(const SherpaOnnxOfflineTts *tts,
```
csukuangfj: Please delete
- SherpaOnnxOfflineTtsGenerateWithConfig
- SherpaOnnxOfflineTtsGenerateWithConfigAndCallback
- SherpaOnnxOfflineTtsGenerateWithConfigAndProgressCallback


We need to keep only 
`SherpaOnnxOfflineTtsGenerateWithConfigAndProgressCallbackWithArg`

Please rename `SherpaOnnxOfflineTtsGenerateWithConfigAndProgressCallbackWithArg`
to `SherpaOnnxOfflineTtsGenerateWithConfig`.

sherpa-onnx/c-api/c-api.h
```diff
@@ -1200,6 +1211,71 @@ SherpaOnnxOfflineTtsGenerateWithZipvoice(const SherpaOnnxOfflineTts *tts,
...
+const SherpaOnnxGeneratedAudio *
+SherpaOnnxOfflineTtsGenerateWithConfigAndProgressCallbackWithArg(
+    const SherpaOnnxOfflineTts *tts, const char *text,
+    SherpaOnnxTtsGenerationConfig *config,
+    SherpaOnnxGeneratedAudioProgressCallbackWithArg callback, void *arg);
+
+// Same as SherpaOnnxGeneratedAudioCallback but you can pass an additional
+// `void* arg` to the callback.
+SHERPA_ONNX_API const SherpaOnnxGeneratedAudio *
+SherpaOnnxOfflineTtsGenerateWithConfigAndCallbackWithArg(
```
csukuangfj: 1. Please rename it to
```
SherpaOnnxOfflineTtsGenerateWithConfig
```

2. Please change the arguments to
```cpp
const SherpaOnnxOfflineTts *tts, const char *text,
    SherpaOnnxTtsGenerationConfig *config,
    const char* extra_opt,
    SherpaOnnxGeneratedAudioCallbackWithArg callback, void *arg)
```

where extra_opt is a json string. You can parse it to construct
https://github.com/k2-fsa/sherpa-onnx/blob/72f77c62be77ec85f8ec64b9b43fc58d0d354266/sherpa-onnx/csrc/offline-tts.h#L81

Note that `extra_opt`, `callback`, and `arg` can be nullptr.

sherpa-onnx/c-api/c-api.h
seven1240: ok, c++ use an internal vector to hold the reference audio which is hard to manage the life cycle in C. without those function the user would need to use SherpaOnnxReadWave to read the wav into the reference audio buffer and remember to destroy it after use, is that what you mean? thanks.

seven1240: and, I just realized that, even you can pass a raw pointer from c to c++, the c++ class internally need a vector, so still need one more memory copy. My whole idea was to avoid unnecessary memory copy as long as possible. I'm not a c++ guru, is there any suggestion for this?

seven1240: I pushed a commit which kind of works, but the GenerationConfig still need some APIs to

- Set extra params
- Manage an internal state to hold param values
- Init and Cleanup functions

It's incomplete, let me know if it makes sense I'll complete this.

Thanks.

seven1240: ok, I saw you said we could use a JSON extra opt, I'll try.

sherpa-onnx/c-api/c-api.h
seven1240: since it's a JSON, can we just add to the config struct instead of add a new arg?

csukuangfj: > since it's a JSON, can we just add to the config struct instead of add a new arg?

Yes, that is also a good idea. In that case, please add
```
const char* extra;
```
to the struct and document that it is a json string.

c-api-examples/pocket-tts-en-c-api.c
```diff
@@ -0,0 +1,121 @@
...
+    cfg.reference_audio = wave->samples;
+    cfg.reference_audio_len = wave->num_samples;
+    cfg.reference_sample_rate = wave->sample_rate;
+  } else {
+    SherpaOnnxTtsGenerationConfigSetReferenceAudioFile(&cfg,
+                                                       reference_audio_file);
+    SherpaOnnxTtsGenerationConfigSetInt(&cfg, "sid", sid);
+    SherpaOnnxTtsGenerationConfigSetFloat(&cfg, "speed", speed);
+    SherpaOnnxTtsGenerationConfigSetFloat(&cfg, "max_reference_audio_len",
+                                          10.0);
```
csukuangfj: Can you delete this else branch?

c-api-examples/pocket-tts-en-c-api.c
```diff
@@ -0,0 +1,121 @@
...
+  // If you don't want to use a callback, then please enable this branch
+  const SherpaOnnxGeneratedAudio *audio =
+      SherpaOnnxOfflineTtsGenerateWithConfig(tts, text, &cfg, NULL, NULL);
+#else
+  const SherpaOnnxGeneratedAudio *audio =
+      SherpaOnnxOfflineTtsGenerateWithConfig(tts, text, &cfg, ProgressCallback,
+                                             NULL);
+#endif
+
+  SherpaOnnxTtsGenerationConfigCleanup(&cfg);
```
csukuangfj: Please delete it. Normally, we don't need to clean up a struct config. We assume the config contains pointers to memory allocated by the user.

sherpa-onnx/c-api/c-api.h
```diff
@@ -1200,6 +1211,38 @@ SherpaOnnxOfflineTtsGenerateWithZipvoice(const SherpaOnnxOfflineTts *tts,
...
+    GenerationConfig *config, const char *audio_file);
+SHERPA_ONNX_API void SherpaOnnxTtsGenerationConfigCleanup(
+    GenerationConfig *config);
+
+SHERPA_ONNX_API void SherpaOnnxTtsGenerationConfigSetStr(
+    GenerationConfig *config, const char *k, const char *v);
+SHERPA_ONNX_API void SherpaOnnxTtsGenerationConfigSetInt(
+    GenerationConfig *config, const char *k, int32_t v);
+SHERPA_ONNX_API void SherpaOnnxTtsGenerationConfigSetFloat(
+    GenerationConfig *config, const char *k, float v);
```
csukuangfj: Can you delete these functions?
```suggestion
```

sherpa-onnx/c-api/c-api.cc
```diff
@@ -1444,6 +1504,82 @@ const SherpaOnnxGeneratedAudio *SherpaOnnxOfflineTtsGenerateWithZipvoice(
...
+  auto wrapper = [callback, arg](const float *samples, int32_t n,
+                                 float progress) {
+    if (!callback) return 0;
+    return callback(samples, n, progress, arg);
+  };
+
+  return SherpaOnnxOfflineTtsGenerateInternal(tts, text, config, wrapper);
+}
+
+void SherpaOnnxTtsGenerationConfigSetReferenceAudioFile(
```
csukuangfj: Please delete this function.

We just set a pointer to the audio in the struct.
The struct does not own the memory pointed to by the pointer.

sherpa-onnx/c-api/c-api.cc
```diff
@@ -1444,6 +1504,82 @@ const SherpaOnnxGeneratedAudio *SherpaOnnxOfflineTtsGenerateWithZipvoice(
...
+}
+
+void SherpaOnnxTtsGenerationConfigCleanup(GenerationConfig *config) {
+  if (!config || !config->pvt) return;
+  sherpa_onnx::GenerationConfig *cfg =
+      (sherpa_onnx::GenerationConfig *)config->pvt;
+  delete cfg;
+}
+
+void SherpaOnnxTtsGenerationConfigSetStr(GenerationConfig *config,
```
csukuangfj: Please remove it. We don't need such functions.

seven1240:
Thanks for the comments, I'll try with a new patch, in the mean time, do we already have an internal JSON parser?

csukuangfj:
> internal JSON parser

Yes! Please see how we read the two json files in the struct.

sherpa-onnx/c-api/c-api.cc
```diff
@@ -1444,6 +1510,18 @@ const SherpaOnnxGeneratedAudio *SherpaOnnxOfflineTtsGenerateWithZipvoice(
...
+const SherpaOnnxGeneratedAudio *SherpaOnnxOfflineTtsGenerateWithConfig(
+    const SherpaOnnxOfflineTts *tts, const char *text, GenerationConfig *config,
+    SherpaOnnxGeneratedAudioProgressCallbackWithArg callback, void *arg) {
+  auto wrapper = [callback, arg](const float *samples, int32_t n,
+                                 float progress) {
+    if (!callback) return 1;
+    return callback(samples, n, progress, arg);
+  };
+
+  return SherpaOnnxOfflineTtsGenerateInternal(tts, text, config, wrapper);
```
csukuangfj:
```suggestion
  return SherpaOnnxOfflineTtsGenerateInternal(tts, text, config, std::move(wrapper));
```

sherpa-onnx/c-api/c-api.cc
```diff
@@ -1359,6 +1377,54 @@ static const SherpaOnnxGeneratedAudio *SherpaOnnxOfflineTtsGenerateInternal(
...
+        config->reference_audio,
+        config->reference_audio + config->reference_audio_len);
+  }
+  if (config->silence_scale > 0) cfg.silence_scale = config->silence_scale;
+  if (config->speed > 0) cfg.speed = config->speed;
+  cfg.sid = config->sid;
+  if (config->reference_sample_rate > 0)
+    cfg.reference_sample_rate = config->reference_sample_rate;
+  if (config->reference_text)
+    cfg.reference_text = std::string(config->reference_text);
```
csukuangfj:
```suggestion
    cfg.reference_text = config->reference_text;
```

sherpa-onnx/c-api/c-api.cc
```diff
@@ -1359,6 +1377,54 @@ static const SherpaOnnxGeneratedAudio *SherpaOnnxOfflineTtsGenerateInternal(
...
+  if (config->num_steps > 0) cfg.num_steps = config->num_steps;
+  if (config->extra) {
+    if (config->extra) {
```
csukuangfj: Only one if is enough.

csukuangfj (commented):
Thanks! Left some minor comments. Otherwise, it looks great to me.

sherpa-onnx/c-api/c-api.cc
seven1240: fixed as suggested, but `std::move` are not used by other wrappers in the same file.

sherpa-onnx/c-api/c-api.cc
seven1240: sorry, was a mistake :)

PR #3136, opened 2026-02-04
Do not merge, untested: Wrap SherpaOnnxWriteWaveToBuffer in the GO API

WilliamTambellini:
Wrap SherpaOnnxWriteWaveToBuffer into the GO API.

Todo:
- unittest

## Summary by CodeRabbit

* **New Features**
  * Retrieve generated audio as an in-memory WAV byte buffer instead of writing to disk.
  * Mirrors existing save behavior: returns nil when no samples are present; otherwise returns the complete WAV data ready for immediate use.
  * Simplifies workflows that consume audio directly (playback, upload, or further processing) without filesystem I/O.

### DIFF
--- a/scripts/go/sherpa_onnx.go
+++ b/scripts/go/sherpa_onnx.go
@@ -1174,6 +1174,17 @@ func (audio *GeneratedAudio) Save(filename string) bool {
 	return ok == 1
 }
 
+func (audio *GeneratedAudio) ToBuffer() []byte {
+	// Similar to Save(): it writes the wave to an allocated buffer; 
+	// Uses the C API: SHERPA_ONNX_API void SherpaOnnxWriteWaveToBuffer(const float *samples, int32_t n, int32_t sample_rate, char *buffer);
+	n := len(audio.Samples)
+	if n == 0 { return nil }
+	fs := C.SherpaOnnxWaveFileSize(C.int(n)) // SHERPA_ONNX_API int64_t SherpaOnnxWaveFileSize(int32_t n_samples);
+	buf := make([]byte, fs)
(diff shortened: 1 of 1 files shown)


WilliamTambellini:
@csukuangfj  could you please advice where to add a unittest?

csukuangfj:
> @csukuangfj could you please advice where to add a unittest?

If you want to add test for it, you can change

https://github.com/k2-fsa/sherpa-onnx/blob/ee398fa98fde44c2a4cccdea8153cdfb72074a42/go-api-examples/non-streaming-tts/main.go#L117

Instead of saving the audio to a file directly with a single call `save`, you can first write it to a buffer, and then call some GO API to save the buffer to a binary file.

To run it locally, please run any bash script in https://github.com/k2-fsa/sherpa-onnx/tree/master/scripts/go/_internal/non-streaming-tts


(This is for testing locally. After checking that it works, you can delete it. We don't have unit tests for Go now.)

scripts/go/sherpa_onnx.go
```diff
@@ -1174,6 +1174,18 @@ func (audio *GeneratedAudio) Save(filename string) bool {
...
 }
 
+func (audio *GeneratedAudio) ToBuffer() []byte {
+	// Similar to Save(): it writes the wave to an allocated buffer; 
+	// Uses the C API: SHERPA_ONNX_API void SherpaOnnxWriteWaveToBuffer(const float *samples, int32_t n, int32_t sample_rate, char *buffer);
+	n := len(audio.Samples)
+	if n == 0 { return nil }
+	fs := C.SherpaOnnxWaveFileSize(C.int(n)) // SHERPA_ONNX_API int64_t SherpaOnnxWaveFileSize(int32_t n_samples);
+	buf := make([]byte, fs) 
+	if n == 0 { return buf }
```
csukuangfj: Line 1181 and 1184 are identical.

Please remove
```suggestion
```

WilliamTambellini: oops tks, removing...

PR #3193, opened 2026-02-16
fixed 180m-flash export and added 1b-flash export

jackalxyz:
Fixed the error when exporting 180m-flash and added export for 1b-flash

## Summary by CodeRabbit

* **New Features**
  * Added ONNX export and test tooling for canary models (including 1B and 180M variants), plus runnable scripts to export, quantize, and validate FP32/INT8 encoder+decoder models and produce transcriptions.

* **Chores**
  * Included encoder ONNXs in validation, attached metadata, and adjusted IR compatibility and attention-mask handling for ONNXRuntime.

* **Bug Fixes**
  * Fixed attention-mask compatibility; corrected an install command typo that could break the run script.

Issue #3190, opened 2026-02-15
Add pretrained model support for canary-1b-flash ?


jackalxyz:
This model ranks really high for multilingual and fast. Would be really nice to have this.

https://huggingface.co/nvidia/canary-1b-flash

csukuangfj:
please follow https://github.com/k2-fsa/sherpa-onnx/blob/master/scripts/nemo/canary/export_onnx_180m_flash.py to export it by yourself.

jackalxyz:
Thanks! 🙏

jackalxyz:
The export script doesn't work for canary-1b-flash, I've did some fixes and the exported onnx also cannot be inferenced

jackalxyz:
@csukuangfj unfortunately, the script isn't really working, I've tried even the older versions of pytorch but no luck. It seems canary-1b-flash is different from canary-180m-flash model? would be really grateful if you can look into this. I'm not familiar with inferencing code.

jackalxyz:
WARNING:root:Please consider to run pre-processing before quantization. Refer to example: https://github.com/microsoft/onnxruntime-inference-examples/blob/main/quantization/image_classification/cpu/ReadMe.md 
old 7
new 9
old 7
new 9
-rw-rw-r-- 1 jackal jackal  76M Feb 16 22:25 decoder.int8.onnx
-rw-rw-r-- 1 jackal jackal 302M Feb 16 22:25 decoder.onnx
-rw-rw-r-- 1 jackal jackal 818M Feb 16 22:25 encoder.int8.onnx
-rw-rw-r-- 1 jackal jackal 3.1G Feb 16 22:25 encoder.onnx
+ ls -lh decoder.int8.onnx decoder.onnx encoder.int8.onnx encoder.onnx
-rw-rw-r-- 1 jackal jackal  76M Feb 16 22:25 decoder.int8.onnx
-rw-rw-r-- 1 jackal jackal 302M Feb 16 22:25 decoder.onnx
-rw-rw-r-- 1 jackal jackal 818M Feb 16 22:25 encoder.int8.onnx
-rw-rw-r-- 1 jackal jackal 3.1G Feb 16 22:25 encoder.onnx
+ log -----fp32------
+ local fname=run_1b_flash.sh
++ date '+%Y-%m-%d %H:%M:%S'
+ echo -e '2026-02-16 22:25:49 (run_1b_flash.sh:30:main) -----fp32------'
[...]

PR #3268, opened 2026-03-07
Support dynamic decoder layers in canary model runtime

mm65x:
Enables canary-1b-v2 which uses 10 decoder memory states (8 transformer layers + initial + final layer norm).

Related to #3190 and #3193.

Problem: `GetInitialDecoderStates()` in the canary model runtime hardcodes 6 decoder layers and 1024 hidden size. This works for canary-180m-flash (6 layers) but fails for larger canary models e.g. canary-1b-v2 which has 10 decoder memory states (8 transformer layers + initial + final layer norm).

This PR makes it read `num_decoder_layers` and `decoder_hidden_size` from the ONNX encoder metadata instead of hardcoding. Defaults to 6/1024 for backward compatibility with existing 180m models that don't have these metadata fields.

The PR introduces changes to the following files:
 - `sherpa-onnx/csrc/offline-canary-model-meta-data.h`: add `num_decoder_layers`, `decoder_hidden_size` fields with defaults
 - `sherpa-onnx/csrc/offline-canary-model.cc`: read from metadata in `InitEncoder()`, use in `GetInitialDecoderStates()`

## Summary by CodeRabbit

* **Refactor**
  * Model initialization now reads decoder architecture (layer count and hidden size) from model metadata instead of fixed values.
  * Metadata extended to include decoder layer count and hidden size, enabling support for more model variants.
  * Backward compatibility preserved via automatic default fallbacks when metadata is missing or invalid.

### DIFF
--- a/sherpa-onnx/csrc/offline-canary-model.cc
+++ b/sherpa-onnx/csrc/offline-canary-model.cc
@@ -181,34 +181,10 @@ class OfflineCanaryModel::Impl {
     SHERPA_ONNX_READ_META_DATA(meta_.subsampling_factor, "subsampling_factor");
     SHERPA_ONNX_READ_META_DATA(meta_.feat_dim, "feat_dim");
 
-    // Read decoder architecture metadata (with defaults for backward compat)
-    {
-      Ort::AllocatorWithDefaultOptions alloc;
-      try {
-        auto num_layers_str = meta_data.LookupCustomMetadataMapAllocated(
-            "num_decoder_layers", alloc);
-        if (num_layers_str) {
-          int32_t num_layers = std::stoi(num_layers_str.get());
-          if (num_layers > 0) {
-            meta_.num_decoder_layers = num_layers;
-          }
-        }
-      } catch (const std::exception &) {
-        // Use default (6) if not present or on parsing error
-      }
-      try {
-        auto hidden_size_str = meta_data.LookupCustomMetadataMapAllocated(
-            "decoder_hidden_size", alloc);
-        if (hidden_size_str) {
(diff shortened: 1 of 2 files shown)


sherpa-onnx/csrc/offline-canary-model.cc
```diff
@@ -178,6 +180,35 @@ class OfflineCanaryModel::Impl {
                                                "normalize_type");
     SHERPA_ONNX_READ_META_DATA(meta_.subsampling_factor, "subsampling_factor");
     SHERPA_ONNX_READ_META_DATA(meta_.feat_dim, "feat_dim");
+
+    // Read decoder architecture metadata (with defaults for backward compat)
```
csukuangfj: Please use
https://github.com/k2-fsa/sherpa-onnx/blob/33554f742c1a02f2f8e3291b42ca84bb90b91d9a/sherpa-onnx/csrc/macros.h#L72

See its usage at
https://github.com/k2-fsa/sherpa-onnx/blob/33554f742c1a02f2f8e3291b42ca84bb90b91d9a/sherpa-onnx/csrc/vocos-vocoder.cc#L162

mm65x: thanks, updated to use the existing macro

csukuangfj (approved):

CHICMELODIE:
@mm65x Thank you so much for your effort supporting canary-1b-v2. Do you know by any chance if the model will be available for download? I tried to generate ONNX models for canary-1b-flash or canary-1b-v2, but I don’t have enough skills and RAM to make it happen.

PR #3274, opened 2026-03-09
Results goes to stdout

phillc:
I'm doubting this commit makes it in as-is, but wanted to suggest or start a conversation.

I noticed that the result of some of these commands goes to stderr with everything else.

More specifically in my case, sherpa-onnx-offline outputs json intermingled with other debugging information. This makes it difficult to pipe it straight to another command.

This change makes it so it goes to stdout, and I changed a few others while I was at it. I tried to keep newlines and everything the same to make it so that you could just redirect stdout to stderr and get the same result as before.

This is still not perfect. In the sherpa-onnx-offline example, we are in a loop. I assume this is to give result of each file input. This change will make it print JSON, but not one big valid JSON (in an array). I wanted to keep my change backwards compatible (assuming stdout redirect), but there could be a bigger change somewhere to standardize the output reporting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Output stream handling updated across multiple command-line tools: keyword spotting, audio tagging, language identification, punctuation, and speech recognition. Results now appear on standard output instead of standard error, while labels and metadata remain on standard error.

### DIFF
--- a/sherpa-onnx/csrc/sherpa-onnx-keyword-spotter.cc
+++ b/sherpa-onnx/csrc/sherpa-onnx-keyword-spotter.cc
@@ -103,8 +103,9 @@ for a list of pre-trained models to download.
       if (!r.keyword.empty()) {
         keyword_spotter.Reset(s.get());
 
-        fprintf(stderr, "%s\n%s\n\n", wav_filename.c_str(),
-                r.AsJsonString().c_str());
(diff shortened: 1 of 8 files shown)


csukuangfj:

> I noticed that the result of some of these commands goes to stderr with everything else.

The output is intended for debugging purposes.

We recommend using one of our 12 supported programming languages for further development.

csukuangfj (approved):

PR #3309, opened 2026-03-12
feat: add SetOption/GetOption CXX wrapper

ZhaoChaoqun:
## Summary

Ref #3101 (Part 1c) — depends on #3308

Add `SetOption`/`GetOption` methods to the C++ wrapper (`cxx-api`) for both `OnlineStream` and `OfflineStream`.

## Files Changed (2 files, +22 lines)

| File | Change |
|------|--------|
| `cxx-api.h` | Add method declarations |
| `cxx-api.cc` | Implementations calling C API |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **New Features**
  * Added per-stream configuration options: set and retrieve custom options for both online and offline streams.
  * Added final chunk signaling for online streams: enables explicit notification when processing the final audio chunk in streaming scenarios.
  * Extended API support across C, C++, and Python interfaces for unified stream control and option management.

Issue #3101, opened 2026-01-28
[Help wanted] Add OnlineStream/OfflineStream.SetOption()


csukuangfj:
In the current design, a recognizer’s configuration becomes immutable once it is created. This makes it impossible to modify settings such as the prompt used in FunASR Nano or the task specified for Whisper.

To address this limitation, we propose adding several new methods to OnlineStream and OfflineStream:

```cpp
void SetOption(const std::string& key, const std::string& value);
bool HasOption(const std::string& key) const;

const std::string& GetOptionStr(const std::string& key) const;
int32_t GetOptionInt(const std::string& key) const;
float GetOptionFloat(const std::string& key) const;
```

During decoding, the recognizer would first check whether a given option is defined in the stream. If so, it uses the stream-level option; otherwise, it falls back to the option defined in the recognizer’s configuration.

Community feedback and contributions are very welcome.

dtee1:
Interested

csukuangfj:
Done.

PR #3308, opened 2026-03-12
feat: add SetOption/GetOption C API and export symbols

ZhaoChaoqun:
## Summary

Ref #3101 (Part 1b) — depends on #3307

Add C API functions for the generic SetOption/GetOption mechanism and export them in the macOS symbol file.

## New C API

```c
void SherpaOnnxOnlineStreamSetOption(stream, key, value);
const char *SherpaOnnxOnlineStreamGetOption(stream, key);
void SherpaOnnxOfflineStreamSetOption(stream, key, value);
const char *SherpaOnnxOfflineStreamGetOption(stream, key);
```

## Files Changed (3 files, +65 lines)

| File | Change |
|------|--------|
| `c-api.h` | Function declarations with documentation |
| `c-api.cc` | Implementations calling C++ core |
| `sherpa-onnx-symbols-c.exp` | Export 4 new symbols |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Online and offline streams now support per-stream option configuration—set, retrieve, and query options
  * Online streams can signal the final chunk of data
  * Python bindings added for final chunk signaling functionality

PR #3310, opened 2026-03-12
feat: migrate Paraformer is_final to use SetOption mechanism

ZhaoChaoqun:
## Summary

Ref #3101 — depends on #3309

Migrate the Paraformer `is_final` support from model-specific `SetParaformerFinalChunk`/`IsParaformerFinalChunk` to the generic `SetOption("is_final", "1")` / `GetOptionInt("is_final", 0)` mechanism.

## Changes
- `sherpa-onnx/csrc/online-recognizer-paraformer-impl.h`: Use `GetOptionInt("is_final", 0)` to detect the final chunk and handle short final chunks with zero-padding.

## Usage
Users call `SetOption("is_final", "1")` on the `OnlineStream` before the last decode to process remaining audio frames.

### DIFF
--- a/sherpa-onnx/csrc/online-recognizer-paraformer-impl.h
+++ b/sherpa-onnx/csrc/online-recognizer-paraformer-impl.h
@@ -160,7 +160,16 @@ class OnlineRecognizerParaformerImpl : public OnlineRecognizerImpl {
   }
 
   bool IsReady(OnlineStream *s) const override {
-    return s->GetNumProcessedFrames() + chunk_size_ < s->NumFramesReady();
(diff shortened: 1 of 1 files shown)


ZhaoChaoqun:
@csukuangfj I've rebased and cleaned up this PR:

- Removed the `SetFinalChunk` C API convenience function (users can call `SetOption("is_final", "true")` directly)
- Removed the test scripts (they contained local paths)
- Removed C API changes (already merged in #3308)
- Now only modifies `online-recognizer-paraformer-impl.h` — the core Paraformer logic that uses `GetOption("is_final")` to handle short final chunks

Could you take a look? Thanks!

PR #3563, opened 2026-04-28
fix: add bounds checks to prevent SIGSEGV in speaker diarization

jasagiri:
## Summary

Add bounds checking in three locations within the offline speaker diarization pipeline to prevent out-of-bounds array access that causes SIGSEGV during batch processing:

1. **`SortBySpeaker()`** (`offline-speaker-diarization-result.cc`): Validate speaker ID is within `[0, NumSpeakers())` before indexing into the result vector.

2. **`ReLabel()`** (`offline-speaker-diarization-pyannote-impl.h`): Validate `new_speaker_index` is within the Eigen matrix column bounds before writing.

3. **`valid_indexes` iteration** (`offline-speaker-diarization-pyannote-impl.h`): Validate index bounds before accessing `chunk_speaker_samples_list_pair` vectors after NaN embedding filtering.

## Problem

When processing hundreds of audio files in batch mode via `SherpaOnnxOfflineSpeakerDiarizationProcess`, intermittent SIGSEGV crashes occur. The root cause is that `FastClustering::Cluster()` can produce cluster labels that exceed the expected range (`max_cluster_index + 1`), leading to out-of-bounds writes in `ReLabel()` and out-of-bounds reads in `SortBySpeaker()`.

### Reproduction
- Process 50-200+ audio files sequentially through the same `SherpaOnnxOfflineSpeakerDiarization` instance
- Crashes manifest as SIGSEGV with no stack trace (signal in worker thread)
- Timing is non-deterministic, depends on audio content and clustering results

### Impact
Without these checks, the SIGSEGV crashes the entire host process. With the bounds checks, out-of-range speaker IDs are logged and skipped, allowing processing to continue safely.

## Test plan
- [x] Built and tested on macOS arm64 (Apple M2)
- [x] Batch-processed 139 audio files: subprocess crashes reduced from 5 (stock v1.12.40) to 1 (with bounds checks)
- [x] No regression in diarization quality for in-range speaker IDs

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/sherpa-onnx/csrc/offline-speaker-diarization-pyannote-impl.h
+++ b/sherpa-onnx/csrc/offline-speaker-diarization-pyannote-impl.h
@@ -149,6 +149,15 @@ class OfflineSpeakerDiarizationPyannoteImpl
       chunk_speaker_pair.reserve(valid_indexes.size());
       sample_indexes.reserve(valid_indexes.size());
       for (auto i : valid_indexes) {
+        if (i < 0 ||
+            i >= static_cast<int32_t>(
+                     chunk_speaker_samples_list_pair.first.size())) {
+          SHERPA_ONNX_LOGE("valid_indexes: index %d out of range [0, %d)",
+                           i,
+                           static_cast<int32_t>(
+                               chunk_speaker_samples_list_pair.first.size()));
+          continue;
+        }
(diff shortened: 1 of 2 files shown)


csukuangfj (approved):

Issue #3564, opened 2026-04-28
Allow per-stream use_itn override for SenseVoice (and other models with similar runtime-tunable parameters)


wangxiuwen:
## Use case

Building a multi-tenant ASR HTTP service on top of `sherpa-onnx` (Go binding) where different clients want different ITN behavior per request:

- Voice-control flows want `enableITN=false` so downstream NLU keyword templates can match Chinese number words ("百分之五十") directly.
- LLM/analytics flows want `enableITN=true` so the model emits "50%" — the industry-standard ASR output (Whisper / Google STT / Aliyun NLS / AWS Transcribe all default to this), and information-preserving (`百分之五十 → 50%` is one-to-one; the reverse is ambiguous, e.g. "50" could be "五十" or "五零").

Both kinds of clients hit the same `turing-asr-cpu` service. Today there's no way to honor both without ugly workarounds.

## Current limitation

`SherpaOnnxOfflineSenseVoiceModelConfig.use_itn` is fixed at recognizer-construction time:

```c
typedef struct SherpaOnnxOfflineSenseVoiceModelConfig {
  const char *model;
  const char *language;
  int32_t use_itn;   // baked in at SherpaOnnxCreateOfflineRecognizer()
} SherpaOnnxOfflineSenseVoiceModelConfig;
```

Workarounds today, none clean:

1. **Two recognizers** (one with `use_itn=1`, one with `use_itn=0`) and route per request — doubles model memory (and 2× the bucketed-pool sessions if the deployment uses bucketed batching), warmup time, and HPA pressure for what is fundamentally one model on disk.
2. **Server-side post-processing** (sherpa always emits with ITN, app layer reverses arabic→Chinese for `enableITN=false` clients) — lossy. We hit a real production bug exactly here: `音量调节到百分之五十` got ITN'd to `音量调节到50%` by sherpa, then our `enablePunctuation=false` post-processing stripped `%` as punctuation, then `reverse_itn` mapped `50→五十`, leaving `音量调节到五十` — "百分之" silently disappeared.
3. **Bypass sherpa's `OfflineRecognizer`** entirely and call the SenseVoice ONNX session directly via onnxruntime — loses sherpa's feature extract / VAD / Homophone Replacer / bucketed batching infra. ~900 LOC rewrite.

## Proposal

[...]

PR #3565, opened 2026-04-28
feat(sense-voice): per-stream use_itn override via Stream::GetOptionInt

wangxiuwen:
## Summary

Lets multi-tenant ASR servers select ITN behavior per request without rebuilding the recognizer. Reads `use_itn` from `OfflineStream::GetOptionInt("use_itn", model_default)` on every `Forward()` instead of from `config_.model_config.sense_voice.use_itn` directly.

Both `DecodeStreams` (batch path, lines 194-198) and `DecodeOneStream` (single path, lines 321-323) updated. Batch path now reads per-stream so each stream in a batch can independently override. Backward compatible — callers who never `SetOption("use_itn", ...)` get the existing behavior.

The `OfflineStream::SetOption / GetOptionInt` infrastructure already exists upstream (`offline-stream.h:118-127`), and the C API exposes `SherpaOnnxOfflineStreamSetOption` (`c-api.h:1366`), so callers in any language can do:

```c
SherpaOnnxOfflineStreamSetOption(stream, "use_itn", "1");
// or "0" to override off
// missing key falls back to ModelConfig
```

No new C API, no struct field on `OfflineStream`, no version bump for bindings.

## Why

Multi-tenant ASR services need per-request ITN control — same shape of contract as Whisper API's `enable_text_normalization` or Google STT's `enable_automatic_punctuation`. The current `ModelConfig.SenseVoice.use_itn` baked-in-at-construction model forces production deployments to:

1. Spin up two recognizers (2× memory + 2× warmup) and route per request, **or**
2. Always emit ITN'd text and lossy-reverse downstream (we hit a real bug: `音量调节到百分之五十` got ITN'd to `50%`, our `enablePunctuation=false` postprocessor then stripped `%` as punctuation, then `reverse_itn` mapped `50→五十`, leaving `音量调节到五十` — the "百分之" silently disappeared).

Discussed in #3564.

## Test plan

- [ ] Existing SenseVoice tests still pass with no `SetOption` call (backward compat).
- [ ] New test: `SetOption("use_itn", "1")` and `"0"` produce different outputs on the same audio with the same recognizer instance.
- [ ] Batch path: two streams in the same batch with opposite `use_itn` produce independent results.

## Summary by CodeRabbit

* **New Features**
  * Enhanced speech recognition with per-stream text normalization configuration. Individual streams can now independently customize their text normalization settings, providing greater flexibility when processing multiple concurrent streams, with automatic fallback to model defaults.

### DIFF
--- a/sherpa-onnx/csrc/offline-recognizer-sense-voice-impl.h
+++ b/sherpa-onnx/csrc/offline-recognizer-sense-voice-impl.h
@@ -191,11 +191,15 @@ class OfflineRecognizerSenseVoiceImpl : public OfflineRecognizerImpl {
     std::vector<int32_t> language_array(n);
     std::fill(language_array.begin(), language_array.end(), language);
 
+    // Per-stream use_itn override via OfflineStream::SetOption("use_itn", "1"|"0").
+    // Falls back to model_config.sense_voice.use_itn when the option is absent.
+    int32_t default_use_itn = config_.model_config.sense_voice.use_itn;
     std::vector<int32_t> text_norm_array(n);
-    std::fill(text_norm_array.begin(), text_norm_array.end(),
-              config_.model_config.sense_voice.use_itn
-                  ? meta_data.with_itn_id
-                  : meta_data.without_itn_id);
+    for (int32_t i = 0; i != n; ++i) {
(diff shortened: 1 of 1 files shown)


csukuangfj (approved):

PR #3575, opened 2026-05-05
feat(nemo): add buffered RNNT streaming path for Parakeet Unified

milanleonard:
## Summary

Adds a buffered (left / center / right) streaming RNNT path for NeMo Parakeet Unified models, complementing the existing cache-aware NeMo path. The new path is selected automatically when the encoder ONNX carries `streaming_model_type=buffered_nemo_rnnt` in its metadata, so existing recognizer configs are unaffected.

This mirrors, on the streaming side, the offline export pattern landed in #3556.

## What's added

C++ (under `sherpa-onnx/csrc/`):
- `online-transducer-nemo-buffered-model.{h,cc}` — encoder/decoder/joiner sessions, reads `buffer_{left,chunk,right}_{feature,encoder}_frames`, `subsampling_factor`, `vocab_size`, `pred_rnn_layers`, `pred_hidden`, `normalize_type` from encoder metadata.
- `online-transducer-greedy-search-nemo-buffered-decoder.{h,cc}` — greedy RNNT decode over only the center encoder frames; predictor (LSTM) state carried across chunks via `OnlineStream::{Get,Set}NeMoDecoderStates`.
- `online-recognizer-transducer-nemo-buffered-impl.h` — feature-window assembly with zero padding at stream start/end, per-window `per_feature` normalization, encoder run, center slice, decode.
- `online-recognizer-impl.cc` — small dispatch that inspects encoder metadata once and routes to the buffered impl when applicable.
- `CMakeLists.txt` — registers the two new `.cc` files.

Python (under `scripts/nemo/parakeet-unified-en-0.6b/`, alongside the existing offline scripts):
- `export_onnx_streaming.py` — exports the 2-input buffered encoder + decoder + joiner, writes the buffer-frame metadata, supports fp32 and int8 dynamic quantization, with latency presets (1120ms / 560ms / 240ms).
- `test_onnx_streaming.py` — end-to-end ONNX streaming smoke test.
- `test_buffered_streaming_helpers.py` — unit tests for the windowing / normalization / center-slice helpers.

## How it's wired up

The buffered encoder ONNX is published with these custom metadata keys:
- `streaming_model_type=buffered_nemo_rnnt` — dispatch selector
- `buffer_left_feature_frames`, `buffer_chunk_feature_frames`, `buffer_right_feature_frames`
- `buffer_left_encoder_frames`, `buffer_chunk_encoder_frames`, `buffer_right_encoder_frames`
- `subsampling_factor`, `vocab_size`, `normalize_type`, `pred_rnn_layers`, `pred_hidden`

No new fields on `OnlineModelConfig` — discovery is entirely via metadata, so users point `model.transducer.{encoder,decoder,joiner}` at the buffered ONNXs and everything else (tokens, decoding method, endpointing) is unchanged.

## Testing

- Python ONNX streaming smoke (fp32 and int8 encoders): pass.
- `test_buffered_streaming_helpers.py` (windowing / per-feature normalization / center slice): pass.
- Sherpa-onnx C++ build (online websocket server target on macOS arm64): pass.
- End-to-end decode on a downstream app (OpenWhispr) using the patched online websocket binary against a 1120ms-preset Parakeet Unified buffered ONNX produced expected text on a multi-utterance test sample.
- Existing Zipformer streaming path verified still working through the same binary (no regression in dispatch).

## A note for reviewers on diff size

I want to be upfront that this PR adds ~1.4k new C++ lines, a fair amount of which structurally mirrors the existing `online-transducer-nemo-{model,greedy-search-decoder}.{h,cc}` files. I deliberately chose **separate, parallel files** rather than extending the existing NeMo classes, because:

1. It matches the convention I see in this codebase — `online-recognizer-paraformer-impl.h`, `online-recognizer-transducer-impl.h`, `online-recognizer-transducer-nemo-impl.h`, etc. are each their own file per model family.
2. It keeps the existing cache-aware NeMo path completely untouched, so this PR carries zero risk of regressing models already in use.
3. It makes the new path easy to read in isolation and easy to remove or evolve later without disturbing the existing one.

The alternative would be to fold the buffered metadata + accessors into the existing `OnlineTransducerNeMoModel`, share `OnlineTransducerGreedySearchNeMoDecoder`, and only add the new `*-impl.h` plus the dispatch hook — that would shrink the diff substantially (likely several hundred lines) but would make `OnlineTransducerNeMoModel` dual-purpose (cache-aware *and* buffered) and would touch a file in active use by an existing model family.

I'm very happy to take it in either direction — if you'd prefer the slimmer, shared-classes shape, please let me know and I'll refactor in a follow-up commit on this branch. I went with the parallel-files approach as the safer default for a first contribution here, but I'll defer to your preference on how the codebase should evolve.

Thanks for taking a look, and apologies in advance for any conventions I've missed — happy to adjust anything.

## Release Notes

* **New Features**
  * Enabled streaming/online speech recognition for Parakeet Unified models with adjustable latency profiles
  * Added buffered decoding engine supporting real-time audio processing
  * Introduced feature normalization and windowing utilities for continuous stream input

* **Tests**
  * Added comprehensive tests validating streaming feature processing and ONNX inference pipelines

[Review Change Stack](https://app.coderabbit.ai/change-stack/k2-fsa/sherpa-onnx/pull/3575)


scripts/nemo/parakeet-unified-en-0.6b/export_onnx_streaming.py
```diff
@@ -0,0 +1,133 @@
...
+from pathlib import Path
+from typing import Dict
+
+import nemo.collections.asr as nemo_asr
+import onnx
+import torch
+from onnxruntime.quantization import QuantType, quantize_dynamic
+
+
+SUBSAMPLING_FACTOR = 8
```
csukuangfj: It turns out we can get this value from
```
asr_model.encoder.subsampling_factor
```

Please don't hard code it. 

Please also help fix the merged code to use `asr_model.encoder.subsampling_factor`.

sherpa-onnx/csrc/online-transducer-nemo-buffered-model.h
```diff
@@ -0,0 +1,70 @@
+// sherpa-onnx/csrc/online-transducer-nemo-buffered-model.h
```
csukuangfj: Can you use
```
online-transducer-nemo-parakeet-unified-model.h
```

sherpa-onnx/csrc/online-transducer-nemo-buffered-model.h
```diff
@@ -0,0 +1,70 @@
...
+//
+// Copyright (c)  2026  Xiaomi Corporation
```
csukuangfj: Also, please update it to use your own information.

scripts/nemo/parakeet-unified-en-0.6b/export_onnx_streaming.py
```diff
@@ -0,0 +1,133 @@
...
+        normalize_type = ""
+
+    meta_data = {
+        "vocab_size": asr_model.decoder.vocab_size,
+        "normalize_type": normalize_type,
+        "pred_rnn_layers": asr_model.decoder.pred_rnn_layers,
+        "pred_hidden": asr_model.decoder.pred_hidden,
+        "subsampling_factor": SUBSAMPLING_FACTOR,
+        "model_type": "EncDecRNNTBPEModel",
+        "streaming_model_type": "buffered_nemo_rnnt",
```
csukuangfj: Can you use
```
nemo_parakeet_unified_streaming
```
as the model type?

csukuangfj: You need to also change the `.cc` model_type validation code if you change the model type here.

scripts/nemo/parakeet-unified-en-0.6b/export_onnx_streaming.py
```diff
@@ -0,0 +1,133 @@
...
+        "buffered_streaming": 1,
+        "buffer_left_encoder_frames": preset["left"],
+        "buffer_chunk_encoder_frames": preset["chunk"],
+        "buffer_right_encoder_frames": preset["right"],
+        "buffer_left_feature_frames": preset["left"] * SUBSAMPLING_FACTOR,
+        "buffer_chunk_feature_frames": preset["chunk"] * SUBSAMPLING_FACTOR,
+        "buffer_right_feature_frames": preset["right"] * SUBSAMPLING_FACTOR,
```
csukuangfj: Please remove the prefix `buffer_` to keep it simpler, e.g., use `left_encoder_frames` to replace `buffer_left_encoder_frames`.

sherpa-onnx/csrc/online-transducer-greedy-search-nemo-buffered-decoder.cc
```diff
@@ -0,0 +1,119 @@
+// sherpa-onnx/csrc/online-transducer-greedy-search-nemo-buffered-decoder.cc
```
csukuangfj: Can you use
```
online-transducer-greedy-search-nemo-parakeet-unified-decoder.cc
```

sherpa-onnx/csrc/online-recognizer-impl.cc
```diff
@@ -68,6 +76,15 @@ std::unique_ptr<OnlineRecognizerImpl> OnlineRecognizerImpl::Create(
     sess_opts.SetIntraOpNumThreads(1);
     sess_opts.SetInterOpNumThreads(1);
 
+    auto encoder_sess = std::make_unique<Ort::Session>(
```
csukuangfj: The encoder is much larger than the joiner and the decoder, which means it takes longer to load it into memory and slows down the initialization.

https://github.com/k2-fsa/sherpa-onnx/blob/0f6e4c007cfe73cfbf61917ab2c54d4cedbc0b37/sherpa-onnx/csrc/online-recognizer-impl.cc#L133

the above line already loads the decoder model into the memory. Can you also add meta data to the decoder model and read the model type from it?

Note that decoder models other than parakeet unified streaming may not contain meta data.

sherpa-onnx/csrc/online-recognizer-transducer-nemo-buffered-impl.h
```diff
@@ -0,0 +1,322 @@
...
+class OnlineRecognizerTransducerNeMoBufferedImpl
+    : public OnlineRecognizerImpl {
+ public:
+  explicit OnlineRecognizerTransducerNeMoBufferedImpl(
+      const OnlineRecognizerConfig &config)
+      : OnlineRecognizerImpl(config),
+        config_(config),
+        endpoint_(config_.endpoint_config),
+        model_(std::make_unique<OnlineTransducerNeMoBufferedModel>(
+            config.model_config)) {
```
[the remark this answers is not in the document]
csukuangfj: please fix this comment.

sherpa-onnx/csrc/online-recognizer-transducer-nemo-buffered-impl.h
```diff
@@ -0,0 +1,322 @@
...
+
+    if (model_->FeatureNormalizationMethod() != "per_feature" &&
+        !model_->FeatureNormalizationMethod().empty()) {
+      SHERPA_ONNX_LOGE("Unsupported NeMo buffered normalization: %s",
+                       model_->FeatureNormalizationMethod().c_str());
+      SHERPA_ONNX_EXIT(-1);
+    }
+  }
+
+  static void NormalizePerFeature(float *features, int32_t num_frames,
```
csukuangfj: Can you move
 https://github.com/k2-fsa/sherpa-onnx/blob/f3b1a9da8d6e0e6cdb21387d41d25f8c8e10d3cc/sherpa-onnx/csrc/offline-stream.cc#L303

to math.h and math.cc
and invoke it here to reduce code duplicate?

sherpa-onnx/csrc/online-recognizer-transducer-nemo-buffered-impl.h
```diff
@@ -0,0 +1,322 @@
...
+    const float *src = encoder_out->GetTensorData<float>();
+    std::vector<float> center((end - start) * num_cols);
+    std::copy(src + start * num_cols, src + end * num_cols, center.begin());
+
+    std::array<int64_t, 3> center_shape{1, end - start, num_cols};
+    Ort::Value ans = Ort::Value::CreateTensor<float>(
+        model_->Allocator(), center_shape.data(), center_shape.size());
+
+    float *dst = ans.GetTensorMutableData<float>();
+    std::copy(center.begin(), center.end(), dst);
```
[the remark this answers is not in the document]
csukuangfj: Please fix this comment.

sherpa-onnx/csrc/online-recognizer-transducer-nemo-parakeet-unified-impl.h
```diff
@@ -0,0 +1,322 @@
...
+        x_shape.size());
+
+    std::array<int64_t, 1> length_shape{1};
+    Ort::Value length = Ort::Value::CreateTensor<int64_t>(
+        model_->Allocator(), length_shape.data(), length_shape.size());
+    length.GetTensorMutableData<int64_t>()[0] = total_feature_frames;
+
+    auto encoder_outs = model_->RunEncoder(std::move(x), std::move(length));
+    Ort::Value encoder_out = Transpose12(model_->Allocator(), &encoder_outs[0]);
+    Ort::Value center_encoder_out =
```
csukuangfj, line 275: Please change `SliceCenterEncoderOut`.

To avoid memory copying, create an `Ort::Value` that shares the underlying memory with `encoder_out`. Simply pass the existing data pointer and specify the correct tensor shape when constructing the value.

sherpa-onnx/csrc/online-recognizer-transducer-nemo-buffered-impl.h
```diff
@@ -0,0 +1,322 @@
...
+    int32_t valid_center_frames =
+        std::min(model_->ChunkFeatureFrames(), ready - processed);
+
+    if (valid_center_frames <= 0) {
+      return;
+    }
+
+    int32_t feature_dim = model_->FeatureDim();
+    int32_t total_feature_frames = model_->TotalFeatureFrames();
+    std::vector<float> features(total_feature_frames * feature_dim);
```
csukuangfj: `features` is initialized to 0 in its constructor by default. No need to re-initialize to 0 in `CopyWindowWithPadding`. 

Please remove the `std::fill` in `CopyWindowWithPadding`.

csukuangfj: A better approach is to change `CopyWindowWithPadding` to return a `std::vector<float>`.

csukuangfj (changes requested):
Thank you for your contribution! The implementation looks excellent. I’ve left a few minor comments for review.
Regarding the naming, I prefer `parakeet-unified` over `nemo-buffered`. 

(Since `online` is already in the name, adding `streaming` would be redundant.)

Once you update the name, please remember to also update the class name and any related code comments.

csukuangfj:
By the way, can you show the decoding results using `build/bin/sherpa-onnx` with this PR and the exported ONNX model?

You can put the decoding command and logs in the PR comments.

milanleonard:
Updated the PR branch with the review follow-up.

Summary of changes:
- Renamed the C++ path from `nemo-buffered` to `nemo-parakeet-unified`, including file names, classes, includes, CMake wiring, and comments.
- Updated the streaming metadata contract to `streaming_model_type=nemo_parakeet_unified_streaming` and removed the `buffer_` prefix from frame metadata keys.
- Added `streaming_model_type` metadata to decoder ONNX outputs and changed recognizer dispatch to inspect the already-loaded decoder session instead of loading the encoder for detection.
- Replaced the hardcoded subsampling factor with `asr_model.encoder.subsampling_factor` in both the new streaming export and the existing Parakeet Unified export.
- Moved NeMo per-feature normalization into `math.h/.cc` and reused it from both offline and online paths.
- Added metadata validation, removed the redundant padding fill, changed center encoder slicing to use an `Ort::Value` view over the existing encoder output memory, fixed initializer ordering, and adjusted decoder state persistence.

I also ran a smoke decode with `build/bin/sherpa-onnx` using local ONNX copies patched to the new metadata contract:

```bash
build/bin/sherpa-onnx \
  --tokens=build/parakeet-unified-smoke/tokens.txt \
  --encoder=build/parakeet-unified-smoke/encoder.int8.onnx \
  --decoder=build/parakeet-unified-smoke/decoder.onnx \
  --joiner=build/parakeet-unified-smoke/joiner.onnx \
  --provider=cpu \
  --num-threads=2 \
  --decoding-method=greedy_search \
  /Users/milanleonard/.cache/openwhispr/parakeet-models/parakeet-unified-en-0.6b/test_wavs/0.wav
```

Output:

```text
Number of threads: 2, Elapsed seconds: 3.8, Audio duration (s): 7.4, Real time factor (RTF) = 3.8/7.4 = 0.52
 Well, I don't wish to see it anymore, observed Phoebe, turning away her eyes. It is certainly very like the old portrait
Recognizer created in 1.07495 s
```

Verification run locally:

```text
.venv/bin/python scripts/nemo/parakeet-unified-en-0.6b/test_buffered_streaming_helpers.py
.venv/bin/python -m py_compile scripts/nemo/parakeet-unified-en-0.6b/export_onnx.py scripts/nemo/parakeet-unified-en-0.6b/export_onnx_streaming.py scripts/nemo/parakeet-unified-en-0.6b/test_onnx_streaming.py scripts/nemo/parakeet-unified-en-0.6b/buffered_streaming_helpers.py
cmake --build build --target sherpa-onnx
cmake --build build --target sherpa-onnx-online-websocket-server
git diff --check
```

sherpa-onnx/csrc/online-recognizer-transducer-nemo-parakeet-unified-impl.h
milanleonard: Changed `SliceCenterEncoderOut` to create the center tensor as an `Ort::Value` view over the existing transposed encoder output buffer. It now passes the offset data pointer and center shape directly, so the center encoder frames are not copied.

sherpa-onnx/csrc/online-recognizer-transducer-nemo-parakeet-unified-impl.h
```diff
@@ -0,0 +1,322 @@
...
+        SliceCenterEncoderOut(&encoder_out, valid_center_encoder_frames);
+
+    s->GetNumProcessedFrames() += valid_center_frames;
+
+    OnlineStream *streams[1] = {s};
+    decoder_->Decode(std::move(center_encoder_out), streams, 1);
+  }
+
+ private:
+  OnlineRecognizerConfig config_;
```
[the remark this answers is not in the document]
milanleonard: I left this one unchanged intentionally. `OnlineRecognizerImpl::config_` is private, and nearby recognizer implementations such as `OnlineRecognizerTransducerNeMoImpl` keep their own `config_` because they mutate model-specific feature settings after construction. Changing that would be a broader recognizer refactor rather than a targeted Parakeet Unified follow-up.

sherpa-onnx/csrc/online-recognizer-impl.cc
milanleonard: Implemented by writing `streaming_model_type=nemo_parakeet_unified_streaming` to `decoder.onnx` and `decoder.int8.onnx`, then dispatching from the already-loaded decoder session metadata. If the decoder metadata is absent, the existing recognizer factory falls through to the previous node-count based paths.

scripts/nemo/parakeet-unified-en-0.6b/export_onnx_streaming.py
milanleonard: Changed both the streaming export script and the already-merged Parakeet Unified export script to use `asr_model.encoder.subsampling_factor` instead of a hardcoded value.

scripts/nemo/parakeet-unified-en-0.6b/export_onnx_streaming.py
milanleonard: Updated the export metadata and C++ validation to use `nemo_parakeet_unified_streaming`.

scripts/nemo/parakeet-unified-en-0.6b/export_onnx_streaming.py
milanleonard: Removed the `buffer_` prefix from the exported frame metadata keys and updated the C++ runtime/test reads to use `left_feature_frames`, `chunk_feature_frames`, `right_feature_frames`, `left_encoder_frames`, `chunk_encoder_frames`, and `right_encoder_frames`.

sherpa-onnx/csrc/online-recognizer-transducer-nemo-buffered-impl.h
milanleonard: Moved the NeMo per-feature normalization helper into `math.h/.cc` and now call it from both `offline-stream.cc` and the Parakeet Unified online path.

sherpa-onnx/csrc/online-recognizer-transducer-nemo-buffered-impl.h
milanleonard: Changed `CopyWindowWithPadding` to return a zero-initialized `std::vector<float>` and removed the redundant `std::fill`.

sherpa-onnx/csrc/online-transducer-nemo-buffered-model.cc
```diff
@@ -0,0 +1,425 @@
...
+    SHERPA_ONNX_READ_META_DATA(chunk_feature_frames_,
+                               "buffer_chunk_feature_frames");
+    SHERPA_ONNX_READ_META_DATA(right_feature_frames_,
+                               "buffer_right_feature_frames");
+    SHERPA_ONNX_READ_META_DATA(left_encoder_frames_,
+                               "buffer_left_encoder_frames");
+    SHERPA_ONNX_READ_META_DATA(chunk_encoder_frames_,
+                               "buffer_chunk_encoder_frames");
+    SHERPA_ONNX_READ_META_DATA(right_encoder_frames_,
+                               "buffer_right_encoder_frames");
```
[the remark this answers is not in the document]
milanleonard: Added validation for the required numeric metadata: positive vocab/subsampling/feature/predictor/chunk values and nonnegative left/right context values.

sherpa-onnx/csrc/online-recognizer-transducer-nemo-buffered-impl.h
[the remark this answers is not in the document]
milanleonard: Fixed the constructor initializer order so it matches the member declaration order.

sherpa-onnx/csrc/online-recognizer-transducer-nemo-buffered-impl.h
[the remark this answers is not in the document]
milanleonard: Changed center slicing to avoid the intermediate center vector; the current implementation uses an `Ort::Value` view over the existing encoder output buffer.

sherpa-onnx/csrc/online-recognizer-impl.cc
```diff
@@ -68,6 +76,15 @@ std::unique_ptr<OnlineRecognizerImpl> OnlineRecognizerImpl::Create(
...
+        env, SHERPA_ONNX_TO_ORT_PATH(config.model_config.transducer.encoder),
+        sess_opts);
```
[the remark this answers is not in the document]
milanleonard: Addressed by avoiding encoder loading in recognizer factory dispatch. The factory now reads Parakeet Unified streaming metadata from the decoder session that it already created.

sherpa-onnx/csrc/online-transducer-nemo-buffered-model.h
milanleonard: Renamed the model files to `online-transducer-nemo-parakeet-unified-model.{h,cc}` and updated the corresponding includes/classes.

sherpa-onnx/csrc/online-transducer-nemo-buffered-model.h
milanleonard: Updated the new-file copyright headers to my own information.

sherpa-onnx/csrc/online-transducer-greedy-search-nemo-buffered-decoder.cc
milanleonard: Renamed the greedy decoder implementation to `online-transducer-greedy-search-nemo-parakeet-unified-decoder.{h,cc}` and updated the class/includes accordingly.

csukuangfj (approved):
Thanks again for your contribution!

We will upload the exported model and update the doc after merging this PR.

jvcrnc:
Nice work. 
I noticed you don't make use of the cache-aware feature, which supports much faster recognition:
asr_model.set_export_config({"cache_support": True})
I tried myself, but have onnx-runtime crashes with the cache-aware model, maybe you ran into the same issue?

jvcrnc:
I just read cache-aware currently is not supported by the model:
The current inference pipeline supports only buffered streaming (left context is recomputed for each chunk) that can be longer than cache-aware streaming.

PR #3578, opened 2026-05-06
Fix x64 gpu onnxruntime url

difey:
The url of onnxruntime provided by x64-gpu cmake file does not exist at all. Replace the url to the release by microsoft.

## Summary by CodeRabbit

* **Chores**
  * Updated Linux x86_64 GPU build to use patched onnxruntime ZIP artifacts (1.24.4-patched) alongside existing CUDA 13 targets.
  * Revised download verification checksum to match the patched artifact.
  * Expanded offline/fallback artifact locations to include patched ZIP paths and added an installation step for the extracted onnxruntime library files.


cmake/onnxruntime-linux-x86_64-gpu.cmake
```diff
@@ -20,19 +20,19 @@ endif()
 
 
 # Requires CUDA 12, cudnn 9
-set(onnxruntime_URL  "https://github.com/csukuangfj/onnxruntime-libs/releases/download/v1.24.4/onnxruntime-linux-x64-gpu-1.24.4.tgz")
```
csukuangfj, line 23: Can you use https://github.com/csukuangfj/onnxruntime-libs/releases/download/v1.24.4/onnxruntime-linux-x64-gpu-1.24.4-patched.zip

It is generated by
https://github.com/csukuangfj/onnxruntime-libs/actions/runs/24170579079/job/70541254384

difey: It looks good to me.

csukuangfj (approved):

PR #3598, opened 2026-05-11
feat(offline-recognizer): expose Ort::RunOptions config entries (transducer)

jonghyunyoo-pinedh:
## Motivation

  `sherpa-onnx` wraps ONNX Runtime but doesn't surface `Ort::RunOptions` to callers,  
  so downstream Python (and Node) users have no way to drive per-Run controls such as 
  **`memory.enable_memory_arena_shrinkage`**. Production deployments that stream      
  variable-length audio through an `OfflineRecognizer` end up resorting to workarounds
   — fixed input-shape buckets, `glibc malloc_trim` via `ctypes`, etc. — to keep RSS  
  bounded, none of which can fully replace ORT's own arena management.

  ## What this PR does

  Adds an opt-in API to forward `Ort::RunOptions` config entries through to the       
  `Session::Run` calls performed by the recognizer:

  ```cpp
  // C++
  recognizer.SetRunOptionsConfigEntry(
      "memory.enable_memory_arena_shrinkage", "cpu:0");
  ```

  ```python
  # Python (single entry)
  recognizer.set_run_options_config_entry(
      "memory.enable_memory_arena_shrinkage", "cpu:0")

  # Python (dict convenience)
  recognizer.set_run_options({
      "memory.enable_memory_arena_shrinkage": "cpu:0",
  })
  ```

  Wired up for `OfflineRecognizerTransducerImpl` → `OfflineTransducerModel`. All three
   encoder/decoder/joiner `Session::Run` calls now use the model-owned
  `Ort::RunOptions` member (default-constructed = empty = behaves identically to the  
  previous `{}`).

  Other recognizer types (paraformer / whisper / sense-voice / ctc / canary / cohere /
   fire-red / funasr-nano / moonshine / qwen3) inherit the base
  `OfflineRecognizerImpl::SetRunOptionsConfigEntry` default, which logs a warning and 
  ignores the setting. Extending them is straightforward and can land in follow-up    
  PRs.

  ## Commits (3)

  1. **`Add OfflineRecognizer::SetRunOptionsConfigEntry (transducer)`** — C++ public  
  API + base no-op + transducer override + model-level wiring (3 `Session::Run`       
  sites).
  2. **`Expose SetRunOptionsConfigEntry to Python`** — pybind11 binding (single-entry 
  + dict convenience) + high-level `OfflineRecognizer` Python wrapper.
  3. **`Add Python tests for OfflineRecognizer set_run_options*`** — 3 smoke tests    
  (single entry / dict / unknown key); skip when no test model is available.

  ## Backward compatibility

  - Default `Ort::RunOptions{}` is identical to the prior empty `{}` argument — no    
  behavioral change unless `SetRunOptionsConfigEntry` is called.
  - `OfflineRecognizerImpl::SetRunOptionsConfigEntry` is a base virtual with a logging
   no-op, so calling it on a non-transducer recognizer is safe.
  - pybind11 surface is additive only; no signatures changed.
  - ABI: a new pure virtual is **not** added — base impl provides a default body.     
  Adding a non-virtual member to `OfflineTransducerModel::Impl` (pimpl) doesn't affect
   the public ABI.

  ## Thread-safety

  `SetRunOptionsConfigEntry` is intended to be called once after recognizer
  construction, before the first decode. It is not synchronized with concurrent       
  `DecodeStream`/`DecodeStreams` calls; this matches how `Ort::RunOptions` itself is  
  documented. The Python wrapper releases the GIL for the single-entry form; the dict 
  form holds the GIL while iterating but otherwise behaves identically.

  ## Scope this PR intentionally does *not* cover

  - **Other recognizer types.** Each impl forwards to its own model class —
  straightforward but mechanical, easier to review as separate PRs.
  - **`OnlineRecognizer` / streaming variants.** Same pattern applies, also deferred. 
  - **`sherpa-onnx-node` / Java / Kotlin / Swift / etc. bindings.** Separate PRs      
  (likely different maintainers).

  ## Tested

  - Local builds pass on the existing test fixtures (skip when models absent).        
  - End-to-end RSS profiling on a downstream Python service (FastAPI WSS server, ~30  
  utterances/min) shows the expected reduction in resident memory growth when
  `memory.enable_memory_arena_shrinkage="cpu:0"` is enabled, with negligible p50      
  latency change and a small (~1–5 ms) p99 increase from arena reallocation — exactly 
  the trade-off this opt-in is designed to expose.

  Happy to adjust naming (`SetRunOptionsConfigEntry` vs `AddRunOptionsConfigEntry` vs 
  other) or split the commits further if preferred.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring ONNX Runtime options in the offline recognizer. Users can now inject custom run-option configuration entries for decode operations via new configuration methods.

* **Tests**
  * Added test coverage for run-options configuration functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/k2-fsa/sherpa-onnx/pull/3598)

### DIFF
--- a/sherpa-onnx/csrc/offline-recognizer-impl.cc
+++ b/sherpa-onnx/csrc/offline-recognizer-impl.cc
@@ -893,6 +893,14 @@ void OfflineRecognizerImpl::SetConfig(const OfflineRecognizerConfig &config) {
   config_ = config;
 }
 
+void OfflineRecognizerImpl::SetRunOptionsConfigEntry(const std::string &key,
+                                                     const std::string &value) {
+  SHERPA_ONNX_LOGE(
+      "SetRunOptionsConfigEntry is not implemented for this recognizer type. "
+      "The setting (%s = %s) will be ignored.",
+      key.c_str(), value.c_str());
+}
+
 #if __ANDROID_API__ >= 9
 template OfflineRecognizerImpl::OfflineRecognizerImpl(
     AAssetManager *mgr, const OfflineRecognizerConfig &config);
--- a/sherpa-onnx/csrc/offline-recognizer-impl.h
+++ b/sherpa-onnx/csrc/offline-recognizer-impl.h
@@ -47,6 +47,17 @@ class OfflineRecognizerImpl {
 
   virtual OfflineRecognizerConfig GetConfig() const = 0;
 
+  // Inject an Ort::RunOptions config entry (e.g.,
+  // "memory.enable_memory_arena_shrinkage" = "cpu:0") that will be applied
+  // to all subsequent Session::Run calls of this recognizer.
+  //
+  // The default implementation logs a warning and ignores the setting.
+  // Concrete recognizers that route their inference through a model class
+  // owning the Ort::Session(s) should override this and forward the entry
+  // to the underlying model.
+  virtual void SetRunOptionsConfigEntry(const std::string &key,
+                                        const std::string &value);
+
   std::string ApplyInverseTextNormalization(std::string text) const;
 
   std::string ApplyHomophoneReplacer(std::string text) const;
--- a/sherpa-onnx/csrc/offline-recognizer-transducer-impl.h
+++ b/sherpa-onnx/csrc/offline-recognizer-transducer-impl.h
@@ -256,6 +256,13 @@ class OfflineRecognizerTransducerImpl : public OfflineRecognizerImpl {
 
   OfflineRecognizerConfig GetConfig() const override { return config_; }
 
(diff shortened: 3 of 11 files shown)


csukuangfj:
If you really need this function, please don't add new methods to recognizers. Instead, you can use the SetOption from Online/OfflineStreams.

See also https://github.com/k2-fsa/sherpa-onnx/issues?q=setoption

Note that you don't need to save the run option inside the model. Your approach is not thread-safe.
You can fetch the run options from a stream and pass them to a model when you invoke `RunXXX`, e.g., `RunDecoder()`.

PR #3615, opened 2026-05-15
docs: add React Native sherpa-onnx wrapper

abretonc7s:
## Summary

Thank you for building sherpa-onnx. I maintain `@siteed/sherpa-onnx.rn`, a React Native wrapper for using sherpa-onnx in iOS, Android, and Web apps, plus a companion demo app called Sherpa Voice.

This PR adds the wrapper and demo app to the "Projects using sherpa-onnx" section so React Native users can find it.

Links:

- NPM: https://www.npmjs.com/package/@siteed/sherpa-onnx.rn
- Source: https://github.com/deeeed/audiolab/tree/main/packages/sherpa-onnx.rn
- Demo app: https://deeeed.github.io/audiolab/sherpa-voice/

## Notes

The wrapper is used to validate sherpa-onnx features across React Native apps, including ASR, TTS, VAD, KWS, speaker ID, diarization, language ID, punctuation, audio tagging, and speech enhancement.

Happy to adjust wording or move this elsewhere if maintainers prefer a different location.

## Summary by CodeRabbit

* **Documentation**
  * README updated with a new "Projects using sherpa-onnx" entry for Sherpa Voice — a React Native demo showcasing ASR, TTS, VAD, KWS, speaker ID/diarization, and language ID; includes links to the NPM package, project repository, and live demo site.

[Review Change Stack](https://app.coderabbit.ai/change-stack/k2-fsa/sherpa-onnx/pull/3615?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #3617, opened 2026-05-15
build: allow overriding ONNX Runtime for mobile

abretonc7s:
## Summary

This is separate from #3615, which only lists the React Native wrapper project.

This PR makes the mobile build scripts more reusable by allowing callers to provide an existing ONNX Runtime build instead of always using the hardcoded default download.

Added knobs:

- `SHERPA_ONNX_ONNXRUNTIME_VERSION` — override the downloaded ONNX Runtime version while preserving the current default.
- `SHERPA_ONNX_ONNXRUNTIME_ROOT` — Android shared-build convenience path with `jni/<abi>/` and `headers/` layout.
- Existing `SHERPA_ONNXRUNTIME_LIB_DIR` + `SHERPA_ONNXRUNTIME_INCLUDE_DIR` are now honored by Android wrapper scripts before downloading.

Why:

React Native apps often package multiple native ASR/TTS libraries. Letting the wrapper build reuse a caller-provided ONNX Runtime makes it easier to keep Android/iOS native packaging aligned and avoid local patching of sherpa-onnx build scripts.

## Validation

- `bash -n build-android-arm64-v8a.sh build-android-armv7-eabi.sh build-android-x86-64.sh build-ios.sh`

Defaults are unchanged when the new environment variables are not set.

## Summary by CodeRabbit

* **Chores**
  * ONNX Runtime version is now configurable via environment variables (defaults: Android v1.24.3, iOS v1.17.1) instead of hardcoded values.
  * Supports externally provided ONNX Runtime library and header directories; when supplied they are validated, normalized, and used instead of derived paths.
  * Shared-library installation is now conditional for shared builds and copies the runtime from the resolved library location.

[Review Change Stack](https://app.coderabbit.ai/change-stack/k2-fsa/sherpa-onnx/pull/3617?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


abretonc7s:
Thanks for the reviews. I verified the Android findings were real and pushed fixes in `f345d731` and `c87b5ae1`.

Fixed:
- shared Android builds now copy `libonnxruntime.so` from the selected `SHERPA_ONNXRUNTIME_LIB_DIR`, not the downloaded-version path
- the copy no longer silently succeeds when the selected runtime is missing
- override/root directories are resolved before entering the per-ABI build directory, so relative paths are not accidentally interpreted under `build-android-*`
- override path normalization now only runs for the selected override mode and emits explicit missing-directory errors

Validation:
- `bash -n build-android-arm64-v8a.sh build-android-armv7-eabi.sh build-android-x86-64.sh build-ios.sh`
- `git diff --check`
- confirmed no remaining hardcoded Android `cp -fv $onnxruntime_version/jni/.../libonnxruntime.so` packaging copy

build-android-arm64-v8a.sh
```diff
@@ -22,6 +22,28 @@ else
   dir=$PWD/build-android-arm64-v8a-static
```
csukuangfj, line 22: Can you also update build-android-x64.sh?

csukuangfj (changes requested):

build-android-arm64-v8a.sh
abretonc7s: Done in d8269ca5. I updated `build-android-x86.sh` as well so it now supports the same ONNX Runtime override knobs:

- `SHERPA_ONNX_ONNXRUNTIME_VERSION`
- `SHERPA_ONNXRUNTIME_LIB_DIR` + `SHERPA_ONNXRUNTIME_INCLUDE_DIR`
- `SHERPA_ONNX_ONNXRUNTIME_ROOT` with `jni/x86/` + `headers/`

Validation:
- `bash -n build-android-arm64-v8a.sh build-android-armv7-eabi.sh build-android-x86-64.sh build-android-x86.sh build-ios.sh`
- `git diff --check`
- stubbed Android x86 build path with a fake NDK/cmake/make and fake `SHERPA_ONNX_ONNXRUNTIME_ROOT`; verified `install/lib/libonnxruntime.so` is copied from `jni/x86/libonnxruntime.so`.

csukuangfj (approved):

PR #3657, opened 2026-06-03
Fix modified_beam_search hallucination/empty text for NeMo TDT

### LINKED ISSUE #3267: modified_beam_search with NeMo TDT (Parakeet) hallucinates or returns empty text ~20% of the time; greedy_search works
> **Disclosure**: This issue was filed by an AI coding agent (Claude Code, by Anthropic) on behalf of a developer. The technical details, reproduction steps, and analysis are based on real-world testing in a production Electron application.

## Summary

When using `modified_beam_search` with the `parakeet-tdt-0.6b-v3` model (NeMo transducer) via the offline WebSocket server binary, approximately 20% of transcription requests return either:
- The hallucinated text **"Yeah."** (on audio containing clear speech), or
- **Empty text** (on audio with good signal, RMS ~0.15)

Switching to `greedy_search` (by omitting `--decoding-method`) on the **same audio** produces correct transcriptions every time.

## Environment

- **sherpa-onnx version**: v1.12.25 (pre-built Windows x64 binary)
- **Binary**: `sherpa-onnx-ws-win32-x64.exe` (offline WebSocket server)
- **Model**: `parakeet-tdt-0.6b-v3` (encoder.int8.onnx, decoder.int8.onnx, joiner.int8.onnx)
- **OS**: Windows 11 x64
- **Hardware**: CPU inference (no CUDA)

## Server launch command (failing)

```bash
sherpa-onnx-ws-win32-x64.exe \
  --encoder=encoder.int8.onnx \
  --decoder=decoder.int8.onnx \
  --joiner=joiner.int8.onnx \
  --tokens=tokens.txt \
  --port=6006 \
  --decoding-method=modified_beam_search \
  --hotwords-file=hotwords.txt \
  --hotwords-score=1.5 \
  --modeling-unit=bpe \
  --bpe-vocab=bpe.vocab
```

## Server launch command (working)

```bash
sherpa-onnx-ws-win32-x64.exe \
  --encoder=encoder.int8.onnx \
  --decoder=decoder.int8.onnx \
  --joiner=joiner.int8.onnx \
  --tokens=tokens.txt \
  --port=6006
```

(Defaults to `greedy_search` — works correctly on all test audio.)

## Reproduction steps

1. Download parakeet-tdt-0.6b-v3 model files
2. Create `hotwords.txt` with a few English words (e.g., one word per line — content doesn't matter, see below)
3. Start the WS server with the "failing" command above
4. Send 30+ transcription requests via WebSocket with clear English speech audio (16kHz mono float32)
5. Observe: ~6/30 return "Yeah.", ~4/30 return empty, ~20/30 return correct text
6. Restart server with the "working" command (no `--decoding-method` flag)
7. Repeat step 4 with the same audio files
8. Observe: 30/30 return correct transcriptions

## Key observations

- **Not caused by hotwords content**: Testing with an empty hotwords file + `--hotwords-score=0.2` still produces hallucinations. The trigger is `--decoding-method=modified_beam_search` itself.
- **Not caused by audio quality**: The same audio that returns "Yeah." with MBS returns perfect transcription with greedy search. Audio RMS is ~0.15 (not silence).
- **Not caused by warm-up state**: A 1-second silence warm-up inference runs before test trials. Hallucinations persist after warm-up.
- **Intermittent**: The same audio file may succeed on one request and hallucinate on the next. No deterministic pattern.
- **Affects the NeMo TDT decoder specifically**: This uses the `OfflineTransducerModifiedBeamSearchNeMoDecoder` code path added in PR #3077.

## Likely related issues

- **#845**: "Decoding method 'modified_beam_search' gives letters/words on silence" — same symptom (MBS hallucinates, greedy works) but on a custom zipformer model, not NeMo TDT. The maintainer was unable to identify the root cause. Our report may share the same underlying bug in the MBS decoder.
- **#2541 / PR #3077**: Added `modified_beam_search` support for NeMo transducer models (merged Feb 5, 2026). The NeMo MBS decoder is relatively new and may have less field testing.

## Workaround

We default to `greedy_search` and only enable `modified_beam_search` when the user explicitly opts in to hotword boosting (with a warning that it's experimental). This eliminates the hallucination for the default path.

## WebSocket protocol used

Standard sherpa-onnx offline WS binary protocol:
```
[int32LE sample_rate][int32LE num_audio_bytes][float32 samples...]
```

Response: JSON with `text` field. After receiving response, client sends `"Done"` and connection closes.

## What we'd like

1. Confirmation of whether this is a known limitation of the NeMo MBS decoder
2. If it's a bug, a fix so that `modified_beam_search` + hotwords works reliably with NeMo TDT models
3. Any recommended `--hotwords-score` values or configuration that might reduce the hallucination rate

Thank you for sherpa-onnx — it's an excellent project and Parakeet TDT transcription quality with `greedy_search` is outstanding.

kittysoftpaw0510:
Fixes #3267

## Problem

`modified_beam_search` with NeMo TDT models (e.g. Parakeet-TDT-0.6b-v2) produces hallucinated output ("Yeah.") or empty text ~20% of the time.

## Root Cause

Three bugs in `OfflineTransducerModifiedBeamSearchNeMoDecoder`, identified by comparing against the working greedy decoder (`DecodeOneTDT`):

### 1. Missing blank+skip==0 guard (primary cause)
The greedy TDT decoder has a critical safety check:
```cpp
if (y == blank_id && skip == 0) {
    skip = 1;  // Force advance to prevent infinite loop
}
```
The beam search decoder was missing this. When blank was predicted with duration 0, the hypothesis stayed on the same frame with `num_symbols` reset to 0, creating a loop that produced empty or hallucinated output.

### 2. `duration_log_prob` added to blank candidates
The duration probability was applied to ALL candidates uniformly. For blank tokens, the decoder state is unchanged and frame advancement uses different logic -- including the duration score skewed beam ranking toward blank-heavy hypotheses that raced through frames without emitting tokens.

### 3. `max_symbols_per_frame` = 10 vs greedy decoder 5
The higher limit allowed more spurious token emissions per frame, contributing to hallucinated short phrases like "Yeah."

## Changes

- Blank candidates now scored without `duration_log_prob`
- Blank with TDT always advances by `std::max(1, predicted_skip)` frames (explicit guard)
- `max_symbols_per_frame` reduced from 10 to 5 to match greedy TDT decoder

## Test plan

- [ ] Run `modified_beam_search` on audio that previously produced "Yeah." or empty text
- [ ] Verify hotword boosting still works correctly with Parakeet-TDT models
- [ ] Compare WER between greedy and modified_beam_search on a test set

## Summary by CodeRabbit

* **Refactor**
  * Cap per-frame emitted non-blank symbols to 5 in time-dependent mode, reducing spurious emissions.
  * Split blank/unknown vs non-blank scoring paths: blanks no longer use duration contribution and reset per-frame counts; non-blanks include duration scores and update decoder state.
  * Preserve and reorganize context-graph boosting so it’s applied at the appropriate expansion stage.

### DIFF
--- a/sherpa-onnx/csrc/offline-transducer-modified-beam-search-nemo-decoder.cc
+++ b/sherpa-onnx/csrc/offline-transducer-modified-beam-search-nemo-decoder.cc
@@ -94,7 +94,7 @@ OfflineTransducerModifiedBeamSearchNeMoDecoder::Decode(
 
   int32_t vocab_size = model_->VocabSize();
   int32_t blank_id = vocab_size - 1;  // NeMo models have blank at the end
-  int32_t max_symbols_per_frame = 5;  // Match greedy TDT decoder limit
+  int32_t max_symbols_per_frame = is_tdt_ ? 5 : 10;  // TDT: match greedy decoder limit
 
   // For TDT models, we need to know the number of duration bins
   // We'll detect this from the joiner output size on first run
--- a/sherpa-onnx/csrc/offline-transducer-modified-beam-search-nemo-decoder.cc
+++ b/sherpa-onnx/csrc/offline-transducer-modified-beam-search-nemo-decoder.cc
@@ -94,7 +94,7 @@ OfflineTransducerModifiedBeamSearchNeMoDecoder::Decode(
 
   int32_t vocab_size = model_->VocabSize();
   int32_t blank_id = vocab_size - 1;  // NeMo models have blank at the end
-  int32_t max_symbols_per_frame = 10;
+  int32_t max_symbols_per_frame = 5;  // Match greedy TDT decoder limit
 
   // For TDT models, we need to know the number of duration bins
   // We'll detect this from the joiner output size on first run
@@ -294,11 +294,6 @@ OfflineTransducerModifiedBeamSearchNeMoDecoder::Decode(
         // Create candidate hypotheses
         for (int32_t idx : top_k_tokens) {
           int32_t token = idx;
-          // For TDT: joint probability = P(token) * P(duration)
-          // In log space: log P(token, duration) = log P(token) + log
-          // P(duration)
-          float token_log_prob =
-              token_logits[token] + duration_log_prob + hyp.log_prob;
 
           NeMoHypothesis new_hyp;
           new_hyp.ys = hyp.ys;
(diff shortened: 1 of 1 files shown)


csukuangfj:

Could you provide a test WAV file that reproduces the bug this PR is intended to fix?

PR #3660, opened 2026-06-04
Support streaming zipformer transducer ASR for QNN.

csukuangfj:
Now both
[pruned_transducer_stateless7_streaming](https://github.com/k2-fsa/icefall/tree/master/egs/librispeech/ASR/pruned_transducer_stateless7_streaming) (zipformer v1)
and
[zipformer](https://github.com/k2-fsa/icefall/tree/master/egs/librispeech/ASR/zipformer) (zipformer v2)
are supported on Qualcomm NPU with QNN.

## TODOs
- [ ] Add CI to upload converted QNN models
- [ ] Add Android demo

Screenshot 2026-06-04 at 18 50 56

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file validation to accept broader path types beyond files
  * Added CJK text processing to remove spaces between Chinese, Japanese, and Korean characters

* **New Features**
  * Enhanced encoder support to handle multiple model configurations with optimized state management

* **Refactor**
  * Reorganized text utility functions for improved code modularity
  * Streamlined tensor casting logic for better performance

### DIFF
--- a/sherpa-onnx/csrc/text-utils.cc
+++ b/sherpa-onnx/csrc/text-utils.cc
@@ -1057,6 +1057,28 @@ bool ContainsCJK(const std::u32string &text) {
   return false;
 }
 
+std::string RemoveSpaceBetweenCjk(const std::string &text) {
+  std::u32string u32 = Utf8ToUtf32(text);
+  if (u32.size() < 3) {
+    return text;
+  }
+
+  std::u32string ans;
+  ans.reserve(u32.size());
+  ans.push_back(u32[0]);
+
+  for (size_t i = 1; i + 1 < u32.size(); ++i) {
+    if (u32[i] == U' ' && IsCJK(u32[i - 1]) && IsCJK(u32[i + 1])) {
+      continue;
+    }
+
+    ans.push_back(u32[i]);
+  }
+
+  ans.push_back(u32.back());
+  return Utf32ToUtf8(ans);
+}
+
 std::string GetWord(const std::vector<std::string> &words, int32_t start,
                     int32_t end) {
   std::string ans;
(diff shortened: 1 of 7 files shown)


sherpa-onnx/csrc/text-utils.cc
```diff
@@ -1057,6 +1057,28 @@ bool ContainsCJK(const std::u32string &text) {
...
+  if (u32.size() < 3) {
+    return text;
+  }
+
+  std::u32string ans;
+  ans.reserve(u32.size());
+  ans.push_back(u32[0]);
+
+  for (size_t i = 1; i + 1 < u32.size(); ++i) {
+    if (u32[i] == U' ' && IsCJK(u32[i - 1]) && IsCJK(u32[i + 1])) {
```
matiaslin, line 1071: Hmm this condition works well to remove a single space between CJK cps, but I believe that it will fail to handle multiple spaces in between CJK chars. Is this the intended behavior?

matiaslin, line 1071: nit: Just note that based on `IsCJK` current implementation, we might not be able to remove the space in the following sceanrio:
```
你好 。
你好 ！
```
Because the punctuation range [`0x3000`; `0x303F`] is not in the `IsCJK` method, if this is intended (i.e. treat the punctuation as not CJK cps), then we should be good.

csukuangfj: Thanks for the review.

The purpose of this function is to handle ASR models that use BPE tokenization for Chinese, where Chinese characters may be prefixed with _. Since the code replaces _ with a space, the output should contain only a single space between adjacent Chinese characters.

csukuangfj: I agree with you.

The current implementation only considers Chinese characters and does not account for punctuation.

PR #3690, opened 2026-06-18
Build sdist in split-package mode so its metadata declares sherpa-onnx-core

### LINKED ISSUE #3689: sdist metadata missing sherpa-onnx-core dependency, causing uv lock installs to fail on macOS
### Summary

When a project depends on `sherpa-onnx==1.13.3`, `uv lock` resolves a lockfile that omits `sherpa-onnx-core`. After `uv sync`, only `sherpa-onnx` is installed, and `import sherpa_onnx` fails on macOS because `libonnxruntime.1.24.4.dylib` is missing.

A direct wheel install works:

```bash
uv pip install sherpa-onnx==1.13.3
```

installs both `sherpa-onnx` and `sherpa-onnx-core`, and the installed wheel metadata reports:

```python
import importlib.metadata as md
print(md.requires("sherpa-onnx"))
# ['sherpa-onnx-core==1.13.3']
```

So the wheel metadata appears correct. The failure seems to come from the source distribution metadata used by project locking: the `uv.lock` entry for `sherpa-onnx` has no dependency on `sherpa-onnx-core`.

### Environment

- macOS arm64
- `uv 0.11.13`
- Python 3.11.15 and Python 3.14.0 both reproduce
- Reproduced with `sherpa-onnx==1.13.3`; also observed with `1.13.2`

### Reproduction

```bash
mkdir /tmp/sherpa-uv-project-repro
cd /tmp/sherpa-uv-project-repro
cat > .python-version <<'EOF_PY'
3.11
EOF_PY
cat > pyproject.toml <<'EOF_TOML'
[project]
name = "sherpa-uv-project-repro"
version = "0.0.0"
requires-python = ">=3.11"
dependencies = [
  "sherpa-onnx==1.13.3",
]
EOF_TOML

UV_PROJECT_ENVIRONMENT=.venv uv lock
rg 'sherpa-onnx-core|requires-dist' uv.lock
UV_PROJECT_ENVIRONMENT=.venv uv sync --no-dev
.venv/bin/python - <<'EOF_PY'
import importlib.metadata as md
print('sherpa', md.version('sherpa-onnx'))
try:
    print('core', md.version('sherpa-onnx-core'))
except Exception as exc:
    print('core missing', type(exc).__name__)
import sherpa_onnx
print('import ok')
EOF_PY
```

### Actual result

The lockfile only contains the project requirement, not the runtime dependency:

```text
requires-dist = [{ name = "sherpa-onnx", specifier = "==1.13.3" }]
```

`uv sync` installs only `sherpa-onnx==1.13.3`, then import fails:

```text
core missing PackageNotFoundError
ImportError: dlopen(.../site-packages/sherpa_onnx/lib/_sherpa_onnx.cpython-311-darwin.so, ...):
  Library not loaded: @rpath/libonnxruntime.1.24.4.dylib
```

### Expected result

The sdist metadata should also declare the matching core package, for example:

```text
Requires-Dist: sherpa-onnx-core==1.13.3
```

Then project lock/sync workflows can install the same runtime dependency that a direct wheel install gets.

### Workaround

Adding the core package explicitly works:

```toml
dependencies = [
  "sherpa-onnx==1.13.3",
  "sherpa-onnx-core==1.13.3",
]
```

After that, `import sherpa_onnx` succeeds and the dylibs are present under `site-packages/sherpa_onnx/lib/`.

Eridanus117:
## Problem

Fixes #3689.

The published wheels are built with `SHERPA_ONNX_SPLIT_PYTHON_PACKAGE=ON` (see all `build-wheels-*.yaml`), so their metadata correctly declares the runtime dependency on `sherpa-onnx-core`:

```python
>>> import importlib.metadata as md; md.requires("sherpa-onnx")
['sherpa-onnx-core==1.13.3']
```

But the **sdist** is built in `build-wheels-linux.yaml` *without* that variable:

```yaml
- name: Build sdist
  ...
  run: |
    python3 setup.py sdist
```

So `need_split_package()` returns `False`, `install_requires` becomes `None`, and the sdist's `PKG-INFO` omits `Requires-Dist: sherpa-onnx-core` entirely. Tools that resolve dependencies from **sdist** metadata (notably `uv lock`) then produce a lockfile without `sherpa-onnx-core`; after `uv sync` only `sherpa-onnx` is installed and `import sherpa_onnx` fails on a missing `libonnxruntime` dylib — even though a direct `uv pip install sherpa-onnx` (which uses the wheel) works.

## Fix

Set `SHERPA_ONNX_SPLIT_PYTHON_PACKAGE=ON` for the sdist build step so the sdist's metadata matches the published wheels. One env var, no other behavior change (`setup.py sdist` does not run the cmake build, so the variable only affects the declared `install_requires`).

## Verification

Built the sdist locally both ways and resolved it with `uv 0.11.13`:

| sdist build | `Requires-Dist` in `PKG-INFO` | `sherpa-onnx-core` in `uv.lock` |
|---|---|---|
| `setup.py sdist` (current) | *(absent)* | ❌ not present |
| `SHERPA_ONNX_SPLIT_PYTHON_PACKAGE=ON setup.py sdist` (this PR) | `sherpa-onnx-core==1.13.3` | ✅ `requires-dist = [{ name = "sherpa-onnx-core", specifier = "==1.13.3" }]` |

With the fix, `uv lock` against the local sdist resolves and locks `sherpa-onnx-core` as expected.

## Summary by CodeRabbit

* **Chores**
  * Updated Linux wheel build configuration to ensure proper dependency declarations in package distributions.

### DIFF
--- a/.github/workflows/build-wheels-linux.yaml
+++ b/.github/workflows/build-wheels-linux.yaml
@@ -425,6 +425,12 @@ jobs:
       - name: Build sdist
         if: matrix.python-version == 'cp38' && matrix.manylinux == 'manylinux2014'
         shell: bash
+        env:
+          # Build the sdist in split-package mode so its metadata declares the
+          # sherpa-onnx-core dependency, matching the published wheels. Without
+          # this, the sdist's Requires-Dist omits sherpa-onnx-core, which breaks
+          # tools that resolve from sdist metadata (e.g. `uv lock`). See #3689.
+          SHERPA_ONNX_SPLIT_PYTHON_PACKAGE: ON
         run: |
           python3 setup.py sdist
           ls -l dist/*


csukuangfj (approved):

Issue #3572, opened 2026-05-03
Support modified_beam_search (and hotwords) for Nemotron streaming transducer


lcrisp:
### Summary

I was benchmarking for a long-form transcription (streaming) project and hit a bit of a dead end... 

`OnlineRecognizer.from_transducer` with `model_type="nemotron"` only supports `decoding_method="greedy_search"`. Passing `modified_beam_search` errors out at:

```
csrc/online-recognizer-transducer-nemo-impl.h:OnlineRecognizerTransducerNeMoImpl:56
Unsupported decoding method: modified_beam_search
```

Because `hotwords_file` and `hotwords_score` are tied to `modified_beam_search`, this sadly leaves the Nemotron streaming transducer with no hotwords / contextual-biasing path. For long-form streaming transcription that includes proper nouns or domain-specific vocabulary (technical terms, place names, person names that aren't in the BPE as units), the model substitutes phonetic neighbours that the hotwords API would otherwise be able to bias correctly.

### Why this matters

`nemotron-speech-streaming-en-0.6b` (PR #3555, 2026-04-25 export) is currently the strongest English streaming transducer in `sherpa-onnx` for low-latency transcription tasks. Tested side-by-side on a 62-second long-form speech sample:

- **Whisper Turbo (offline, beam=5)**: WER 11.6%, RTF 0.140
- **Nemotron jan2026 streaming + tuned endpoints**: WER 12.6%, RTF 0.036

Streaming Nemotron is ~4× faster than offline Whisper-Turbo at an almost-equal WER, with the streaming property that offline Whisper can't provide. The remaining 1-word gap on this sample is a single proper-noun substitution (`re-extract` → `rextract`) that hotwords would have caught. Closing that gap on common domain vocabulary would make the model production-ready for a class of applications (meeting transcription, lecture capture, podcast indexing, dictation) where Whisper is currently chosen primarily for the hotwords path.

### What works today (for reference)

`hotwords_file` works correctly with Zipformer streaming transducers and offline Conformer / Paraformer recognisers when paired with `modified_beam_search`. The C++ implementation `OnlineRecognizerTransducerImpl` (non-Nemo) already supports both decoding methods. The Nemotron variant lives at `online-recognizer-transducer-nemo-impl.h` and explicitly rejects anything other than `greedy_search`.

### Asking for

[...]

Issue #2541, opened 2025-08-29
Why parakeet (transducer model) can't support hotwords?


paulxin001:
Why parakeet can't support hotwords?
I check code in offline-recognizer-transducer-nemo-impl.h 

`class OfflineRecognizerTransducerNeMoImpl : public OfflineRecognizerImpl {
 public:
  explicit OfflineRecognizerTransducerNeMoImpl(
      const OfflineRecognizerConfig &config)
      : OfflineRecognizerImpl(config),
        config_(config),
        symbol_table_(config_.model_config.tokens),
        model_(std::make_unique<OfflineTransducerNeMoModel>(
            config_.model_config)) {
    if (config_.decoding_method == "greedy_search") {
      decoder_ = std::make_unique<OfflineTransducerGreedySearchNeMoDecoder>(
          model_.get(), config_.blank_penalty, model_->IsTDT());
    } else {
      SHERPA_ONNX_LOGE("Unsupported decoding method: %s",
                       config_.decoding_method.c_str());
      SHERPA_ONNX_EXIT(-1);
    }
    PostInit();
  }`

Will support on future version ?

paulxin001:
I understand that the Parakeet model is also a transducer-based decoder, so it should be able to directly reuse the modified_beam_searchmodule, right? Why hasn't this been implemented? I'm considering trying to reuse the modified_beam_searchfrom sherpa-onnx for decoding. Have you attempted any related work? If you have and it failed, please let me know the reasons.

csukuangfj:
> it should be able to directly reuse the modified_beam_searchmodule, right?

You need to adapt the existing modified_beam_search code.

TDT models use a stateful decoder network. All transducer models from icefall use a stateless decoder network.

Shannu1220:
@csukuangfj  Can Whisper models support hotwords?

csukuangfj:
@dosu

nefastosaturo:
I was checking the same for parakeet v3 and I got here and here #2753  . Considering that this framework is awesome I would like to help and give my contribution in some way. About the context biasing for Nemo I get two things. The first is how to recover the bpe.model / bpe.vocab doing this

```python
import nemo.collections.asr as nemo_asr
asr_model = nemo_asr.models.ASRModel.from_pretrained(model_name="nvidia/parakeet-tdt-0.6b-v3")
# Access the SentencePieceProcessor
sp = asr_model.tokenizer.tokenizer

# Save the BPE model
[...]

PR #3723, opened 2026-07-06
Support per-stream hotwords in the JavaScript (node-addon) API for non-streaming ASR

### LINKED ISSUE #2753: Anyway to use hotwords for parakeet v2?
Sorry will try to make the inquiry quick because perhaps it's an obvious no (or yes?)

But I know NVIDIA's framework allows for phrase boosting (an example of some discussion here: https://github.com/NVIDIA-NeMo/NeMo/issues/14500) but beam search doesn't work on sherpa-onnx-nemo-parakeet-tdt-0.6b-v2-fp16 and thus cannot use hotwords (also apparently theirs you can use greedy with keyphrases too, and apparently has optional bpe which I think is what hotwords in sherpa does).

Is there perhaps some alternative way that I may not know about? For my purpose even re-weighting tokens in decoding is helpful. My apologies if I've made a misunderstanding as I'm somewhat still new to all of this and may not understand the correct.

cmurrayis:
Fixes #2753

## What this PR does

The C++ API (`OfflineRecognizer::CreateStream(const std::string &hotwords)`) and the C API (`SherpaOnnxCreateOfflineStreamWithHotwords`) already support **per-stream hotwords** for non-streaming recognition — added for NeMo transducer models in #3077 — but the JavaScript (node-addon) API never exposed them: `createStream()` accepts no arguments and always calls `SherpaOnnxCreateOfflineStream`.

This PR forwards an optional hotwords string through the addon:

```js
const recognizer = new sherpa_onnx.OfflineRecognizer(config);

// unchanged (backward compatible)
const stream1 = recognizer.createStream();

// NEW: per-stream contextual biasing
const stream2 = recognizer.createStream('PHOEBE :2.0/DON QUIXOTE');
```

This makes it possible to bias each utterance towards different context (e.g. per-phone-call speaker names from a CRM) without recreating the recognizer, which is expensive.

## Changes

- `harmony-os/.../cpp/non-streaming-asr.cc` — `CreateOfflineStreamWrapper` accepts an optional second string argument and calls `SherpaOnnxCreateOfflineStreamWithHotwords` (shared with `scripts/node-addon-api/src/` via the existing symlink)
- `scripts/node-addon-api/lib/non-streaming-asr.js` — `createStream(hotwords)` with JSDoc
- `nodejs-addon-examples/test_asr_non_streaming_nemo_parakeet_tdt_v2_hotwords.js` — new example (see below)
- `nodejs-addon-examples/README.md` — table row + run commands
- `.github/scripts/test-nodejs-addon-npm.sh` — runs the new example in the existing parakeet block (model already downloaded there)

## Example output

The parakeet-tdt-0.6b-v2 test wave contains the name "Phoebe", which the model spells "Phebe" by default. With the per-stream hotword, this stream (and only this stream) resolves to the expected spelling:

```
Without hotwords: Well, I don't wish to see it any more, observed Phebe, turning away her eyes. It is certainly very like the old portrait.
With hotwords   : Well, I don't wish to see it any more, observed Phoebe, turning away her eyes. It is certainly very like the old portrait.
```

Since the parakeet release does not ship `bpe.vocab`, the example derives an encoding-equivalent vocab from `tokens.txt` (equal scores → longest-match), so it runs out of the box against the released tarball. (If you would consider adding `bpe.vocab` to the NeMo transducer release tarballs, hotword usage would become simpler — happy to open a separate issue.)

## Notes

- Backward compatible: `createStream()` with no argument is unchanged.
- Hotwords are only supported by transducer models with `decodingMethod: 'modified_beam_search'` (as documented for the C++ API); the JSDoc and the C++ comment say so.
- Tested by building the linux-x64 addon from this branch and running the new example against `sherpa-onnx-nemo-parakeet-tdt-0.6b-v2-int8` (output above). `npm run typecheck` in `scripts/node-addon-api` passes.
- Related: #2541 (transducer hotwords background), #3572 (streaming NeMo hotwords — not addressed here).

## Summary by CodeRabbit

* **New Features**
  * Added optional per-stream hotwords support for offline speech recognition stream creation.
  * Added a new non-streaming ASR example that runs both with no hotwords and with a hotword-biased decode.
  * Updated the examples documentation with the new hotword scenario and run command.
* **Bug Fixes**
  * Improved argument validation and clarified error handling when creating offline recognizers with hotwords.
* **Tests**
  * Extended the Node.js test suite with an additional non-streaming hotwords-based ASR test.

### DIFF
--- a/.github/scripts/test-nodejs-addon-npm.sh
+++ b/.github/scripts/test-nodejs-addon-npm.sh
@@ -170,6 +170,7 @@ tar xvf sherpa-onnx-nemo-parakeet-tdt-0.6b-v2-int8.tar.bz2
 rm sherpa-onnx-nemo-parakeet-tdt-0.6b-v2-int8.tar.bz2
 
 node ./test_asr_non_streaming_nemo_parakeet_tdt_v2.js
+node ./test_asr_non_streaming_nemo_parakeet_tdt_v2_hotwords.js
 rm -rf sherpa-onnx-nemo-parakeet-tdt-0.6b-v2-int8
 
 echo "----------non-streaming ASR dolphin CTC----------"
--- a/harmony-os/SherpaOnnxHar/sherpa_onnx/src/main/cpp/non-streaming-asr.cc
+++ b/harmony-os/SherpaOnnxHar/sherpa_onnx/src/main/cpp/non-streaming-asr.cc
@@ -688,6 +688,15 @@ static Napi::External<SherpaOnnxOfflineStream> CreateOfflineStreamWrapper(
     stream = SherpaOnnxCreateOfflineStream(recognizer);
   }
 
+  if (!stream) {
+    Napi::TypeError::New(env,
+                         "Failed to create offline stream. Please check if "
+                         "your model and decoding method support hotwords.")
+        .ThrowAsJavaScriptException();
+
+    return {};
+  }
+
   return Napi::External<SherpaOnnxOfflineStream>::New(
       env, const_cast<SherpaOnnxOfflineStream *>(stream),
       [](Napi::Env env, SherpaOnnxOfflineStream *stream) {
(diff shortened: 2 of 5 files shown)


cmurrayis:
Thanks for the review — both suggestions applied in 18e452a: added the null-pointer check after stream creation, and simplified the example to call `createStream(hotwords)` directly. Re-ran the example against a build of this branch; output unchanged (Phebe -> Phoebe with the hotword).

csukuangfj (approved):

PR #3750, opened 2026-07-11
Fix dropped diacritics in Supertonic TTS text frontend

bogdanr:
While testing Supertonic 3 with Romanian I noticed that sounds tied to diacritics were silently missing from the synthesized speech: "sinteză" came out as "sintez", "propozițiile" as "propoziiile". The characters were simply gone from the audio.

### Root cause

The unicode indexer was built from NFKD-normalized text, so it only knows base letters and combining marks (e.g. `a` = 60, combining breve U+0306 = 150), while precomposed characters map to -1:

```
ă U+0103 -> -1    ș U+0219 -> -1    ț U+021B -> -1
č U+010D -> -1    ą U+0105 -> -1    ő U+0151 -> -1    ế U+1EBF -> -1
```

The reference implementation handles this by running `unicodedata.normalize("NFKD", text)` before indexing ([py/helper.py](https://github.com/supertone-inc/supertonic/blob/main/py/helper.py)). The C++ port instead has a hand-written `DecomposeLatin()` switch that only covers Latin-1 precomposed characters (á, é, ü, ç, ...). Everything from Latin Extended-A/B and Latin Extended Additional falls through, gets looked up as-is, hits -1 and is dropped. This affects Romanian, Czech, Slovak, Polish, Hungarian, Turkish, Croatian, Latvian, Lithuanian, Vietnamese, ...  most of the non-English Latin-script languages Supertonic supports.

### Fix

Replace the partial switch with a generated NFKD lookup table for the whole BMP (Hangul syllables keep the existing algorithmic path). The table is generated by the new `scripts/supertonic/generate_nfkd_table.py` from Python's own unicodedata, so it can be regenerated easily. Lookup is a binary search over a sorted array, no allocations.

As a bonus this also brings the compatibility decompositions in line with the reference implementation (fullwidth forms, ligatures, etc.).

### Testing

- Added `offline-tts-supertonic-unicode-processor-test.cc` (gtest) covering ASCII, Latin-1, Romanian/Czech/Polish/Hungarian/Turkish/Vietnamese diacritics, Hangul, compatibility forms and non-BMP input. All pass.
- End-to-end with `sherpa-onnx-supertonic-3-tts-int8-2026-05-11`, `--lang=ro`:

  `"Vreau să aud cât de natural sună sinteza și pronunția propozițiilor: ă, â, î, ș, ț."`

  Before: 7.72s of audio with every ă/ș/ț sound missing. After: 8.49s and all syllables are rendered. Korean output is unchanged.

## Summary by CodeRabbit

* **Enhancements**
  * Improved Supertonic offline TTS Unicode processing with broader NFKD decomposition support.
  * Added handling for additional accented characters, compatibility forms, and Hangul syllables.
  * Non-BMP decompositions are safely excluded to maintain supported output compatibility.

* **Tests**
  * Added comprehensive coverage for Unicode decomposition, pass-through behavior, and edge cases.


csukuangfj (approved):

bogdanr:
> Thank you for your contribution!

You're welcome! I'm also enjoying using your stuff 👍🏼 
I see you have a showcase section so I added my project there in this PR: github.com/k2-fsa/sherpa-onnx/pull/3775

Issue #2468, opened 2025-08-08
Any plan to integrate source separation in to flutter/android/ios


hokamc:
Are there any planned integrations for audio source separation in Flutter, Android, and iOS platforms?

csukuangfj:
We have already provided C++ API. Would you like to bind it?

We can help you if you have any issues.

luckysmg:
@csukuangfj 作者你好，但是我看到在的`c-api.h`,`cxx-api.h` 中并没有找到相关 api(我理解上层应该用的是这个吧^_^), 但我在`csrc`里面才找到了提到的`OfflineSourceSeparationConfig`API，但在编译脚本里面貌似没有把这部分放进去(我用的`build-ios-shared.sh`)，这个是操作不对吗

PR #3765, opened 2026-07-17
Support exporting Spleeter 4stems (vocals/drums/bass/other)

bp-brendan:
`scripts/spleeter/` could only export 2stems, and the README/docs say so: *"We only support the `2-stem` model at present."* This adds 4stems.

Extending it hinged on one thing that isn't visible in the code, so leading with that.

## 4stems uses ELU

Spleeter selects the U-Net activations from config:

```jsonc
// configs/2stems/base_config.json
"model": { "type": "unet.unet", "params": {} }
//                              ^^ defaults: LeakyReLU(0.2) + ReLU

// configs/4stems/base_config.json
"model": { "type": "unet.unet", "params": {
    "conv_activation": "ELU", "deconv_activation": "ELU" } }
```

The architecture and **every weight shape are identical** between the two. So `unet.py`'s hardcoded `LeakyReLU`/`ReLU` loads the 4stems weights via `load_state_dict` **without complaint** and returns garbage. It presented as a max error of **945 against a TF output whose entire range was 178** — the error larger than the signal — with the op-name mapping, the input tensor, the inference-mode BatchNorm and the kernel shapes all checking out clean, because the bug was in none of them.

`UNet` now takes `conv_activation` / `deconv_activation`, **defaulting to the current values**, so 2stems is untouched.

Incidentally: your existing TF-vs-torch assertion in `convert_to_torch.py` is what made this findable instead of silent. Without it this ships wrong and you learn about it by ear, much later. It's a good check and I kept it as the proof below.

## Changes

- **`unet.py`** — `UNet(conv_activation="LeakyReLU", deconv_activation="ReLU")`. Defaults preserve current behaviour.
- **`convert_to_torch.py`** — `--model {2stems,4stems}`, and the per-stem op-name offsets are derived rather than enumerated. They were already regular: each stem's U-Net owns a contiguous block in `instrument_list` order — 7 `conv2d`, 6 `conv2d_transpose`, 12 `batch_normalization` indices. Verified by reproducing the existing 2stems mapping exactly. The 12-vs-11 BatchNorm gap is preserved and now commented (`conv5` has no BatchNorm but the index is still allocated, which is why `bn5` reads `batch_normalization_6`, not `_5`).
- **`export_onnx.py`** — `--model`, exports that model's stems, and writes `conv_activation`/`deconv_activation` into the ONNX metadata so the above doesn't have to be rediscovered.
- **`run.sh`** — `./run.sh [2stems|4stems]`, default `2stems`.
- **`.gitignore`** — `4stems`, `4stems.tar.gz`.
- Drive-by: `add_meta_data` said `sample_rate: 41000`. Spleeter is 44100.

Net −44 lines.

## Verification

Your assertion, every stem:

| | max err |
|---|---|
| `2stems/vocals` | 3.04e-04 *(unchanged)* |
| `2stems/accompaniment` | 1.07e-02 *(unchanged)* |
| `4stems/vocals` | 2.59e-04 |
| `4stems/drums` | 1.33e-02 |
| `4stems/bass` | 1.12e-03 |
| `4stems/other` | 2.59e-03 |

End-to-end, the exported 4stems graphs reconstruct the mix to **−153.1 dB** (the ratio masks sum to 1, so the stems must sum back to the input — that invariant tests the STFT, the mask, the band extension and the iSTFT at once). Inference is ~1.69 s for 30 s of 44.1 kHz stereo, all four stems, CPU.

```bash
cd scripts/spleeter && ./run.sh 4stems
```

## Notes

- **5stems is out of scope, and not for the reason I first assumed.** I originally wrote that it was "likely just another entry in `STEMS`/`ACTIVATIONS`". That's wrong — it uses `unet.softmax_unet`, not `unet.unet`. See the comment below for what it would actually take; it needs a different export shape, not a config entry.
- Possibly relevant to #2468 (source separation on Flutter/Android/iOS) — that's what I was doing when I hit this.
- Exported models and the full pipeline, if useful: https://huggingface.co/Best-Practice/spleeter-4stems-onnx · https://github.com/madewith-bestpractice/spleeter-4stems-onnx

## Summary by CodeRabbit

* **New Features**
  * Added parameterized support for both **2-stem** and **4-stem** model conversions and exports.
  * Introduced **configurable activation functions** in the separation network to match model variants.
  * Enhanced **ONNX exports** with model-specific metadata and variant-aware naming.
* **Bug Fixes**
  * Improved validation for selected stem names against the chosen model variant.
  * Added output consistency checks during the conversion process.
* **Chores**
  * Updated ignore rules for additional generated stem archive paths.

### DIFF
--- a/scripts/spleeter/.gitignore
+++ b/scripts/spleeter/.gitignore
@@ -1,2 +1,4 @@
 2stems.tar.gz
 2stems
+4stems.tar.gz
+4stems
--- a/scripts/spleeter/convert_to_torch.py
+++ b/scripts/spleeter/convert_to_torch.py
@@ -11,6 +11,27 @@ import torch
 
 from unet import UNet
 
+# Spleeter's instrument_list, in the order configs/<model>/base_config.json
+# declares it. That order is what fixes each stem's block of TF op names.
+STEMS = {
+    "2stems": ["vocals", "accompaniment"],
+    "4stems": ["vocals", "drums", "bass", "other"],
+}
+
+# The U-Net activations, also from base_config.json. 2stems ships "params": {},
+# i.e. the unet.unet defaults; 4stems asks for ELU on both. Architecture and
+# weight shapes are identical either way, so getting this wrong loads cleanly
+# and returns garbage rather than raising.
+ACTIVATIONS = {
+    "2stems": ("LeakyReLU", "ReLU"),
+    "4stems": ("ELU", "ELU"),
+}
+
+
+def op(family, index):
+    """TF names the first op of a family with no suffix: conv2d, conv2d_1, ..."""
+    return family if index == 0 else f"{family}_{index}"
+
 
 def load_graph(frozen_graph_filename):
     # This function is modified from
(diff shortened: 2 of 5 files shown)


bp-brendan:
Thanks — both points addressed in cb1e047.

**`export_onnx.py` shadowing** — correct, `model` was shadowed by `model = onnx.load(filename)` on the next line. Renamed to `model_id`.

**`unet.py` activations** — took the suggestion's shape (resolve once in `__init__`), though for a different reason than the one given. The inference overhead isn't real: it's eleven string comparisons against a ~10 GFLOP forward pass, so nanoseconds against ~400 ms.

But `__init__` is the right place regardless, because it's where validation belongs — and validation turned out to matter here. **Both my original and the suggested replacement fall through to a default on an unknown name**, so `UNet(conv_activation="elu")` silently gives you LeakyReLU. That's the exact failure this parameter exists to prevent: 2stems and 4stems have identical architectures and identical weight shapes, so a wrong activation loads without complaint and returns garbage. A silent default would reintroduce it. Unknown names now raise:

```python
>>> UNet("elu", "ELU")
ValueError: unknown activation 'elu'; expected one of ['ELU', 'LeakyReLU', 'ReLU']
```

Re-verified after the refactor — unchanged, including the 2stems regression:

| | max err |
|---|---|
| `2stems/vocals` | 3.04e-04 |
| `2stems/accompaniment` | 1.07e-02 |
| `4stems/vocals` | 2.59e-04 |
| `4stems/drums` | 1.33e-02 |
| `4stems/bass` | 1.12e-03 |
| `4stems/other` | 2.59e-03 |

And the exported 4stems graph contains **11 `Elu` and zero `LeakyRelu`/`Relu`**, so the activation survives tracing rather than being quietly defaulted.

---

**One unrelated thing, not touched in this PR.** `export_onnx.py` does:

```python
from onnxmltools.utils.float16_converter import convert_float_to_float16
```

which raises `ModuleNotFoundError` on current `onnxmltools` — the helper now lives in `onnxconverter_common.float16`. This affects 2stems identically, so it's pre-existing rather than a regression here, and I shimmed it locally to test the export path. I've left it alone since I can't tell whether you pin an older `onnxmltools`; happy to fix it here or separately if that'd help.

bp-brendan:
**Correction on 5stems.** I originally wrote that it was "likely just another entry in `STEMS`/`ACTIVATIONS`". I've now checked, and that's wrong. I've corrected the PR description. Posting the detail because it's the answer to "why not 5stems too", and it's not obvious.

**5stems is a different model function:**

```jsonc
// configs/4stems/base_config.json
"model": { "type": "unet.unet",         "params": {"conv_activation": "ELU", "deconv_activation": "ELU"} }

// configs/5stems/base_config.json
"model": { "type": "unet.softmax_unet", "params": {"conv_activation": "ELU", "deconv_activation": "ELU"} }
```

`softmax_unet` calls `apply_unet(..., output_mask_logit=True)` per instrument — so each U-Net emits **logits**, with no sigmoid and no multiply — then applies `Softmax(axis=4)` across the **stacked stems**, in-graph, and multiplies after.

That breaks the per-stem export this design relies on. Measured, freezing a single stem:

| | 4stems (`unet`) | 5stems (`softmax_unet`) |
|---|---|---|
| frozen graph, one stem | 39.4 MB | **196.9 MB** |
| `conv2d` | 7 | **35** (5 × 7) |
| `conv2d_transpose` | 6 | **30** (5 × 6) |
| `batch_normalization` | 11 | **55** (5 × 11) |
| `Sigmoid` | 1 | **0** |
| `Softmax` | 0 | **1** |

Asking for `piano_spectrogram/mul` drags in all five U-Nets, because the softmax needs every stem's logits. The instrument order differs too — `["vocals", "piano", "drums", "bass", "other"]`, so piano is index **1**, not 4. (The BatchNorm indices skip 5, 17, 29, 41, 53 — i.e. `12k+5` — so the 12-allocated/11-used block pattern does hold across all five.)

**What 5stems would need**, if you want it later:

1. A logit-output mode on `UNet` (skip the final `sigmoid` and the multiply).
2. Freezing each stem's final conv (`conv2d_{7k+6}/BiasAdd`) rather than `{stem}_spectrogram/mul`, to get one U-Net per graph instead of five.
3. Softmax-across-stems + multiply done outside the graph, replacing the ratio mask.

That's the same overall shape as today — N per-stem ONNX files, mask maths outside — so it's tractable, but it's a real port rather than a config entry. Happy to do it as a separate PR if it's wanted; I've left it out of this one to keep the diff to what I've actually verified.

PR #3766, opened 2026-07-18
Add Japanese support for Kokoro multi-lang TTS

doracpphp:
## What does this PR do?

This PR adds Japanese TTS support to the Kokoro multi-lang models
(`kokoro-multi-lang-v1_0`), following the same lexicon-based design as the
existing Chinese support.

The model itself is already Japanese-capable: `voices.bin` contains the 5
Japanese speakers (`jf_alpha`, `jf_gongitsune`, `jf_nezumi`, `jf_tebukuro`,
`jm_kumo`, sid 37-41) and `tokens.txt` covers the Japanese phonemes. What was
missing is the text frontend: Japanese runs (kana + kanji) could not be
converted to token IDs.

## Changes

1. **`scripts/kokoro/v1.0/generate_lexicon_ja.py`** (new)
   Generates `lexicon-ja.txt` offline using misaki's *cutlet-version* Japanese
   G2P (this is the phoneme format v1.0 was trained with; the newer
   pyopenjtalk-based G2P of misaki emits pitch-string characters like `_ ^ -`
   that are not in v1.0 `tokens.txt`). The lexicon contains:
   - the ~100k most frequent Japanese words (wordfreq)
   - word + particle compounds (`word+は`, `word+へ`), so that greedy
     longest-match resolves the context-dependent particle readings
     (は=wa, へ=e) and wins over collisions such as 今日|はい|い
   - every single kana and small-kana combination as a fallback, so any
     kana-only OOV text is always synthesizable
   - 127,162 entries, ~2.3 MB (similar to `lexicon-zh.txt`)

2. **`sherpa-onnx/csrc/kokoro-multi-lang-lexicon.cc`**
   - When `--kokoro-lang=ja` is given, kana/kanji runs are routed to the
     lexicon phrase matcher. The default (no `lang`) is unchanged, so the
     existing Chinese/English behavior is fully backward compatible.
   - A space token is inserted between matched Japanese words, which keeps
     the prosody close to the reference sentence-level misaki G2P output.
   - Non-CJK runs (e.g. embedded English words) use the word lexicon +
     espeak-ng fallback instead of an espeak "ja" voice.
   - Japanese punctuation (「」『』・ and full-width space) is normalized.

## Usage

```bash
sherpa-onnx-offline-tts \
  --kokoro-model=./kokoro-multi-lang-v1_0/model.onnx \
  --kokoro-voices=./kokoro-multi-lang-v1_0/voices.bin \
  --kokoro-tokens=./kokoro-multi-lang-v1_0/tokens.txt \
  --kokoro-data-dir=./kokoro-multi-lang-v1_0/espeak-ng-data \
  --kokoro-lexicon=./lexicon-ja.txt,./kokoro-multi-lang-v1_0/lexicon-us-en.txt \
  --kokoro-lang=ja \
  --sid=38 \
  --output-filename=./ja.wav \
  "今日はいい天気ですね。映画を見に行きませんか。"
```

## How this was tested

- Verified via the C API (Windows x64, CPU) with
  `kokoro-multi-lang-v1_0`: Japanese runs are routed to the lexicon,
  embedded English words (e.g. "hobby") are routed to the English lexicon,
  and punctuation is normalized as expected (checked with `--debug`).
- Round-trip check with Whisper (ja): the synthesized audio for test
  sentences (greetings, weather, mixed Japanese/English) is transcribed back
  to the original text.
- Compared the lexicon-based phonemization against the reference
  sentence-level misaki G2P on a test sentence set: with the word+particle
  compound entries, the outputs match on all test sentences.

## Known limitations

- A word-level lexicon cannot fully disambiguate context-dependent readings
  (e.g. 今日 = きょう/こんにち, 行った = いった/おこなった). This is the
  same trade-off as the existing Chinese support (heteronyms).
- Numbers and symbols are not yet converted to their Japanese readings
  (no num2kana equivalent). This could be a follow-up (e.g. an FST like
  `number-zh.fst`).
- v1.0 was trained without pitch-accent information (cutlet phonemes), so
  pitch accent is up to the model. If a future Kokoro release is trained
  with misaki's newer pitch-aware Japanese format, only the lexicon needs
  to be regenerated.

## Note

The generated `lexicon-ja.txt` (127,162 entries, ~2.3 MB) is intentionally
not included in this PR, since lexicons are distributed inside the model
packages (like `lexicon-zh.txt` in `kokoro-multi-lang-v1_0.tar.bz2`) rather
than in the repository. I can provide the generated file for inclusion in
the kokoro-multi-lang model packages, or it can be regenerated with the
script added in this PR.

## Summary by CodeRabbit

* **New Features**
  * Added a standalone Japanese lexicon generator that produces a new `lexicon-ja.txt`, including frequent-token coverage, particle compounds, full kana handling, and manual pronunciation overrides.
  * Added optional spacing insertion in generated Japanese token IDs to improve text rendering/shaping.
* **Bug Fixes**
  * Improved Japanese vs non-Japanese tokenization by normalizing quotes/punctuation/whitespace and using case-insensitive language detection for safer routing and more accurate matching.

### DIFF
--- a/scripts/kokoro/v1.0/generate_lexicon_ja.py
+++ b/scripts/kokoro/v1.0/generate_lexicon_ja.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python3
+# Copyright    2026  (author: dora)
+"""
+Generate lexicon-ja.txt for kokoro-multi-lang-v1_0.
+
+Requirements:
+    pip install "misaki[ja]" wordfreq
+    python -m unidic download
+
+Note: kokoro-multi-lang-v1_0 was trained with phonemes produced by the
+*cutlet* version of misaki's Japanese G2P (e.g. こんにちは -> koɲɲiʨiβa).
+The newer pyopenjtalk-based G2P of misaki emits a pitch string
+(characters like _ ^ -) and extra phonemes (e.g. ᶄ) that are NOT in
+tokens.txt of v1.0, so we must use JAG2P(version='cutlet') here.
+
+The lexicon contains:
+  - the most frequent Japanese words (wordfreq) and their phonemes
+  - every single kana (hiragana + katakana) and small-kana combination,
+    so that any kana-only OOV text can still be synthesized
+  - explicit entries for particles such as は/へ, mapped to their
+    particle reading (wa/e). Longest-match lookup consumes real words
+    first, so a remaining single は is most likely a topic particle.
+"""
+
+from typing import Dict, List
+
+from misaki import ja
+from wordfreq import top_n_list
+
+NUM_WORDS = 100000
+NUM_PARTICLE_WORDS = 20000
+
+# Explicit overrides. Applied last, they win over generated entries.
+MANUAL = {
+    "は": "βa",  # topic particle
+    "へ": "e",  # direction particle
+    "を": "o",
+    "っ": "ʔ",
+    "ッ": "ʔ",
+    "ー": "ː",
+    "ん": "ɴ",
+    "ン": "ɴ",
+}
+
+HIRAGANA_SMALL = "ゃゅょぁぃぅぇぉゎ"
+KATAKANA_SMALL = "ャュョァィゥェォヮ"
+
+
+def is_kana(ch: str) -> bool:
+    return 0x3040 <= ord(ch) <= 0x309F or 0x30A0 <= ord(ch) <= 0x30FF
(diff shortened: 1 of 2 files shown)


csukuangfj:

Could you please upload a few sample generated audio files along with their corresponding transcripts/text so we can easily test and verify the output quality?

Issue #2355, opened 2025-07-07
inference is low with directml


heibaidaolx123:
Hi, I built sherpa-onnx with directml on a Thinkpad X1 Carbon(Intel 11th-Gen Core  i7-1165G7 CPU and Intel Iris Xe Graphics).
When testing speaker diarization, the RTF for CPU EP is 0.2, and 10+ for directml EP, which was unexpected.

The command to build:
```
cmake -DSHERPA_ONNX_ENABLE_DIRECTML=ON ../
cmake --build . --target sherpa-onnx-speaker-diarization --config Release
```

The test command:
```
.\sherpa-onnx-offline-speaker-diarization.exe --clustering.num-clusters=4 --segmentation.pyannote-model=D:\download\models\sherpa-onnx-pyannote-segmentation-3-0\model.onnx --embedding.model=D:\download\models\3dspeaker_speech_eres2net_base_sv_zh-cn_3dspeaker_16k.onnx --segmentation.provider=directml --embedding.provider=directml D:\download\models\0-four-speakers-zh.wav
```

What I tried:
- when testing, set num-threads to higher integer:  no improvement ❌
- when building, set SHERPA_ONNX_LINK_D3d:  no improvement ❌

Note
- I changed microsoft.ml.onnxruntime.directml.1.14.1.nupkg to 1.17.1, otherwise it would crash with an 'opset unsupported' error.
- I checked the DirectX conf with `dxdiag`, it seems normal, as shown in the fig below

<img width="800"  alt="Image" src="https://github.com/user-attachments/assets/c6cfae2f-fe23-410f-b866-ec82b4658f60" />

Any clue? What did i miss?
Thanks in advance.

heibaidaolx123:
I tested building with microsoft.ml.onnxruntime.directml.1.17.1 + Microsoft.AI.DirectML.1.15.0, 1.17.1 + 1.15.1, and 1.20.0 + 1.15.4.
No improvements, either.

heibaidaolx123:
I tried updating the graphic driver to the latest version, rebuilt, and reran. No improvement.

csukuangfj:
Do you have any example model.that can be accelerated with directml using onnxruntime?

heibaidaolx123:
@csukuangfj It's on my todo list. Will report after testing

csukuangfj:
Pay attention that not all models can be accelerated by directml.

directml only supports a subset of operators from onnxruntime.

[...]

PR #3768, opened 2026-07-19
Print per-stage timing for offline speaker diarization in debug mode

JulianPscheid:
When `--segmentation.debug=true` is set, the offline pyannote diarization pipeline now logs per-stage wall-clock timings at the end of `Process()`:

```
OfflineSpeakerDiarization: segmentation 40.712 s
OfflineSpeakerDiarization: embedding 5.438 s
OfflineSpeakerDiarization: clustering 0.012 s
OfflineSpeakerDiarization: total 46.315 s, audio 180.000 s, RTF 0.257
```

Motivation: offline diarization is often the dominant cost in end-user pipelines (on mobile CPUs we see RTF 0.25+), and today there is no way to see how the time splits between the segmentation model, embedding extraction, and clustering without patching the source. Having the split visible makes it much easier to diagnose provider/backend regressions (e.g. the kind of slowdown reported in #2355) and to know which stage is worth optimizing on a given device.

Notes:

- Output is gated behind the existing `config_.segmentation.debug` flag; no new config field or API surface.
- The three stage lines bracket only the segmentation forward loop, `ComputeEmbeddings`, and `Cluster`; the total also includes label conversion, stitching, and relabeling, so the stages will not sum exactly to the total.
- The one-chunk shortcut path logs only segmentation and total (embedding/clustering do not run there).
- Early exits that occur after real work also log the stages that ran (segmentation only for the no-speakers exit; all stages for the empty-clustering exit) — those are exactly the runs where the split helps diagnose a regression. The empty-segmentations exit stays silent since it can be reached with `n <= 0`.
- Timings use `std::chrono::steady_clock`; the non-debug path only gains a few monotonic clock reads.
- The total is captured before the log lines are emitted, so logging I/O does not inflate it.

## Summary by CodeRabbit

* **Performance Monitoring**
  * Added debug-controlled runtime timing logs for speaker diarization, covering segmentation, embedding computation, clustering, and overall processing duration.
  * When diagnostic mode is enabled, logs also report estimated audio duration and real-time factor (RTF) to improve end-to-end performance visibility.
  * In cases where diarization produces no output, concise total/RTF logging is still emitted for easier troubleshooting.

### DIFF
--- a/sherpa-onnx/csrc/offline-speaker-diarization-pyannote-impl.h
+++ b/sherpa-onnx/csrc/offline-speaker-diarization-pyannote-impl.h
@@ -8,6 +8,7 @@
 #include <chrono>
 #include <cmath>
 #include <memory>
+#include <optional>
 #include <unordered_map>
 #include <utility>
 #include <vector>
@@ -104,6 +105,7 @@ class OfflineSpeakerDiarizationPyannoteImpl
     // segmentations[i] is for chunk_i
     // Each matrix is of shape (num_frames, num_powerset_classes)
     if (segmentations.empty()) {
+      // Timings are intentionally not logged since n can be non-positive.
       return {};
     }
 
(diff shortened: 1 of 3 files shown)


csukuangfj (changes requested):
Can you create a timer class that takes an argument `bool debug`.
- When debug is false, all of it member methods does nothing.
- It starts counting on creation if debug is true
- It has a method, `Log()`, that prints the elapsed time sinces it creation at the point of calling this function when debug is true.

You can creat two files, `timer.h` and `timer.cc`.

JulianPscheid:
Done in 8dc4dd75. That did shrink the code, thanks.

`sherpa-onnx/csrc/timer.h` and `timer.cc` already existed, so I extended that
`Timer` instead of adding new files:

```c++
explicit Timer(bool debug = true);
void Reset() const;
double Elapsed() const;
void Log(const char *tag) const;
```

With `debug` false it never reads the clock at construction, `Reset()` and
`Log()` do nothing, and `Elapsed()` returns 0.

I defaulted `debug` to true because `sherpa-onnx.cc:106` default-constructs a
`Timer` and always prints `Elapsed()`. A false default would quietly turn that
into `0.00000`. If you'd rather the default be false, I can update that call
site instead.

Each stage now logs itself right after the work it measures, which let me drop
the `std::optional` stage durations and the duplicated log block. Only the
combined line is still built by hand, since it carries the audio duration and
RTF as well. Floats use `%{public}` under `__OHOS__`.

The segmentation `Log()` sits after the `segmentations.empty()` early return so
that path stays silent. `n` can be non-positive there, which would make the RTF
divisor zero.

Output on a 25 s clip:

```
OfflineSpeakerDiarization: segmentation 0.637 s
OfflineSpeakerDiarization: embedding 1.600 s
OfflineSpeakerDiarization: clustering 0.000 s
OfflineSpeakerDiarization: total 2.238 s, audio 25.000 s, RTF 0.090
```

With `--segmentation.debug=0` there is no timing output at all.

csukuangfj (approved):

PR #3781, opened 2026-07-22
Fix use-after-free in async TTS progress callbacks

### LINKED ISSUE #3780: [BUG] Node.js async TTS progress callback causes a use-after-free crash
### Important Instructions ⚠️

Before submitting this issue, please **read carefully** and confirm the following:

- [X] I am using the **latest version** of `sherpa-onnx`. If not, I will upgrade and check if the issue persists.
- [X] I understand that **issues not reproducible on the latest version** may be closed without response.
- [X] I have verified that the error is reproducible using **only `sherpa-onnx`**, not code from my own project.
- [X] I will provide **detailed steps** to reproduce the issue.
- [X] I will provide **complete logs** and error messages.
- [X] I understand that **issues not following these instructions** may be closed or receive no response (如果直接删掉这个issue template 里的步骤，我们很有可能**不会回复**您的issue).

---

### Describe the issue

`OfflineTts.generateAsync({..., onProgress})` in the Node.js addon can deterministically abort the process during multi-chunk generation once progress callbacks are being delivered. Deterministic — not a slow leak.

Root cause (verified in source): in `harmony-os/SherpaOnnxHar/sherpa_onnx/src/main/cpp/non-streaming-tts.cc` (the Node addon uses this file via symlink), the TSFN queue holds raw `TtsCallbackData*` owned by the AsyncWorker's `data_list_`, and nothing orders TSFN queue drain before `~TtsGenerateWorker` frees every entry. When the destructor wins the race, `InvokeJsCallback` reads a freed `samples.size()` and the corrupted size aborts inside `ArrayBuffer::New`. The mutex added in PR #3419 protects the list operations but not this lifetime ordering (the destructor takes no lock and the queue keeps dangling pointers).

Two notes for triage:
- Short texts that produce a single batch may never invoke the progress callback, which is why the one-shot `*_play_async.js` examples don't hit this.
- This is distinct from #1939 — that is ONNX Runtime memory-pattern/arena growth affecting sync and async alike; this is a use-after-free specific to the addon's callback machinery.

I have a fix validated locally (single-owner RAII chunks, bounded queue with backpressure, `napi_closing` handling, drain-before-settle) — PR incoming.

### Steps to reproduce

1. `npm install sherpa-onnx-node@1.13.4` and download `vits-piper-en_GB-cori-medium` from the tts-models release.
2. Save as `repro.js`:

```js
const sherpa = require('sherpa-onnx-node');
const tts = new sherpa.OfflineTts({
  model: {
    vits: {
      model: './vits-piper-en_GB-cori-medium/en_GB-cori-medium.onnx',
      tokens: './vits-piper-en_GB-cori-medium/tokens.txt',
      dataDir: './vits-piper-en_GB-cori-medium/espeak-ng-data',
    },
    numThreads: 2, provider: 'cpu', debug: false,
  },
  maxNumSentences: 1,
});
(async () => {
  const r = await tts.generateAsync({
    text: 'Hello there, how are you doing on this fine morning today?',
    generationConfig: {sid: 0, speed: 1.0},
    onProgress: (p) => { console.log('progress', p.progress); return 1; },
  });
  console.log('done', r.samples.length); // never reached — process aborts
})();
```

3. `node repro.js`

The text must be long enough for the progress callback to fire (short texts = single batch = no callback = no crash).

### Expected behavior

`onProgress` is invoked with each audio chunk and the promise resolves with the full audio — as the `test_tts_non_streaming_vits_piper_en_play_async.js` example implies.

### Environment
- `sherpa-onnx` version: sherpa-onnx-node 1.13.4 (npm, latest)
- OS: macOS 15 (arm64); also reproduced on Linux arm64 (Ubuntu/JetPack)
- Python version: N/A — Node.js addon (reproduced on Node 22.17.0 and 24.4.0)
- onnxruntime version: 1.27.0 (the libonnxruntime bundled with sherpa-onnx-darwin-arm64 / sherpa-onnx-linux-arm64)
- GPU (if applicable): N/A (provider: cpu)
- NPU (if applicable): N/A

### Logs
```
progress 0.5
FATAL ERROR: v8::ArrayBuffer::New Allocation failed - process out of memory
----- Native stack trace -----

 1: 0x104ebff4c node::OOMErrorHandler(char const*, v8::OOMDetails const&) [node]
 2: 0x10509e418 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, v8::OOMDetails const&) [node]
 3: 0x1050c3570 v8::ArrayBuffer::New(v8::Isolate*, unsigned long, v8::BackingStoreInitializationMode) [node]
 4: 0x104e67800 napi_create_arraybuffer [node]
 5: 0x127e21f00 Napi::TypedThreadSafeFunction<Napi::Reference<Napi::Value>, TtsCallbackData, &InvokeJsCallback(Napi::Env, Napi::Function, Napi::Reference<Napi::Value>*, TtsCallbackData*)>::CallJsInternal(napi_env__*, napi_value__*, void*, void*) [sherpa-onnx.node]
 6: 0x104e81e28 v8impl::(anonymous namespace)::ThreadSafeFunction::AsyncCb(uv_async_s*) [node]
 7: 0x105ad56b8 uv__async_io [node]
 8: 0x105ae9a50 uv__io_poll [node]
 9: 0x105ad5c20 uv_run [node]
10: 0x104d99a98 node::SpinEventLoopInternal(node::Environment*) [node]
11: 0x104eec1a0 node::NodeMainInstance::Run() [node]
12: 0x104e5a538 node::Start(int, char**) [node]
13: 0x18cbcfe00 start [/usr/lib/dyld]
```

kecoco16:
Fixes #3780

## Problem

`OfflineTts.generateAsync()` with an `onProgress` callback can deterministically abort the process during multi-chunk generation (full stack and minimal reproduction in #3780). Root cause: the TSFN queue holds raw pointers to `TtsCallbackData` objects owned by the AsyncWorker's `data_list_`, and nothing orders queue drain before the worker destructor frees them.

## Changes

- Give each callback chunk single-owner RAII cleanup.
- Remove the global callback mutex and manual pending-data lists.
- Use per-generation cancellation state.
- Add a bounded TSFN queue with backpressure (the producer runs on a worker thread, so a full queue blocks synthesis, never the event loop).
- Handle `napi_closing` without accessing the TSFN again.
- Drain queued callbacks before settling the generation promise — signaled by a FIFO done-sentinel processed on the main thread; TSFN finalizers never run JS (unsafe during environment teardown).
- Reject the promise when `onProgress` throws.
- Preserve best-effort cancellation semantics for already-generated chunks.
- Cover both the legacy and `generationConfig` async workers.
- Cover external and copied output buffers.

The file is shared with the HarmonyOS binding via symlink, so both platforms get the fix.

## Behavior change

A throwing `onProgress` previously invoked undefined behavior (the exception unwound through the N-API boundary). It now cancels the generation (best-effort) and the promise rejects with an error containing the thrown callback message. Exceptions never cross the N-API boundary, in both node-addon-api exception modes.

## Testing

Added `test_tts_async_callback_stress.js`, covering:

- Repeated async generation.
- Both async worker paths.
- Cancellation from `onProgress` (one callback total; no callbacks after cancellation).
- Throwing progress callbacks (promise rejects with an error containing the thrown callback message; nothing reaches `uncaughtException`).
- Concurrent generations with isolated cancellation.
- `enableExternalBuffer: false`.
- Steady-state RSS reporting.

The test is included in the existing Node.js addon CI script using the Piper English model already downloaded by that workflow.

Tested locally with:
```bash
node test_tts_async_callback_stress.js ./vits-piper-en_GB-cori-medium 30
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of async offline text-to-speech completion.
  - Enhanced cancellation and error handling so progress updates and final promise resolution behave consistently.
  - Improved stability when running multiple TTS generations concurrently, without disrupting progress callbacks.

- **Tests**
  - Added/expanded a stress test covering repeated, concurrent, cancelled, and throwing async TTS progress-callback scenarios, plus audio/buffering behavior checks.
  - Extended CI coverage by running the async callback stress test for an additional TTS model.

### DIFF
--- a/.github/scripts/test-nodejs-addon-npm.sh
+++ b/.github/scripts/test-nodejs-addon-npm.sh
@@ -340,6 +340,7 @@ rm vits-piper-en_GB-cori-medium.tar.bz2
 
 node ./test_tts_non_streaming_vits_piper_en.js
 node ./test_tts_non_streaming_vits_piper_en_async.js
+node ./test_tts_async_callback_stress.js ./vits-piper-en_GB-cori-medium 30
 rm -rf vits-piper-en_GB-cori-medium
 
 curl -SL -O https://github.com/k2-fsa/sherpa-onnx/releases/download/tts-models/vits-coqui-de-css10.tar.bz2
--- a/harmony-os/SherpaOnnxHar/sherpa_onnx/src/main/cpp/non-streaming-tts.cc
+++ b/harmony-os/SherpaOnnxHar/sherpa_onnx/src/main/cpp/non-streaming-tts.cc
@@ -790,6 +790,13 @@ static void SettleIfReady(Napi::Env env,
     return;
   }
 
+  if (state->audio == nullptr) {
+    state->deferred.Reject(
+        Napi::Error::New(env, "TTS generation produced no audio").Value());
+    state->settled = true;
+    return;
+  }
+
   Napi::Object ans = Napi::Object::New(env);
   if (state->use_external_buffer) {
     const SherpaOnnxGeneratedAudio *audio = state->audio;
@@ -1020,7 +1027,7 @@ class TtsGenerateWorker : public Napi::AsyncWorker {
     state_->generation_done = true;
     state_->audio = audio_;
     state_->use_external_buffer = use_external_buffer_;
-    if (sentinel_failed_) {
+    if (sentinel_failed_ || tsfn_closing_) {
       // No drain signal will ever arrive; fail the generation instead of
       // leaving the promise pending forever
       if (state_->error.empty()) {
@@ -1238,7 +1245,7 @@ class TtsGenerateWithConfigWorker : public Napi::AsyncWorker {
     state_->generation_done = true;
     state_->audio = audio_;
     state_->use_external_buffer = use_external_buffer_;
-    if (sentinel_failed_) {
+    if (sentinel_failed_ || tsfn_closing_) {
       // No drain signal will ever arrive; fail the generation instead of
       // leaving the promise pending forever
       if (state_->error.empty()) {
--- a/harmony-os/SherpaOnnxHar/sherpa_onnx/src/main/cpp/non-streaming-tts.cc
+++ b/harmony-os/SherpaOnnxHar/sherpa_onnx/src/main/cpp/non-streaming-tts.cc
@@ -5,7 +5,6 @@
 #include <algorithm>
 #include <atomic>
 #include <memory>
-#include <mutex>  // NOLINT
 #include <sstream>
 #include <string>
 #include <vector>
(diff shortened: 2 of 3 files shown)


csukuangfj (approved):

PR #3787, opened 2026-07-23
Fix adaptive KV cache for models with hard-coded cache length

Kn0688:
This PR fixes the CI failure reported after #3777.

#3777 assumed the decoder KV cache length is dynamic, but FireRedASR v1 models hard-code it to 1024 in the ONNX graph.

Fix: detect the hard-coded length from `in_n_layer_self_k_cache` input shape and use it when present. v2 models with dynamic length continue to use adaptive allocation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with speech recognition models that use a fixed decoder cache length.
  * Enhanced decoder initialization to automatically detect and apply the model’s configured cache size.
  * Preserved existing cache-length validation and fallback behavior for models without a fixed cache length.


csukuangfj (approved):

Kn0688:
Thank you for merging!

Issue #3791, opened 2026-07-23
[BUG] KeywordSpotter and streaming zipformer2 ASR broken on macOS arm64 (Apple M4) in 1.13.4 — bundled ORT 1.27.0 SME Conv bug, fixed in ORT 1.27.1


kecoco16:
### Important Instructions ⚠️

Before submitting this issue, please **read carefully** and confirm the following:

- [x] I am using the **latest version** of `sherpa-onnx`. If not, I will upgrade and check if the issue persists.
- [x] I understand that **issues not reproducible on the latest version** may be closed without response.
- [x] I have verified that the error is reproducible using **only `sherpa-onnx`**, not code from my own project.
- [x] I will provide **detailed steps** to reproduce the issue.
- [x] I will provide **complete logs** and error messages.
- [x] I understand that **issues not following these instructions** may be closed or receive no response (如果直接删掉这个issue template 里的步骤，我们很有可能**不会回复**您的issue).

---

### Describe the issue

On macOS arm64 machines with `FEAT_SME` (reproduced on an Apple M4 Max), sherpa-onnx **1.13.4** silently breaks the tested streaming **zipformer2** inference paths:

- **KeywordSpotter detects nothing, ever** — even the canonical `nodejs-addon-examples/test_keyword_spotter_transducer.js` with the wenetspeech KWS model and its own `test_wavs/3.wav` returns an empty result. Same for the gigaspeech and zh-en-2025-12-20 KWS models, fp32 and int8, at any `keywordsThreshold` (tried down to 0.01 with `keywordsScore` 5).
- **Online zipformer transducer ASR returns garbage** — `sherpa-onnx-streaming-zipformer-en-2023-06-26` transcribes a clean "what time is it right now" utterance as `"MY WOMAN"`.

**1.13.3 behaves correctly** on the same machine, same models, same wavs, same scripts. Not affected under 1.13.4: offline transducer models (`nemo_transducer` Parakeet TDT), online nemotron FastConformer, offline TTS (Kokoro), Silero VAD — which is why the breakage is silent for most pipelines.

[...]

PR #3792, opened 2026-07-23
Forward `SessionConfig.`-prefixed provider-config keys to Ort::SessionOptions::AddConfigEntry

kecoco16:
## Description

The provider config file mechanism (`provider: "cpu:/path/to/config.conf"`) currently understands a fixed set of keys (`GraphOptimizationLevel`, `LogSeverityLevel`, `ProfilingFilePrefix`, `EnableMemPattern`, `EnableCpuMemArena`, plus provider-specific ones); everything else is silently dropped. That means onnxruntime session configuration entries (`onnxruntime_session_options_config_keys.h`) are unreachable without a dedicated sherpa-onnx option for each.

This PR forwards keys carrying an explicit `SessionConfig.` prefix to `Ort::SessionOptions::AddConfigEntry`, with the prefix stripped:

```
# provider config file
SessionConfig.mlas.disable_kleidiai=1
```

becomes `AddConfigEntry("mlas.disable_kleidiai", "1")`.

Only the explicit namespace is forwarded — unknown un-prefixed keys keep their current behavior, so provider-specific keys (e.g. `SPACEMIT_EP_*`) and un-prefixed typos are not silently passed to onnxruntime. (A typo *inside* the `SessionConfig.` namespace is intentionally treated as an ORT session-config entry; unsupported entries may be ignored by ORT.)

## Tests

`session-test.cc` (registered in `sherpa_onnx_test_srcs`) covers: prefixed keys forwarded with the prefix stripped (verified via `HasSessionConfigEntry`/`GetSessionConfigEntry`), the prefixed spelling itself not registered, known sherpa-onnx keys (`EnableCpuMemArena`) consumed and not forwarded, un-prefixed provider-specific keys (`SPACEMIT_EP_*`) left alone, and the empty-key edge (`SessionConfig.=1`) ignored without error.

## Motivation

While diagnosing #3791 (darwin-arm64 KWS/zipformer2-streaming breakage in 1.13.4: bundled ORT 1.27.0's KleidiAI SME Conv miscomputes axis-asymmetric `pads`, fixed upstream in ORT 1.27.1 via microsoft/onnxruntime#28571 / #29628), the verified user-side mitigation was the ORT session config entry `mlas.disable_kleidiai=1` — but there is currently no way to reach it through sherpa-onnx. A generic, namespaced passthrough makes this class of runtime workaround (and legitimate tuning entries like `session.disable_prepacking`, `session.dynamic_block_base`, etc.) available without new sherpa-onnx releases.

This is independent of — and complementary to — bumping the bundled ORT to 1.27.1 or later (the actual fix for #3791).

## Summary by CodeRabbit

* **New Features**
  * Provider configuration files can now forward `SessionConfig.*` entries into ONNX Runtime session options.
  * The `SessionConfig.` prefix is automatically stripped for the forwarded session setting names.
* **Bug Fixes**
  * Ensures only `SessionConfig.*` keys are forwarded, and ignores empty `SessionConfig.` keys.
* **Tests**
  * Added a dedicated session configuration test to verify correct forwarding, filtering, and key-handling behavior when tests are enabled.

### DIFF
--- a/sherpa-onnx/csrc/CMakeLists.txt
+++ b/sherpa-onnx/csrc/CMakeLists.txt
@@ -832,6 +832,7 @@ if(SHERPA_ONNX_ENABLE_TESTS)
     packed-sequence-test.cc
     pad-sequence-test.cc
     regex-lang-test.cc
+    session-test.cc
     slice-test.cc
     stack-test.cc
     text-utils-test.cc
--- a/sherpa-onnx/csrc/session-test.cc
+++ b/sherpa-onnx/csrc/session-test.cc
@@ -1,0 +1,72 @@
+// sherpa-onnx/csrc/session-test.cc
+//
+// Copyright (c)  2026  Kevin Castillo
+
+#include "sherpa-onnx/csrc/session.h"
+
+#include <cstdio>
+#include <fstream>
+#include <string>
+
+#include "gtest/gtest.h"
+
+namespace sherpa_onnx {
+
+static bool HasConfigEntry(const Ort::SessionOptions &sess_opts,
+                           const char *key) {
+  int has = 0;
+  Ort::ThrowOnError(Ort::GetApi().HasSessionConfigEntry(
+      static_cast<const OrtSessionOptions *>(sess_opts), key, &has));
+  return has != 0;
+}
+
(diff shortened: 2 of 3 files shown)


csukuangfj (approved):

PR #3796, opened 2026-07-24
Raise FireRedASR decode cap from 6 to 8 tokens/sec for fast speech

Kn0688:
Trades ~31% larger KV cache and ~3% e2e time for truncation safety up to 8 tokens/s. Texts verified token-identical on test wavs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved speech recognition decoding capacity for longer or faster-spoken audio by using a configurable token-rate limit during greedy decoding.
  * Reduced the risk of prematurely limiting decoding steps by replacing a fixed token-per-second heuristic with the stream option value (default remains 6), with minimal impact to decoding time.

### DIFF
--- a/sherpa-onnx/csrc/offline-fire-red-asr-decoder.h
+++ b/sherpa-onnx/csrc/offline-fire-red-asr-decoder.h
@@ -27,12 +27,14 @@ class OfflineFireRedAsrDecoder {
    *                              (num_decoder_layers, N, T, d_model).
    * @param n_layer_cross_v       A 4-D tensor of shape
    *                              (num_decoder_layers, N, T, d_model).
+   * @param num_feature_frames    Number of feature frames in the input audio.
+   * @param max_token_per_second  Maximum tokens generated per second of audio.
    *
    * @return Return a vector of size `N` containing the decoded results.
    */
   virtual std::vector<OfflineFireRedAsrDecoderResult> Decode(
       Ort::Value n_layer_cross_k, Ort::Value n_layer_cross_v,
-      int32_t num_feature_frames) = 0;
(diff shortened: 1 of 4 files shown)


csukuangfj:
Thank you for your contribution! Can you make it configurable per stream?

Please refer to
https://github.com/k2-fsa/sherpa-onnx/blob/77322d37a8cb9e336e710b4a17bff3d9ea69bb99/sherpa-onnx/csrc/offline-recognizer-qwen3-asr-impl.cc#L673-L674

You can change
https://github.com/k2-fsa/sherpa-onnx/blob/77322d37a8cb9e336e710b4a17bff3d9ea69bb99/sherpa-onnx/csrc/offline-recognizer-fire-red-asr-impl.h#L124-L125
to

```c++

int32_t max_token_per_second = stream->GetOptionInt("max_token_per_second", 6); 

auto results = decoder_->Decode(std::move(cross_kv.first), 
                                 std::move(cross_kv.second), num_frames, max_token_per_second);
```

You need to change the signature of `Decode()` to accept an additional argument.

csukuangfj (approved):
Thanks!

Kn0688:
One more thing to discuss: in my tests with short videos, I noticed that when `max_token_per_second` is set to 6, some characters are occasionally dropped. Using 8 seems more stable in those cases.So do you prefer to keep the default at 6, or would you be open to 8?

csukuangfj:
> One more thing to discuss: in my tests with short videos, I noticed that when `max_token_per_second` is set to 6, some characters are occasionally dropped. Using 8 seems more stable in those cases.So do you prefer to keep the default at 6, or would you be open to 8?

8 is fine with me.

PR #3806, opened 2026-07-26
fix UB

x6prl:
Default constructor does not init `debug` variable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability by ensuring homophone replacement debugging is disabled by default.
  * Prevented unpredictable behavior caused by an uninitialized debug setting.


csukuangfj (approved):

PR #3818, opened 2026-07-28
Publish an android-x64 .NET runtime package

### LINKED ISSUE #3817: [BUG] `.NET` package: no `android-x64` RID, so RID fallback ships the glibc `linux-x64` `.so` (4 KB aligned) into Android app bundles
### Important Instructions ⚠️

Before submitting this issue, please **read carefully** and confirm the following:

- [x] I am using the **latest version** of `sherpa-onnx`. If not, I will upgrade and check if the issue persists.
- [x] I understand that **issues not reproducible on the latest version** may be closed without response.
- [x] I have verified that the error is reproducible using **only `sherpa-onnx`**, not code from my own project.
- [x] I will provide **detailed steps** to reproduce the issue.
- [x] I will provide **complete logs** and error messages.
- [x] I understand that **issues not following these instructions** may be closed or receive no response.

---

### Describe the issue

This is a **packaging** issue in the .NET/NuGet distribution, not in the C++ build. The Android release tarball is already correct — I verified it below.

When a .NET MAUI / .NET for Android app references `org.k2fsa.sherpa.onnx` and builds for the `x86_64` ABI, the resulting APK/AAB contains **`lib/x86_64/libsherpa-onnx-c-api.so` and `lib/x86_64/libonnxruntime.so` with `PT_LOAD` `p_align = 0x1000` (4 KB)**. Every other 64-bit `.so` in the bundle is 16 KB aligned, including the `arm64-v8a` sherpa-onnx libraries.

This blocks Google Play submission: Play requires all apps targeting Android 15+ to support 16 KB memory page sizes, and from 31 May 2026 updates that don't comply are rejected.

**Likely root cause.** `scripts/dotnet/generate.py` only calls `process_android(s, "arm64")`, and `scripts/dotnet/run.sh` has `RIDS=(linux-x64 linux-arm64 android-arm64 macos-x64 macos-arm64 windows-x64 windows-x86 windows-arm64)`. There is no `org.k2fsa.sherpa.onnx.runtime.android-x64` package. Because the .NET RID graph falls back `android-x64` → `linux-bionic-x64` → `linux-x64`, the build resolves `runtimes/linux-x64/native/libsherpa-onnx-c-api.so` and `libonnxruntime.so` — the **manylinux2014 glibc** binaries taken from the Python wheel — and packs them into `lib/x86_64/` of the Android bundle.

That has two consequences:

1. **Alignment** — the Linux builds are 4 KB aligned, so the whole app fails Play's 16 KB check.
2. **Wrong libc** — these are glibc binaries being shipped into a bionic app. Their `DT_NEEDED` entries include `ld-linux-x86-64.so.2`, `libc.so.6` and `librt.so.1` (full output under *Steps to reproduce*), none of which exist on Android. They cannot `dlopen` on an x86_64 device or emulator regardless of alignment, so `android-x64` is effectively broken for .NET consumers today — the alignment failure is just the symptom that Play surfaces first.

**The correct binaries already exist.** I extracted the official `sherpa-onnx-v1.13.4-android.tar.bz2` release asset and every `x86_64` library in it is already 16 KB aligned:

```
$ for abi in arm64-v8a x86_64; do
    for f in jniLibs/$abi/*.so; do
      echo "$f  $(readelf -lW "$f" | awk '$1=="LOAD"{print $NF}' | sort -u | tr '\n' ' ')"
    done
  done

jniLibs/arm64-v8a/libonnxruntime.so         0x4000
jniLibs/arm64-v8a/libsherpa-onnx-c-api.so   0x4000
jniLibs/arm64-v8a/libsherpa-onnx-cxx-api.so 0x4000
jniLibs/arm64-v8a/libsherpa-onnx-jni.so     0x4000
jniLibs/x86_64/libonnxruntime.so            0x4000
jniLibs/x86_64/libsherpa-onnx-c-api.so      0x4000
jniLibs/x86_64/libsherpa-onnx-cxx-api.so    0x4000
jniLibs/x86_64/libsherpa-onnx-jni.so        0x4000
```

So the fix is only to publish an `android-x64` runtime package sourced from `jniLibs/x86_64/` of that same tarball, exactly as `android-arm64` is already sourced from `jniLibs/arm64-v8a/`.

### Steps to reproduce

Using only sherpa-onnx artifacts, no project code:

1. Download and unzip the runtime package that RID fallback selects, and inspect the two libraries it contains:

   ```
   curl -L -o linux-x64.nupkg \
     https://www.nuget.org/api/v2/package/org.k2fsa.sherpa.onnx.runtime.linux-x64/1.13.4
   unzip -oq linux-x64.nupkg -d linux-x64

   for f in linux-x64/runtimes/linux-x64/native/*.so; do
     echo "== $f"
     readelf -lW "$f" | awk '$1=="LOAD"{print $NF}' | sort -u
     readelf -dW "$f" | grep NEEDED
   done
   ```

   Actual output:

   ```
   == linux-x64/runtimes/linux-x64/native/libonnxruntime.so
   0x1000
    0x0000000000000001 (NEEDED)  Shared library: [libdl.so.2]
    0x0000000000000001 (NEEDED)  Shared library: [librt.so.1]
    0x0000000000000001 (NEEDED)  Shared library: [libpthread.so.0]
    0x0000000000000001 (NEEDED)  Shared library: [libstdc++.so.6]
    0x0000000000000001 (NEEDED)  Shared library: [libm.so.6]
    0x0000000000000001 (NEEDED)  Shared library: [libgcc_s.so.1]
    0x0000000000000001 (NEEDED)  Shared library: [libc.so.6]
    0x0000000000000001 (NEEDED)  Shared library: [ld-linux-x86-64.so.2]
   == linux-x64/runtimes/linux-x64/native/libsherpa-onnx-c-api.so
   0x1000
    0x0000000000000001 (NEEDED)  Shared library: [libpthread.so.0]
    0x0000000000000001 (NEEDED)  Shared library: [libonnxruntime.so]
    0x0000000000000001 (NEEDED)  Shared library: [libdl.so.2]
    0x0000000000000001 (NEEDED)  Shared library: [libm.so.6]
    0x0000000000000001 (NEEDED)  Shared library: [libstdc++.so.6]
    0x0000000000000001 (NEEDED)  Shared library: [libgcc_s.so.1]
    0x0000000000000001 (NEEDED)  Shared library: [libc.so.6]
    0x0000000000000001 (NEEDED)  Shared library: [ld-linux-x86-64.so.2]
   ```

   `0x1000` matches the alignment observed in the app bundle exactly, and the `DT_NEEDED` list confirms these are glibc binaries — `ld-linux-x86-64.so.2` and `libc.so.6` cannot be satisfied under bionic, where the dynamic linker is `linker64`. Note also that `libsherpa-onnx-c-api.so` depends on `libonnxruntime.so`, which is why both files are pulled into `lib/x86_64/` together.

2. Confirm there is no `android-x64` runtime package published, while `android-arm64` is:

   ```
   for rid in android-x64 android-arm64; do
     curl -s -o /dev/null -w "$rid: %{http_code}\n" \
       https://api.nuget.org/v3-flatcontainer/org.k2fsa.sherpa.onnx.runtime.$rid/index.json
   done
   ```

3. Reproduce end to end in a bare .NET Android app (no sherpa-onnx API calls needed — the libraries are packed by the RID resolver alone):

   ```
   dotnet new android -o SherpaRidRepro
   cd SherpaRidRepro
   dotnet add package org.k2fsa.sherpa.onnx --version 1.13.4
   # set <RuntimeIdentifiers>android-arm64;android-x64</RuntimeIdentifiers> in the csproj
   dotnet publish -c Release -f net9.0-android
   ```

4. Unzip the resulting `.aab`/`.apk` and check every 64-bit `.so`:

   ```
   readelf -lW base/lib/x86_64/libsherpa-onnx-c-api.so | grep LOAD   # 0x1000
   readelf -lW base/lib/arm64-v8a/libsherpa-onnx-c-api.so | grep LOAD # 0x4000
   ```

Removing the `org.k2fsa.sherpa.onnx` package reference is the only thing that makes the bundle pass; the two failing entries disappear and every remaining 64-bit library is 16 KB aligned.

### Expected behavior

`x86_64` Android builds should get the bionic, 16 KB-aligned binaries that are already published in `sherpa-onnx-vX.Y.Z-android.tar.bz2`.

Concretely: add `process_android(s, "x64")` (and ideally `arm`/`x86` for completeness) to `scripts/dotnet/generate.py`, add `android-x64` to `RIDS` in `scripts/dotnet/run.sh`, copying from `jniLibs/x86_64/` of the existing Android tarball. That also removes the `linux-x64` fallback entirely, fixing the glibc-in-a-bionic-app problem at the same time.

### Environment

- `sherpa-onnx` version: **1.13.4** (NuGet `org.k2fsa.sherpa.onnx`); also reproduced on 1.12.25
- OS: build host Windows 11 (VS 2026); target Android (`net9.0-android`, `minSdk` 24, `targetSdk` 35)
- Python version: N/A (.NET 9 / .NET MAUI consumer)
- onnxruntime version: whichever is bundled in the linux-x64 wheel for 1.13.4
- GPU (if applicable): N/A
- NPU (if applicable): N/A

### Logs

Alignment audit of the signed `.aab` (only the failing rows and their `arm64-v8a` counterparts shown; all 50 other 64-bit libraries report `OK_16KB`):

```
Path                                              Abi        Bits MinAlign Status
----                                              ---        ---- -------- ------
base/lib/x86_64/libonnxruntime.so                 x86_64       64 0x1000   FAIL_4KB
base/lib/x86_64/libsherpa-onnx-c-api.so           x86_64       64 0x1000   FAIL_4KB
base/lib/arm64-v8a/libonnxruntime.so              arm64-v8a    64 0x4000   OK_16KB
base/lib/arm64-v8a/libsherpa-onnx-c-api.so        arm64-v8a    64 0x4000   OK_16KB

Total: 52  OK: 50  Fail: 2  Skipped32: 0

FAIL: 64-bit libraries below 16 KB alignment:
  base/lib/x86_64/libonnxruntime.so
  base/lib/x86_64/libsherpa-onnx-c-api.so
```

On 1.12.25 the same two `x86_64` entries reported `0x1000`, while the `arm64-v8a` pair reported `0x10000`. After removing the package reference:

```
Total: 50  OK: 50  Fail: 0  Skipped32: 0
PASS: All 64-bit native libraries are 16 KB aligned.
```

kfrancis:
Fixes #3817

## Problem

The .NET distribution ships per-RID native packages plus the `org.k2fsa.sherpa.onnx` meta-package, but there is no `android-x64` runtime package. The .NET RID graph falls back `android-x64` → `linux-bionic-x64` → `linux-x64`, so any .NET for Android or MAUI app building for the `x86_64` ABI silently resolves the **manylinux glibc** binaries from `org.k2fsa.sherpa.onnx.runtime.linux-x64` and packs them into `lib/x86_64/` of the APK/AAB. Those binaries cannot load under bionic at all, and additionally fail Google Play's 16 KB page size requirement (which blocks app updates from 31 May 2026).

`android-x64` is the only affected ABI: `android-arm64` already has a more specific package that wins RID resolution, and the 32-bit Android RIDs have no `linux-arm`/`linux-x86` packages to fall back to, so they fail loudly at restore/build time instead of packing broken binaries.

## Evidence

Broken binaries currently resolved via fallback — `org.k2fsa.sherpa.onnx.runtime.linux-x64` 1.13.4, `runtimes/linux-x64/native/`:

```
libsherpa-onnx-c-api.so
  PT_LOAD p_align: 0x1000
  NEEDED: libpthread.so.0, libonnxruntime.so, libdl.so.2, libm.so.6,
          libstdc++.so.6, libgcc_s.so.1, libc.so.6, ld-linux-x86-64.so.2
libonnxruntime.so
  PT_LOAD p_align: 0x1000
  NEEDED: libdl.so.2, librt.so.1, libpthread.so.0, libstdc++.so.6,
          libm.so.6, libgcc_s.so.1, libc.so.6, ld-linux-x86-64.so.2
```

4 KB `p_align` fails the 16 KB page size requirement; the glibc/`ld-linux-x86-64.so.2` dependencies cannot be satisfied by bionic.

Correct binaries already published — `sherpa-onnx-v1.13.4-android.tar.bz2`, `jniLibs/x86_64/`:

```
libsherpa-onnx-c-api.so
  PT_LOAD p_align: 0x4000
  NEEDED: libandroid.so, liblog.so, libonnxruntime.so, libm.so, libdl.so, libc.so
libonnxruntime.so
  PT_LOAD p_align: 0x4000
  NEEDED: libdl.so, liblog.so, libm.so, libc.so
```

16 KB aligned, bionic-only dependencies. This is the same release asset the `android-arm64` package is already built from — the fix is packaging only; nothing is recompiled.

## Changes

- `scripts/dotnet/run.sh` — add `android-x64` to `RIDS`; restructure the Android download block so the single `sherpa-onnx-v<VERSION>-android.tar.bz2` is downloaded at most once and feeds both `android-arm64` (from `jniLibs/arm64-v8a/`) and `android-x64` (from `jniLibs/x86_64/`), keeping the per-RID idempotency guards. Renamed `android_arm64_tarball*` vars to `android_tarball*` since the tarball was never arm64-specific.
- `scripts/dotnet/generate.py` — add `process_android(s, "x64")`; `process_android` was already RID-parameterized.
- `scripts/dotnet/sherpa-onnx.csproj.in` — add `android-x64` to `<RuntimeIdentifiers>` and a `PackageReference` to `org.k2fsa.sherpa.onnx.runtime.android-x64` in the meta-package.
- `scripts/dotnet/.gitignore` — ignore the generated `android-x64/` staging directory, matching the other RID directories.

Verified locally: `bash -n run.sh` and `ast.parse(generate.py)` pass; rendering `sherpa-onnx.csproj.runtime.in` for `android-x64` produces `<RuntimeIdentifier>android-x64</RuntimeIdentifier>`, `<PackageId>org.k2fsa.sherpa.onnx.runtime.android-x64</PackageId>`, and `PackagePath` `runtimes/android-x64/native/%(Filename)%(Extension)`; the meta template renders the new RID and package reference.

No changes to `.github/workflows/` were needed: `dot-net.yaml` and `test-dot-net.yaml` invoke `scripts/dotnet/run.sh` and push `org.k2fsa.sherpa.onnx.*.nupkg` with a wildcard — they don't enumerate RIDs. All other `android-arm64` grep hits are NDK build scripts (`build-android-arm64-v8a.sh`), APK scripts, or Flutter packaging, which are unrelated to .NET RIDs.

## Notes

- This takes effect for consumers only after a maintainer with NuGet publish credentials runs `scripts/dotnet/run.sh` (via the `dot-net.yaml` release workflow) for a release; merging alone publishes nothing.
- `android-arm` / `android-x86` (32-bit) are deliberately not added. They are not silently broken today — with no `linux-arm`/`linux-x86` fallback packages, resolution fails visibly instead of packing wrong binaries. The tarball does contain `armeabi-v7a/` and `x86/` libraries, so the same pattern could add them later if anyone still targets 32-bit Android; kept out of scope here to keep the review focused.

## Summary by CodeRabbit

- **New Features**
  - Added support for Android x64 runtime distribution alongside Android arm64.
  - Android x64 native libraries are now included in generated .NET runtime packages.

- **Bug Fixes**
  - Improved Android runtime packaging to reliably acquire and include required native libraries for both supported architectures.

- **Chores**
  - Updated runtime configuration and build automation to recognize the Android x64 target.

### DIFF
--- a/scripts/dotnet/.gitignore
+++ b/scripts/dotnet/.gitignore
@@ -4,6 +4,7 @@ macos-x64
 linux-x64
 linux-arm64
 android-arm64
+android-x64
 windows-arm64
 windows-x64
 windows-x86
--- a/scripts/dotnet/generate.py
+++ b/scripts/dotnet/generate.py
@@ -121,6 +121,7 @@ def main():
     process_linux(s, "x64")
     process_linux(s, "arm64")
     process_android(s, "arm64")
+    process_android(s, "x64")
     process_windows(s, "x64")
     process_windows(s, "x86")
     process_windows(s, "arm64")
--- a/scripts/dotnet/run.sh
+++ b/scripts/dotnet/run.sh
@@ -24,7 +24,7 @@ export src_dir
 mkdir -p $src_dir
 pushd $src_dir
 
-RIDS=(linux-x64 linux-arm64 android-arm64 macos-x64 macos-arm64 windows-x64 windows-x86 windows-arm64)
+RIDS=(linux-x64 linux-arm64 android-arm64 android-x64 macos-x64 macos-arm64 windows-x64 windows-x86 windows-arm64)
 
 mkdir -p ${RIDS[@]}
 
@@ -34,8 +34,8 @@ linux_x64_wheel=$src_dir/$linux_x64_wheel_filename
 linux_arm64_wheel_filename=sherpa_onnx_core-${SHERPA_ONNX_VERSION}-py3-none-manylinux2014_aarch64.whl
 linux_arm64_wheel=$src_dir/$linux_arm64_wheel_filename
 
-android_arm64_tarball_filename=sherpa-onnx-v$SHERPA_ONNX_VERSION-android.tar.bz2
-android_arm64_tarball=$src_dir/$android_arm64_tarball_filename
+android_tarball_filename=sherpa-onnx-v$SHERPA_ONNX_VERSION-android.tar.bz2
+android_tarball=$src_dir/$android_tarball_filename
 
 macos_x64_wheel_filename=sherpa_onnx_core-${SHERPA_ONNX_VERSION}-py3-none-macosx_10_15_x86_64.whl
 macos_x64_wheel=$src_dir/$macos_x64_wheel_filename
@@ -88,24 +88,32 @@ if [ ! -f $src_dir/linux-arm64/libsherpa-onnx-c-api.so ]; then
   cd ..
 fi
 
-if [ ! -f $src_dir/android-arm64/libsherpa-onnx-c-api.so ]; then
-  echo "---android arm64---"
-  cd android-arm64
+if [ ! -f $src_dir/android-arm64/libsherpa-onnx-c-api.so ] || [ ! -f $src_dir/android-x64/libsherpa-onnx-c-api.so ]; then
+  echo "---android---"
   mkdir -p tarball
   cd tarball
-  if [ -f $android_arm64_tarball  ]; then
-    cp -v $android_arm64_tarball .
+  if [ -f $android_tarball  ]; then
+    cp -v $android_tarball .
   else
-    curl -OL "https://github.com/k2-fsa/sherpa-onnx/releases/download/v$SHERPA_ONNX_VERSION/$android_arm64_tarball_filename"
+    curl -OL "https://github.com/k2-fsa/sherpa-onnx/releases/download/v$SHERPA_ONNX_VERSION/$android_tarball_filename"
   fi
-  tar xjf $android_arm64_tarball_filename
-  cp -v jniLibs/arm64-v8a/lib{onnxruntime,sherpa-onnx-c-api}.so ../
+  tar xjf $android_tarball_filename
+
+  if [ ! -f $src_dir/android-arm64/libsherpa-onnx-c-api.so ]; then
+    echo "---android arm64---"
+    cp -v jniLibs/arm64-v8a/lib{onnxruntime,sherpa-onnx-c-api}.so ../android-arm64/
+    ls -lh ../android-arm64
+  fi
+
+  if [ ! -f $src_dir/android-x64/libsherpa-onnx-c-api.so ]; then
+    echo "---android x64---"
+    cp -v jniLibs/x86_64/lib{onnxruntime,sherpa-onnx-c-api}.so ../android-x64/
+    ls -lh ../android-x64
+  fi
 
   cd ..
 
   rm -rf tarball
-  ls -lh
-  cd ..
 fi
 
 if [ ! -f $src_dir/macos-x64/libsherpa-onnx-c-api.dylib ]; then
--- a/scripts/dotnet/sherpa-onnx.csproj.in
+++ b/scripts/dotnet/sherpa-onnx.csproj.in
@@ -5,7 +5,7 @@
     <OutputType>Library</OutputType>
     <LangVersion>10.0</LangVersion>
     <TargetFrameworks>net10.0;net8.0;net7.0;net6.0;net45;net40;net35;net20;netstandard2.0</TargetFrameworks>
-    <RuntimeIdentifiers>linux-x64;linux-arm64;android-arm64;osx-x64;osx-arm64;win-x64;win-x86;win-arm64</RuntimeIdentifiers>
+    <RuntimeIdentifiers>linux-x64;linux-arm64;android-arm64;android-x64;osx-x64;osx-arm64;win-x64;win-x86;win-arm64</RuntimeIdentifiers>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <AssemblyName>sherpa-onnx</AssemblyName>
     <Version>{{ version }}</Version>
@@ -51,6 +51,7 @@
     <PackageReference Include="org.k2fsa.sherpa.onnx.runtime.linux-x64" Version="{{ version }}" />
     <PackageReference Include="org.k2fsa.sherpa.onnx.runtime.linux-arm64" Version="{{ version }}" />
     <PackageReference Include="org.k2fsa.sherpa.onnx.runtime.android-arm64" Version="{{ version }}" />
+    <PackageReference Include="org.k2fsa.sherpa.onnx.runtime.android-x64" Version="{{ version }}" />
     <PackageReference Include="org.k2fsa.sherpa.onnx.runtime.osx-x64"   Version="{{ version }}" />
     <PackageReference Include="org.k2fsa.sherpa.onnx.runtime.osx-arm64" Version="{{ version }}" />
     <PackageReference Include="org.k2fsa.sherpa.onnx.runtime.win-x64"   Version="{{ version }}" />


kfrancis:
Thanks — keeping the single-file guard as-is, for consistency with the rest of the script. All seven other download blocks in run.sh use the same one-sentinel idiom (check a single library, e.g. libsherpa-onnx-c-api.so), and this PR deliberately preserves the guard that the original android-arm64 block already had.

A partially-staged directory can't occur through normal execution: both libraries are copied by a single cp invocation, and the script runs under set -e, so a failed copy aborts the run before leaving partial state. CI starts from a fresh /tmp each run, so the only way to hit the missing-libonnxruntime.so case is manually deleting one file from a local cache — and rm -rf $src_dir/android-* recovers that.

Happy to switch to the both-files check if the maintainers prefer it, but I'd rather not diverge from the established style of the script in this PR.

csukuangfj (approved):

PR #263, opened 2023-08-12
Support streaming paraformer

csukuangfj:
## TODOs
- [ ] Update C API
- [ ] Update C# API
- [ ] Update Go API
- [ ] Update jni
- [ ] Update Swift API

PR #3821, opened 2026-07-29
Fix streaming paraformer feature frontend drift from its training config

slipstr34m:
## Problem

The offline paraformer configures the feature frontend the model was trained with (offline-recognizer-paraformer-impl.h:210-217):

```cpp
void InitFeatConfig() {
  // Paraformer models assume input samples are in the range
  // [-32768, 32767], so we set normalize_samples to false
  config_.feat_config.normalize_samples = false;
  config_.feat_config.window_type = "hamming";
  config_.feat_config.high_freq = 0;
  config_.feat_config.snip_edges = true;
}
```

The online paraformer sets only `normalize_samples`, in both constructors, so streaming decoding runs on the fbank defaults (povey window, high_freq=-400, snip_edges=false) while the model's upstream config.yaml (modelscope speech_paraformer, online variant) specifies `window: hamming`. We noticed this while running the streaming and offline paraformer on the same audio: the streaming output was consistently worse, and the frontend difference turned out to be the reason. A user in #3511 found the same drift and worked around it with `--high-freq=0 --snip-edges=true --window-type=hamming`. Related: #3511.

## Fix

Give the online impl the same private `InitFeatConfig()` and call it from both constructors. The gap dates to 6038e2aa (#263), which copied the offline paraformer's then-current frontend handling; 25f0a104 (#1148) later upgraded only the offline impl.

## Tested

New `test_streaming_paraformer` in test_online_recognizer.py decodes test_wavs/2.wav of sherpa-onnx-streaming-paraformer-bilingual-zh-en (greedy search, deterministic):

```
before: 这个是平繁的啊不认识接下来 frequently 苹繁的
after:  这个是频繁的啊不认识接下来 frequently 频繁的
```

The audio says 频繁 twice; the default frontend misrecognizes it both times. Individual tokens on other utterances can flip either way, but aggregate accuracy improves (about 10 errors fixed on a 4-minute zh recording, converging toward the offline paraformer's output on the same audio). The test fails on master without this change and passes with it; the full test_online_recognizer.py suite passes with no new skips. `./scripts/check_style_cpplint.sh` passes.

Only touches online-recognizer-paraformer-impl.h and its Python test; the offline paraformer and all other online recognizers are untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Paraformer streaming speech recognition feature configuration for more consistent audio processing.

* **Tests**
  * Added coverage for streaming Paraformer recognition with bilingual Chinese-English transcription.
  * Validates decoding results against an expected transcript.

### DIFF
--- a/sherpa-onnx/csrc/online-recognizer-paraformer-impl.h
+++ b/sherpa-onnx/csrc/online-recognizer-paraformer-impl.h
@@ -116,9 +116,7 @@ class OnlineRecognizerParaformerImpl : public OnlineRecognizerImpl {
       SHERPA_ONNX_EXIT(-1);
     }
 
-    // Paraformer models assume input samples are in the range
-    // [-32768, 32767], so we set normalize_samples to false
-    config_.feat_config.normalize_samples = false;
+    InitFeatConfig();
   }
 
   template <typename Manager>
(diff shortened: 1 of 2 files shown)


csukuangfj (approved):

PR #3838, opened 2026-08-03
Fix Dart speaker diarization input buffer leak

zhangheng2022:
## Summary

- release the native waveform buffer allocated by Dart speaker diarization APIs
- clean up diarization results on every return or exception path
- close the progress callback wrapper reliably

## Root cause

Both `OfflineSpeakerDiarization.process()` and `processWithCallback()` copy the complete waveform into a `calloc<Float>` buffer, but never free that buffer. The allocation therefore remains in the process native heap after each call.

For example, a 30-minute 16 kHz waveform allocates 28.8 million floats, or 115.2 MB, per invocation.

## Validation

- `flutter pub get`
- `flutter analyze`
- `git diff --check`

The `flutter/sherpa_onnx` package currently has no unit test directory; this change is limited to deterministic FFI resource cleanup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved speaker diarization resource cleanup.
  * Ensured processing buffers and native results are released reliably, including when errors occur.
  * Improved callback cleanup after diarization completes.

### DIFF
--- a/flutter/sherpa_onnx/lib/src/offline_speaker_diarization.dart
+++ b/flutter/sherpa_onnx/lib/src/offline_speaker_diarization.dart
@@ -268,20 +268,24 @@ class OfflineSpeakerDiarization {
 
     final n = samples.length;
     final Pointer<Float> p = calloc<Float>(n);
+    try {
+      final pList = p.asTypedList(n);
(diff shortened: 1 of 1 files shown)


csukuangfj (approved):

Issue #3776, opened 2026-07-20
[BUG] Node addon OnlineRecognizer returns near-empty results on some macOS arm64 machines after 1.12.34 -> 1.13.4


0xlau:
### Important Instructions ⚠️

Before submitting this issue, please **read carefully** and confirm the following:

- [x] I am using the **latest version** of `sherpa-onnx`. If not, I will upgrade and check if the issue persists.
- [x] I understand that **issues not reproducible on the latest version** may be closed without response.
- [x] I have verified that the error is reproducible using **only `sherpa-onnx`**, not code from my own project.
- [x] I will provide **detailed steps** to reproduce the issue.
- [ ] I will provide **complete logs** and error messages.
- [x] I understand that **issues not following these instructions** may be closed or receive no response (如果直接删掉这个issue template 里的步骤，我们很有可能**不会回复**您的issue).

---

### Describe the issue

On some native macOS arm64 machines, the Node addon `OnlineRecognizer` returns empty or near-empty results after upgrading from `sherpa-onnx-node@1.12.34` to the latest `1.13.4`.

With `1.13.4`, the native addon and online Zipformer model load without an exception, and PCM audio continues to be supplied to the recognizer, but almost no text is produced. Pinning the Node package back to `1.12.34` restores normal recognition in the same Electron application flow.

This appears to affect only some macOS arm64 machines rather than every Apple Silicon Mac.

### Steps to reproduce

The issue is currently reproducible in an Electron application on the affected machines. We do not yet have a standalone reproduction outside the application, so the corresponding confirmation box above is intentionally left unchecked.

1. Use a native macOS arm64 machine with Electron `25.8.4`.
2. Install `sherpa-onnx-node@1.13.4`.
3. Create a CPU `OnlineRecognizer` using an online Zipformer transducer model.
4. Continuously feed 16-kHz mono PCM audio and decode ready frames.
5. Observe that the addon and model initialize successfully and PCM is received, but recognition output is empty or nearly empty.
6. Pin `sherpa-onnx-node` to `1.12.34` and repeat the same application flow. Recognition returns to normal.

### Expected behavior

`1.13.4` should produce recognition output comparable to `1.12.34` from the same audio input and online Zipformer workflow.

### Environment

- `sherpa-onnx-node` version: `1.13.4`
- Working baseline: `1.12.34`
- OS: macOS, native arm64 (not Rosetta)
- Electron: `25.8.4`
- Python version: N/A
[...]

PR #3850, opened 2026-08-09
Use onnxruntime 1.27.1 for Linux aarch64 and Android

yentur:
### Problem

On arm64 CPUs that implement `FEAT_SME`, onnxruntime 1.27.0 computes the wrong result for the Conv nodes in the zipformer2 frontend, so streaming zipformer ASR and `KeywordSpotter` return garbage or nothing at all, with no error message. That is #3791 and #3776.

macOS moved to 1.27.1 in #3831 and WASM in #3836, but `cmake/onnxruntime-linux-aarch64{,-static}.cmake` and the four `build-android-*.sh` scripts are still on 1.27.0.

### Root cause

1.27.0's route gate `MlasConvSupportsSymmetricChannelsLast2DFloatKernel` only required begin padding to equal end padding per axis, so `pads=[0,1,0,1]` was admitted into the KleidiAI SME kernel, whose LHS indirection table uses one scalar padding for both axes. Fixed upstream by microsoft/onnxruntime#28571, first shipped in 1.27.1. The full analysis is @kecoco16's, in #3791.

That kernel is compiled into the aarch64 archives this repo pins, not just the macOS ones:

```
# onnxruntime-linux-aarch64-glibc2_17-Release-1.27.0.zip
$ strings lib/libonnxruntime.so | grep -c ConvolveSme
2
# onnxruntime-android-1.27.0.zip
$ strings jni/arm64-v8a/libonnxruntime.so | grep -c ConvolveSme
2
```

### Fix

Bump the two Linux aarch64 cmake files and the default `onnxruntime_version` in the Android build scripts to 1.27.1. All four Android scripts extract from the same `onnxruntime-android-<version>.zip`, so they move together.

The Windows and x86_64 pins are deliberately left on 1.27.0. KleidiAI is aarch64 only, so x86_64 cannot hit this, and the Windows cmake files pin eight per-configuration SHA256s each which I have no way to check against a real build.

### Verification

Apple M4 Pro, `hw.optional.arm.FEAT_SME: 1`. master at 634265c built with `BUILD_SHARED_LIBS=ON`, then only `libonnxruntime.dylib` swapped between the two runs:

```
$ sherpa-onnx-keyword-spotter --keywords-file=.../test_keywords.txt .../test_wavs/3.wav
onnxruntime version : 1.27.0
Audio duration: 8.030 s
                                <- nothing

onnxruntime version : 1.27.1
Audio duration: 8.030 s
{"start_time":0.00, "keyword": "文森特卡索", "timestamps": [0.32, ...], "tokens":["w", "én", ...]}
{"start_time":0.00, "keyword": "法国", "timestamps": [2.32, 2.40, 2.60, 2.64], "tokens":["f", "ǎ", "g", "uó"]}
```

Both edited cmake files were run through `FetchContent` on their own; each downloads and passes its hash check. Changing one hex digit of the new hash makes cmake report `Hash mismatch, removing...`, so the check is live. The GitHub and hf-mirror copies of both archives have the same SHA256 as the ones written into the diff, and the same commands reproduce the 1.27.0 hashes that are already in the repo.

`onnxruntime-android-1.27.1.zip` contains all four `jni/<abi>/libonnxruntime.so` the scripts expect.

I have no arm64 Linux or Android hardware here, so what I verified there is that the archives are the right ones, not that a full build passes.

Refs #3791, #3776

## Summary by CodeRabbit

* **Chores**
  * Updated the bundled ONNX Runtime version to 1.27.1 for Android and Linux aarch64 builds.
  * Updated download sources, checksums, and fallback package references to match the new version.
  * Existing environment-based version overrides remain supported.

### DIFF
--- a/build-android-arm64-v8a.sh
+++ b/build-android-arm64-v8a.sh
@@ -90,7 +90,7 @@ fi
 
 echo "ANDROID_NDK: $ANDROID_NDK"
 sleep 1
-onnxruntime_version=${SHERPA_ONNX_ONNXRUNTIME_VERSION:-1.27.0}
+onnxruntime_version=${SHERPA_ONNX_ONNXRUNTIME_VERSION:-1.27.1}
 
 if [ -n "${SHERPA_ONNXRUNTIME_LIB_DIR:-}" ] && [ -n "${SHERPA_ONNXRUNTIME_INCLUDE_DIR:-}" ]; then
   echo "Using externally provided ONNX Runtime"
--- a/build-android-armv7-eabi.sh
+++ b/build-android-armv7-eabi.sh
@@ -91,7 +91,7 @@ fi
 echo "ANDROID_NDK: $ANDROID_NDK"
 sleep 1
 
-onnxruntime_version=${SHERPA_ONNX_ONNXRUNTIME_VERSION:-1.27.0}
+onnxruntime_version=${SHERPA_ONNX_ONNXRUNTIME_VERSION:-1.27.1}
 
 if [ -n "${SHERPA_ONNXRUNTIME_LIB_DIR:-}" ] && [ -n "${SHERPA_ONNXRUNTIME_INCLUDE_DIR:-}" ]; then
   echo "Using externally provided ONNX Runtime"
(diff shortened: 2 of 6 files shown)


csukuangfj (approved):

PR #3851, opened 2026-08-09
kotlin-api: validate native ptr after creation

sigkillsdottir:
Config-invalid (e.g. missing model files) makes native factory return null; first method call then SIGSEGVs. require(ptr != 0L) in all binding classes and streams, mirroring Java binding and Dart PR #3329.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved initialization error handling across Kotlin audio, speech, recognition, punctuation, denoising, speaker, language identification, text-to-speech, and voice activity detection components.
  - Invalid configurations now fail immediately with a clear `IllegalArgumentException` instead of allowing unusable objects to be created.
  - Stream and text-to-speech initialization now consistently detect creation failures.

### DIFF
--- a/sherpa-onnx/kotlin-api/AudioTagging.kt
+++ b/sherpa-onnx/kotlin-api/AudioTagging.kt
@@ -38,7 +38,9 @@ class AudioTagging(
         } else {
             newFromFile(config)
         }
-    }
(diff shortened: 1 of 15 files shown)


csukuangfj (approved):

PR #3857, opened 2026-08-10
Fix float32 catastrophic cancellation in NemoNormalizePerFeature

### LINKED ISSUE #3853: [BUG] Parakeet Unified: one empty stream makes otherwise-correct batched streams empty
### Important Instructions ⚠️

- [x] I am using the **latest version** of `sherpa-onnx`. The issue reproduces on v1.13.4.
- [x] I verified this using only the `sherpa-onnx-offline` C++ binary and official public audio.
- [x] Detailed reproduction steps and logs are below.

### Describe the issue

With the official Parakeet Unified EN 0.6B INT8 non-streaming model, results from `DecodeStreams` are not independent. A short LibriSpeech utterance that produces an empty result by itself causes a separate, normally transcribed utterance to become empty when both are decoded in one batch.

The behavior is independent of input order and reproduces with both 1 and 8 CPU threads. `sherpa-onnx-offline` exits 0 and emits structurally valid but empty JSON for every stream.

This reproduces on the latest v1.13.4 (ONNX Runtime 1.27.0), as well as v1.13.2 (ONNX Runtime 1.24.4).

### Steps to reproduce

1. Build the v1.13.4 `sherpa-onnx-offline` C++ binary for Linux x86-64 CPU.
2. Download and extract the official model:

   https://github.com/k2-fsa/sherpa-onnx/releases/download/asr-models/sherpa-onnx-nemo-parakeet-unified-en-0.6b-int8-non-streaming.tar.bz2

   Archive SHA-256: `99f63605b3a85a54c250c0869670a687b7d6598a47bf2421515e1f839a76e150`

3. From LibriSpeech `test-clean`, use these two public files:

   - Trigger: `LibriSpeech/test-clean/5142/36600/5142-36600-0000.flac` (2.525 s; reference: `CHAPTER SEVEN ON THE RACES OF MAN`)
   - Control: `LibriSpeech/test-clean/237/134500/237-134500-0006.flac` (4.225 s; reference: `JUST SMELL THE WILD ROSES THEY ARE ALWAYS SO SPICY AFTER A RAIN`)

4. Convert each losslessly to standard 16 kHz mono PCM16 WAV:

~~~bash
ffmpeg -nostdin -hide_banner -loglevel error -y \
  -i 5142-36600-0000.flac -map_metadata -1 -vn -sn -dn \
  -ar 16000 -ac 1 -c:a pcm_s16le -f wav 5142-36600-0000.wav

ffmpeg -nostdin -hide_banner -loglevel error -y \
  -i 237-134500-0006.flac -map_metadata -1 -vn -sn -dn \
  -ar 16000 -ac 1 -c:a pcm_s16le -f wav 237-134500-0006.wav
~~~

Prepared WAV SHA-256 values:

- `5142-36600-0000.wav`: `6d654971d0f491de3961a76af35153dfe30f695c4aedc43ef06dc19b75f42e87`
- `237-134500-0006.wav`: `877052d58f9f108f4a0c5e9300f04d6cf0df74bea6f8dbefc77aecc2a91b4da`

5. Define the common arguments (replace `MODEL` and `BIN` as needed):

~~~bash
MODEL=/path/to/sherpa-onnx-nemo-parakeet-unified-en-0.6b-int8-non-streaming
BIN=/path/to/sherpa-onnx-offline

common=(
  --num-threads=1
  --encoder="$MODEL/encoder.int8.onnx"
  --decoder="$MODEL/decoder.int8.onnx"
  --joiner="$MODEL/joiner.int8.onnx"
  --tokens="$MODEL/tokens.txt"
  --model-type=nemo_transducer
)
~~~

6. Decode the control alone:

~~~bash
"$BIN" "${common[@]}" 237-134500-0006.wav
~~~

It correctly returns:

~~~json
{"text":"Just smell the wild roses they are always so spicy after a rain"}
~~~

7. Decode the trigger and control together:

~~~bash
"$BIN" "${common[@]}" 5142-36600-0000.wav 237-134500-0006.wav
echo "$?"
~~~

Both outputs are empty and the exit status is 0:

~~~text
Creating recognizer ...
recognizer created in 1.712 s
Started
Done!

5142-36600-0000.wav
----
237-134500-0006.wav
----
num threads: 1
decoding method: greedy_search
Elapsed seconds: 0.951 s
Real time factor (RTF): 0.951 / 6.750 = 0.141
{"lang": "", "emotion": "", "event": "", "text": "", "timestamps": [], "durations": [], "tokens":[], "ys_log_probs": [], "words": []}
{"lang": "", "emotion": "", "event": "", "text": "", "timestamps": [], "durations": [], "tokens":[], "ys_log_probs": [], "words": []}
0
~~~

Reversing the WAV argument order also makes both outputs empty.

Minimal result matrix:

| Inputs | Result |
|---|---|
| Trigger alone | Empty result, exit 0 |
| Control alone | Correct transcript, exit 0 |
| Trigger, control | Both empty, exit 0 |
| Control, trigger | Both empty, exit 0 |
| Two non-trigger controls | Both correct, exit 0 |

The same matrix reproduces with `--num-threads=8`. It is unchanged between v1.13.2 and v1.13.4.

### Expected behavior

Each stream's result should be independent. Even if the trigger utterance legitimately decodes to empty, adding it to a batch should not change the control stream from a correct transcript to an empty result.

### Environment

- `sherpa-onnx` version: 1.13.4
- Binary version output: `sherpa-onnx Git SHA1: 753609d0`; `sherpa-onnx Git date: Tue Jul 7 19:00:41 2026`
- Source tag commit: `142807252687d81b40d6315f23470a1512a00de3`
- OS: Ubuntu 24.04 container on Linux 7.0.0-28-generic, x86-64
- CPU: Intel Core i7-1185G7
- Python version: N/A (C++ CLI)
- ONNX Runtime version: 1.27.0
- Provider: CPU
- Decoding method: `greedy_search`
- GPU/NPU: N/A

### Larger-run evidence

On the complete 2,620-utterance LibriSpeech `test-clean` split, an initial batched run lost 44 utterances in exactly four complete groups: every affected `sherpa-onnx-offline` process exited 0 and every result in that process was empty. Recursive bisection showed that a trigger stream could make otherwise-good sibling streams empty.

A public reproducibility report and successful workaround run are here:

- https://github.com/gracee3/native-asr/blob/main/docs/reproducibility-report.md#initial-full-split-test-clean-validation
- https://github.com/gracee3/native-asr/tree/main/benchmarks/published/2026-08-09-librispeech-test-clean-pair
- Workaround implementation: https://github.com/gracee3/native-asr/commit/44f2eafd6a6513617ead992714dab26c120b9befc

The current workaround recursively bisects an exit-zero all-empty batch, retries affected singletons with VAD, and uses lossless bounded chunks only if VAD is also empty. It recovers the corpus, but it adds many model loads and does not address the underlying batched behavior.

From source inspection, `OfflineRecognizerTransducerNemoImpl::DecodeStreams` pads/batches the feature matrices before `RunEncoder`, while the greedy decoder processes batch elements independently afterward. That makes the batched encoder/length path a possible area to inspect, but I have not isolated the root cause and do not want to propose a speculative fallback as the upstream fix.

Would you recommend particular encoder inputs/outputs to log to narrow this down, or is this a known constraint of the Unified model export?

_Disclosure: this report was prepared with help from OpenAI Codex; all reproductions and measurements above were run and verified on the reporter's host._

Eoin-Houstoun:
Fixes #3853.

`NemoNormalizePerFeature` computes per-feature variance as `E[x^2] - (E[x])^2` in float32. For the trigger utterance in #3853, one mel bin sits at the log floor near -15.94, so both terms are about 254.1 and their true difference (about 2.5e-5) is below float32 resolution at that magnitude. The subtraction cancels to exactly 0, `inv_std` becomes `1/(0 + 1e-5) = 1e5`, and that bin's small deviations normalize to +7515 while every other feature stays within +/-17.

That is also why one bad stream empties the whole batch: the padded batch goes through the INT8 encoder in one session, its per-tensor dynamic-quantization scale is computed over the outliers, the quant step grows to about 30, and all real features collapse into a single bucket. The decoder then emits blank for every stream.

The fix computes variance from centered values, `mean((x - mean)^2)`. Same semantics and epsilon, no cancellation. An exactly constant column now normalizes to 0 instead of exploding.

Tests: two new cases in `math-test.cc`. A floor-pinned column with one off-floor frame normalizes to 7510 before the fix and fails the test; with the fix it stays bounded. A constant column normalizes to 0.

End to end with the Parakeet Unified EN 0.6B INT8 model from the issue (LibriSpeech test-clean, trigger = 5142-36600-0000, control = 237-134500-0006):

| Case | v1.13.4 | this PR |
|---|---|---|
| trigger alone | "" | "CHAPTER seven ON THE RACES OF MAN" |
| control alone | correct | correct (unchanged) |
| [trigger, control] batch | both "" | both correct |
| [control, trigger] batch | both "" | both correct |
| max normalized feature (trigger, mel bin at log floor) | +7515 | within normal range |

`ComputeMeanAndInvStd` in the same file uses the same variance form. I can change it to the centered form here too if you want; I kept the diff minimal for now.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved feature normalization for near-constant inputs, reducing numerical instability.
  * Constant-valued features now normalize reliably to zero.

* **Tests**
  * Added coverage for near-constant and constant feature normalization scenarios.

### DIFF
--- a/sherpa-onnx/csrc/math-test.cc
+++ b/sherpa-onnx/csrc/math-test.cc
@@ -4,6 +4,7 @@
 
 #include "sherpa-onnx/csrc/math.h"
 
+#include <cmath>
 #include <vector>
 
 #include "gtest/gtest.h"
@@ -134,6 +135,39 @@ TEST(ComputeMeanAndInvStd, Case1) {
   for (int32_t i = 0; i < num_cols; ++i) {
     EXPECT_NEAR(mean[i], expected_mean[i], 1e-6f) << "at index " << i;
     EXPECT_NEAR(inv_std[i], expected_inv_std[i], 1e-6f) << "at index " << i;
+  }
+}
+
+TEST(NemoNormalizePerFeature, NearConstantFeatureStaysBounded) {
+  // A feature pinned at a constant log floor with one slightly-off frame,
+  // as produced by a mel bin with no energy. E[x^2] - E[x]^2 cancels to 0
+  // in float32 here, which used to normalize the off-floor frame to
+  // magnitudes in the thousands (7510 with these exact values).
+  int32_t num_frames = 1000;
+  int32_t feature_dim = 2;
+  std::vector<float> x(num_frames * feature_dim);
+  for (int32_t i = 0; i < num_frames; ++i) {
+    x[i * feature_dim] = -15.94f;
+    x[i * feature_dim + 1] = static_cast<float>(i % 7);
+  }
+  x[(num_frames / 2) * feature_dim] = -15.94f + 0.075f;
+
+  NemoNormalizePerFeature(x.data(), num_frames, feature_dim);
+
+  for (int32_t i = 0; i < num_frames; ++i) {
+    EXPECT_LE(std::fabs(x[i * feature_dim]), 100.0f) << "at frame " << i;
+  }
+}
+
+TEST(NemoNormalizePerFeature, ConstantFeatureNormalizesToZero) {
+  int32_t num_frames = 8;
+  int32_t feature_dim = 1;
+  std::vector<float> x(num_frames, 3.5f);
+
+  NemoNormalizePerFeature(x.data(), num_frames, feature_dim);
+
+  for (int32_t i = 0; i < num_frames; ++i) {
+    EXPECT_NEAR(x[i], 0.0f, 1e-6f) << "at frame " << i;
   }
 }
 
--- a/sherpa-onnx/csrc/math.cc
+++ b/sherpa-onnx/csrc/math.cc
@@ -137,8 +137,12 @@ void NemoNormalizePerFeature(float *p, int32_t num_frames,
   Eigen::Map<RowMajorMat> x(p, num_frames, feature_dim);
 
   Eigen::RowVectorXf mean = x.colwise().mean();
+
+  // E[x^2] - E[x]^2 cancels catastrophically in float32 when a feature
+  // stays near a constant value, yielding var == 0 and huge normalized
+  // outliers, so compute the variance from centered values instead.
   Eigen::RowVectorXf var =
-      (x.array().square().colwise().mean() - mean.array().square()).max(0.0f);
+      (x.rowwise() - mean).array().square().colwise().mean();
 
   Eigen::RowVectorXf inv_std = (var.array().sqrt() + 1e-5f).inverse();
 


sherpa-onnx/csrc/math-test.cc
```diff
@@ -137,4 +138,37 @@ TEST(ComputeMeanAndInvStd, Case1) {
   }
 }
 
+TEST(NemoNormalizePerFeature, NearConstantFeatureStaysBounded) {
+  // A feature pinned at a constant log floor with one slightly-off frame,
+  // as produced by a mel bin with no energy. E[x^2] - E[x]^2 cancels to 0
```
csukuangfj, line 143: > as produced by a mel bin with no energy. E[x^2] - E[x]^2 cancels to 0

Can you test that 0 is from `max(0.)`?
That is, E[x^2] - E[x]^2 becomes negative due to compuation?

csukuangfj (approved):

sherpa-onnx/csrc/math-test.cc
Eoin-Houstoun: Thanks for reviewing! @csukuangfj 

Checked this: yes. For the exact data in the new test (n=1000 at -15.94, one frame +0.075), the raw float32 E[x^2] - (E[x])^2 comes out at -8.07e-3, negative against a true variance of +5.6e-6, so the old max(0.0f) was clamping a negative difference to 0. It can also land positive but wrong by orders of magnitude on other data (+7.5e-4 where the true variance is 1.1e-5). 
The centered form is non-negative by construction, so the clamp is no longer load-bearing.

csukuangfj: Thank you for testing!

PR #3870, opened 2026-08-11
Expose the pyannote window shift ratio in the C API and language bindings

JulianPscheid:
Follow-up to #3769, which added `window_shift_ratio` to the C++ core and the CLI but deliberately left the bindings alone to stay reviewable. This exposes `OfflineSpeakerSegmentationPyannoteModelConfig::window_shift_ratio` through the C and CXX APIs and the language bindings that already surface that config, so the knob is reachable outside the CLI.

The C struct treats `0` and negative values as unset and maps them to `0.1`. C API callers and raw bindings commonly zero-initialize config structs, and passing a zero straight through would fail the native `(0, 1]` validation and break every existing caller. The sentinel lives in one place, the config conversion in `c-api.cc`, and `c-api.h` documents it. NaN is not treated as unset; it still reaches native validation and is rejected. Typed binding defaults are `0.1`.

For the standard `window_size=160000` model the default path still computes a 16000-sample shift. A C API probe returned 16000 samples from a zero-initialized field and 32000 samples at `0.2`. A build of untouched `upstream/master` and a build of this branch produced byte-identical segment output with the ratio unset (same ten segments, same SHA-256).

Two things worth calling out beyond the mechanical binding work. There is a new debug-only log line in `offline-speaker-segmentation-pyannote-model.cc` that reports the computed shift in samples, which is what makes propagation observable from a binding; it is inside the existing `config_.debug` guard and does not change normal output. And the WASM binding's struct packing moves from 4 to 8 bytes with its compile-time layout assertion updated to match. I could not compile that one locally, so it is worth a look in CI.

The benchmark that motivated the core option, for context: a 644 s two-speaker English WAV on macOS arm64, `0.10` at 40 s, `0.15` at 27 s (1.48x), `0.20` at 20 s (2.00x), with 99.89% and 99.93% frame-level label agreement against `0.10`.

| Check | Result |
|---|---|
| Release C++ core, CLI, C API, CXX API, C/CXX examples | Passed |
| Default output vs untouched `upstream/master` | Identical (same 10 segments and SHA-256) |
| C API zero / negative / `0.2` native shift | 16000 / 16000 / 32000 samples |
| JNI shared library | Passed |
| Python build and import | Passed |
| Java compile | Passed (existing deprecation warnings only) |
| Flutter library analyze | Passed |
| Dart example analyze | Passed with existing warnings and info |
| Go wrapper tests and example build | Passed |
| Node typecheck, native addon build, JS syntax | Passed |
| Swift typecheck | Passed |
| Kotlin and Android app | Not compiled locally; Kotlin compiler and API artifact unavailable |
| .NET | Not compiled locally; `dotnet` unavailable |
| Rust | Not compiled locally; Cargo unavailable |
| Pascal | Not compiled locally; FPC unavailable |
| WASM C++ | Not compiled locally; Emscripten unavailable; JS syntax passed |
| HarmonyOS ETS | Not compiled locally; Harmony toolchain unavailable; shared native bridge compiled through Node |

`git diff --check` passes and no generated artifacts are included.

## Summary by CodeRabbit

- **New Features**
  - Added configurable Pyannote segmentation window-shift ratio for offline speaker diarization.
  - The setting is available across supported APIs and platforms, with a default value of `0.1`.
  - Values can be adjusted to control segmentation window overlap.

- **Documentation**
  - Updated speaker diarization guidance and examples to demonstrate the new configuration option.

- **Bug Fixes**
  - Improved consistency by ensuring the configured window-shift ratio is applied throughout speaker diarization workflows.

### DIFF
--- a/android/SherpaOnnxSpeakerDiarization/app/src/main/java/com/k2fsa/sherpa/onnx/speaker/diarization/SpeakerDiarizationObject.kt
+++ b/android/SherpaOnnxSpeakerDiarization/app/src/main/java/com/k2fsa/sherpa/onnx/speaker/diarization/SpeakerDiarizationObject.kt
@@ -48,7 +48,8 @@ object SpeakerDiarizationObject {
             val config = OfflineSpeakerDiarizationConfig(
                 segmentation = OfflineSpeakerSegmentationModelConfig(
                     pyannote = OfflineSpeakerSegmentationPyannoteModelConfig(
-                        segmentationModel
+                        model = segmentationModel,
+                        windowShiftRatio = 0.1f,
                     ),
                     debug = true,
                 ),
--- a/c-api-examples/offline-speaker-diarization-c-api.c
+++ b/c-api-examples/offline-speaker-diarization-c-api.c
@@ -73,6 +73,8 @@ int main() {
   memset(&config, 0, sizeof(config));
 
   config.segmentation.pyannote.model = segmentation_model;
+  // A value of 0 also uses the default of 0.1.
+  config.segmentation.pyannote.window_shift_ratio = 0.1f;
   config.embedding.model = embedding_extractor_model;
 
   // the test wave ./0-four-speakers-zh.wav has 4 speakers, so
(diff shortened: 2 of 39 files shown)


csukuangfj (approved):

PR #3872, opened 2026-08-12
Avoid bundling the web WASM assets in native Flutter apps

### LINKED ISSUE #3871: Native Flutter builds bundle the sherpa_onnx_web WASM assets
A native-only Flutter app that depends on `sherpa_onnx: ^1.13.5` ships 15,084,204 bytes of `sherpa_onnx_web` assets it can never execute. Flutter 3.44.8, macOS.

```console
flutter create --platforms=android,ios,macos,windows,linux native_asset_probe
cd native_asset_probe
flutter pub add sherpa_onnx:^1.13.5
flutter build bundle
find build/flutter_assets/packages/sherpa_onnx_web -type f -exec stat -f '%z %N' {} \;
```

That is the 14,865,876-byte WASM runtime, the 93,039-byte Emscripten loader, and seven JS wrappers totaling 125,289 bytes. Summed file sizes; the whole bundle is 70,656,531 bytes.

`sherpa_onnx` depends on `sherpa_onnx_web` unconditionally, and Flutter copies package-declared assets without filtering by build target, so Android, iOS, macOS, Linux and Windows all get them.

One way out is to make `sherpa_onnx_web` a non-endorsed web implementation: drop it from `sherpa_onnx`'s dependencies, have web apps depend on it directly, and let its generated plugin registration install the WASM loader behind a channel that `initBindingsAsync()` calls. An app that misses the dependency then fails immediately with an error naming the package to add. I have this working locally: the same probe drops to 55,571,027 bytes with no `packages/sherpa_onnx_web` directory, and the web hello_world example still loads the WASM and returns version info in Chrome.

Happy to send it as a PR if the tradeoff looks right to you.

JulianPscheid:
Fixes #3871.

`sherpa_onnx` depends on `sherpa_onnx_web` unconditionally, and Flutter copies package-declared assets regardless of build target, so every native app carries the ~15 MB web WASM payload.

This makes `sherpa_onnx_web` a non-endorsed web implementation. It is removed from `sherpa_onnx`'s dependencies and from its `web: default_package` entry; web apps add it directly. `sherpa_onnx_web` declares `implements: sherpa_onnx`, and its `registerWith` now serves the existing idempotent `loadWasm()` over a method channel that `initBindingsAsync()` invokes on web. A missing package surfaces as a `StateError` naming `sherpa_onnx_web` and the `flutter pub get` step instead of a `MissingPluginException`. The web-capable examples get the explicit dependency.

Native behavior is unchanged: `initBindings()` is untouched, and the non-web branch of `initBindingsAsync()` still calls `initNativeBindings` directly.

I considered a shared Dart hook owned by `sherpa_onnx` instead of a channel, but that needs the web package to depend back on `sherpa_onnx`, or a new platform-interface package. The channel uses the plugin boundary that already exists and keeps the two published packages independent.

One caveat: an app targeting both web and native from a single pubspec still gets the assets in its native builds, since it has to depend on the web package. The saving is for native-only dependency graphs.

Measured on a fresh `flutter create` probe (Flutter 3.44.8, macOS): `flutter_assets` fell from 70,656,531 to 55,571,027 bytes and no longer contains `packages/sherpa_onnx_web`.

Testing:
- `flutter analyze` in `flutter/sherpa_onnx`, and in `sherpa_onnx_web` with the release-generated assets present
- `flutter build bundle` in a fresh native-only app against the patched path dependency
- `flutter build web` for `flutter-examples/hello_world` against both patched packages
- In Chrome: two concurrent `initBindingsAsync()` calls completed, and `getVersion()`, `getGitSha1()`, `getGitDate()` and `getOnnxruntimeVersion()` returned values
- In Chrome without the web package: `initBindingsAsync()` reported that `sherpa_onnx_web` must be added to `pubspec.yaml`

Versioning for the two packages is left to you, since publication order matters here.

## Summary by CodeRabbit

- **New Features**
  - Added improved Flutter web support for loading the required WebAssembly resources.
  - Flutter web applications can now explicitly integrate the web plugin for runtime initialization.

- **Documentation**
  - Updated setup guidance to require both the core and web packages as direct dependencies.
  - Refreshed Flutter examples and dependency versions to support web builds.

- **Bug Fixes**
  - Added clearer runtime errors when the web plugin is missing or not registered.

### DIFF
--- a/flutter-examples/hello_world/README.md
+++ b/flutter-examples/hello_world/README.md
@@ -19,7 +19,8 @@ Edit `pubspec.yaml`, replace the `dependencies:` section with:
 dependencies:
   flutter:
     sdk: flutter
-  sherpa_onnx: ^1.13.4
+  sherpa_onnx: ^1.13.5
+  sherpa_onnx_web: ^1.13.5 # Required only when building for web.
 ```
 
 Then run:
--- a/flutter-examples/hello_world/pubspec.yaml
+++ b/flutter-examples/hello_world/pubspec.yaml
@@ -16,6 +16,9 @@ dependencies:
   sherpa_onnx: ^1.13.5
   # sherpa_onnx:
   #   path: ../../flutter/sherpa_onnx
+  sherpa_onnx_web: ^1.13.5
+  # sherpa_onnx_web:
+  #   path: ../../flutter/sherpa_onnx_web
 
 dev_dependencies:
   flutter_test:
--- a/flutter-examples/offline-punctuation/pubspec.yaml
+++ b/flutter-examples/offline-punctuation/pubspec.yaml
@@ -20,6 +20,9 @@ dependencies:
   sherpa_onnx: ^1.13.5
   # sherpa_onnx:
   #   path: ../../flutter/sherpa_onnx
+  sherpa_onnx_web: ^1.13.5
+  # sherpa_onnx_web:
+  #   path: ../../flutter/sherpa_onnx_web
   url_launcher: 6.2.6
   url_launcher_linux: 3.1.0
 
--- a/flutter-examples/online-punctuation/pubspec.yaml
+++ b/flutter-examples/online-punctuation/pubspec.yaml
@@ -20,6 +20,9 @@ dependencies:
   sherpa_onnx: ^1.13.5
   # sherpa_onnx:
   #   path: ../../flutter/sherpa_onnx
+  sherpa_onnx_web: ^1.13.5
+  # sherpa_onnx_web:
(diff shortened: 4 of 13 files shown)


JulianPscheid:
#3874 covers this better than this PR does. Per-asset `platforms:` filtering keeps the package endorsed and needs no API change, which is the right shape.

I checked that it does what's needed: in a fresh app, a local package declaring one asset with `platforms: [web]` and one without ships only the unfiltered one in `flutter build bundle`. So the WASM stays out of native bundles.

Closing in favor of #3874.

Issue #3880, opened 2026-08-18
[FEATURE] Compute per-segment confidence value for OfflineSpeakerDiarization


matiaslin:
### Describe the feature

When performing offline speaker diarization, we will add an option to compute the [**silhouette coefficient**](https://www.geeksforgeeks.org/machine-learning/what-is-silhouette-score/) against the final cluster assignment as the per-segment confidence value.

Normally, we would use standard Euclidian distance to calculate the silhouette coefficient. But since we already calculated cosine dissimilarity distance in `fast-clustering.cc`, see: https://github.com/k2-fsa/sherpa-onnx/blob/1cb484af5e69d3c7803c1eb0b3b5ab8041e0e911/sherpa-onnx/csrc/fast-clustering.cc#L35-L49
Reusing that matrix will have an added cost of $O(n^2)$.

To preserve the legacy `OfflineSpeakerDiarization` output format, we will only print the confidence levels if `clustering.compute-confidence` is set to true.

### Motivation / Use Case

Diarization is not a solved problem. Depending on the input audio and the parameters being set, it is not uncommon to see the wrong cluster assignment for certain segments. The confidence score will better inform the users how to interpret the returned diarizations.

### Additional context

As an example, the following command:
```bash
./sherpa-onnx-offline-speaker-diarization \
    --segmentation.pyannote-model=SEGMENTATION_MODEL \
    --embedding.model=EMBEDDING_MODEL \
    --clustering.compute-confidence=1 \
    INPUT_WAV
```
will yield:
```bash
...
12.890 -- 14.223 speaker_02 confidence=0.219
15.269 -- 16.501 speaker_02 confidence=0.231
17.092 -- 18.779 speaker_02 confidence=0.240
18.779 -- 22.205 speaker_03 confidence=0.834
19.572 -- 20.045 speaker_02 confidence=0.209
22.390 -- 24.550 speaker_04 confidence=0.162
26.137 -- 26.272 speaker_04 confidence=0.117
26.272 -- 27.115 speaker_01 confidence=0.207
27.858 -- 28.702 speaker_04 confidence=0.181
32.431 -- 33.038 speaker_04 confidence=0.225
34.085 -- 34.405 speaker_04 confidence=0.256
39.063 -- 39.653 speaker_04 confidence=0.296
46.673 -- 47.602 speaker_01 confidence=0.258
47.905 -- 48.749 speaker_01 confidence=0.269
[...]

PR #3881, opened 2026-08-19
Add per-segment confidence computation for OfflineSpeakerDiarization

matiaslin:
Motivating issue: #3880 

When performing offline speaker diarization, we add an option to toggle the computation of the [silhouette coefficient](https://www.geeksforgeeks.org/machine-learning/what-is-silhouette-score/) against the final cluster assignment as the per-segment confidence value. The added cost is $O(n^2)$, since we can reuse the cosine dissimilarity matrix already calculated in `fast-clustering.cc` (i.e. no asymptotic regression).

The option was added to `SherpaOnnxFastClusteringConfig`. This feature is backwards compatible:
- We toggle this enhancement via `clustering.compute-confidence` (disabled by default). Example:
```bash
./sherpa-onnx-offline-speaker-diarization \
    --segmentation.pyannote-model=SEGMENTATION_MODEL \
    --embedding.model=EMBEDDING_MODEL \
    --clustering.compute-confidence=1 \
    INPUT_WAV
```
(See original [issue](https://github.com/k2-fsa/sherpa-onnx/issues/3880) for an example output).

Confidence values:
- `confidence=[-1,1]`: the mean of the silhouette coefficients of the embedding intervals overlapping a given segment.
- `confidence=-2.0f` (printed as `n/a`):
  - no embedding window for this segment's speaker overaps its time range (e.g. short-overlapping split segments).
  - clustering only produced one cluster (i.e. we are unable to calculate the coefficient).
  - if for some reason `clustering.compute-confidence` is disabled, but a confidence diarization output is requested (via `ToString(with_confidence=true)`.
- `confidence=0.0`: singleton clusters (per convention).

A minimal set of unit tests has been added to `fast-clustering-test.cc`

Supported bindings: 
- C, Python and C++
- Rust, Pascal, Dart, C#, Swift, WASM (just mirrored the struct to keep ABI in layouts synced (higher-level wrapper is deferred to future PRs).

## Summary by CodeRabbit

- **New Features**
  - Added optional confidence scoring for offline speaker-diarization segments.
  - Confidence values are available through C++ and C APIs and align with start-time-sorted segments.
  - Added controls to enable or disable confidence computation.
  - Diarization output can display confidence values when available; unavailable values are reported as `NaN`.

- **Bug Fixes**
  - Improved consistency by sorting diarization segments by start time before returning results.

- **Tests**
  - Added coverage for cluster-quality scores, including well-separated, singleton, and single-cluster cases.

### DIFF
--- a/flutter/sherpa_onnx/lib/src/sherpa_onnx_bindings.dart
+++ b/flutter/sherpa_onnx/lib/src/sherpa_onnx_bindings.dart
@@ -66,6 +66,9 @@ final class SherpaOnnxOfflineSpeakerDiarizationSegment extends Struct {
 
   @Int32()
   external int speaker;
+
+  @Float()
+  external double confidence;
 }
 
 final class SherpaOnnxOfflineSpeakerSegmentationPyannoteModelConfig
@@ -94,6 +97,9 @@ final class SherpaOnnxFastClusteringConfig extends Struct {
 
   @Float()
   external double threshold;
+
+  @Int32()
+  external int computeConfidence;
 }
 
 final class SherpaOnnxOfflineSpeakerDiarizationConfig extends Struct {
--- a/sherpa-onnx/c-api/c-api.cc
+++ b/sherpa-onnx/c-api/c-api.cc
@@ -3178,6 +3178,9 @@ GetOfflineSpeakerDiarizationConfig(
   sd_config.clustering.threshold =
       SHERPA_ONNX_OR(config->clustering.threshold, 0.5);
 
+  sd_config.clustering.compute_confidence =
+      config->clustering.compute_confidence != 0;
+
   sd_config.min_duration_on = SHERPA_ONNX_OR(config->min_duration_on, 0.3);
 
   sd_config.min_duration_off = SHERPA_ONNX_OR(config->min_duration_off, 0.5);
@@ -3234,6 +3237,9 @@ void SherpaOnnxOfflineSpeakerDiarizationSetConfig(
   sd_config.clustering.threshold =
       SHERPA_ONNX_OR(config->clustering.threshold, 0.5);
 
+  sd_config.clustering.compute_confidence =
+      config->clustering.compute_confidence != 0;
+
   sd->impl->SetConfig(sd_config);
 }
 
@@ -3266,6 +3272,7 @@ SherpaOnnxOfflineSpeakerDiarizationResultSortByStartTime(
     ans[i].start = s.Start();
     ans[i].end = s.End();
     ans[i].speaker = s.Speaker();
+    ans[i].confidence = s.Confidence();
   }
 
   return ans;
--- a/sherpa-onnx/c-api/c-api.h
+++ b/sherpa-onnx/c-api/c-api.h
@@ -3875,6 +3875,8 @@ typedef struct SherpaOnnxFastClusteringConfig {
   int32_t num_clusters;
   /** Distance threshold used when the number of speakers is unknown. */
   float threshold;
+  /** When non-zero, per-segment confidence values are computed. */
+  int32_t compute_confidence;
 } SherpaOnnxFastClusteringConfig;
 
 /**
@@ -3965,6 +3967,12 @@ typedef struct SherpaOnnxOfflineSpeakerDiarizationSegment {
   float end;
   /** Speaker label, typically an integer cluster ID. */
   int32_t speaker;
+  /**
+   * Per-segment confidence in [-1, 1] (higher is more confident), or -2 if
+   * unavailable (compute_confidence disabled, or the score could not be
+   * computed for this segment).
+   */
+  float confidence;
 } SherpaOnnxOfflineSpeakerDiarizationSegment;
 
 /**
--- a/sherpa-onnx/csrc/fast-clustering-config.cc
+++ b/sherpa-onnx/csrc/fast-clustering-config.cc
@@ -15,7 +15,8 @@ std::string FastClusteringConfig::ToString() const {
 
   os << "FastClusteringConfig(";
   os << "num_clusters=" << num_clusters << ", ";
-  os << "threshold=" << threshold << ")";
+  os << "threshold=" << threshold << ", ";
+  os << "compute_confidence=" << (compute_confidence ? "True" : "False") << ")";
 
   return os.str();
 }
@@ -31,6 +32,10 @@ void FastClusteringConfig::Register(ParseOptions *po) {
                "If num_clusters is not specified, then it specifies the "
                "distance threshold for clustering. smaller value -> more "
                "clusters. larger value -> fewer clusters");
+
+  po->Register("compute-confidence", &compute_confidence,
+               "True to compute confidence value for each clustered "
+               "embedding and surface for every diarization segment.");
 }
 
 bool FastClusteringConfig::Validate() const {
--- a/sherpa-onnx/csrc/fast-clustering-config.h
+++ b/sherpa-onnx/csrc/fast-clustering-config.h
@@ -24,10 +24,17 @@ struct FastClusteringConfig {
   // The larger, the fewer clusters it will generate.
   float threshold = 0.5;
 
+  // When true, we compute the confidence value for every clustered
+  // embedding and surface it on each returned diarization segment.
+  bool compute_confidence = false;
+
   FastClusteringConfig() = default;
 
-  FastClusteringConfig(int32_t num_clusters, float threshold)
-      : num_clusters(num_clusters), threshold(threshold) {}
+  FastClusteringConfig(int32_t num_clusters, float threshold,
+                       bool compute_confidence = false)
+      : num_clusters(num_clusters),
+        threshold(threshold),
+        compute_confidence(compute_confidence) {}
 
   std::string ToString() const;
 
--- a/sherpa-onnx/csrc/fast-clustering.cc
+++ b/sherpa-onnx/csrc/fast-clustering.cc
@@ -4,6 +4,11 @@
 
 #include "sherpa-onnx/csrc/fast-clustering.h"
 
+#include <algorithm>
+#include <cassert>
+#include <cmath>
+#include <cstddef>
+#include <limits>
 #include <vector>
 
 #include "Eigen/Dense"
@@ -16,12 +21,20 @@ class FastClustering::Impl {
   explicit Impl(const FastClusteringConfig &config) : config_(config) {}
 
   std::vector<int32_t> Cluster(float *features, int32_t num_rows,
-                               int32_t num_cols) const {
+                               int32_t num_cols,
+                               std::vector<float> *silhouettes) const {
+    if (silhouettes) {
+      silhouettes->clear();
+    }
+
     if (num_rows <= 0) {
       return {};
     }
 
     if (num_rows == 1) {
+      if (silhouettes) {
+        silhouettes->assign(1, 0.0f);
+      }
       return {0};
     }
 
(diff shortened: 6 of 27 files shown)


sherpa-onnx/csrc/offline-speaker-diarization-result.cc
```diff
@@ -42,19 +43,30 @@ OfflineSpeakerDiarizationSegment::Merge(
...
   if (end_ < other.start_ && end_ + gap >= other.start_) {
-    return OfflineSpeakerDiarizationSegment(start_, other.end_, speaker_);
+    OfflineSpeakerDiarizationSegment merged(start_, other.end_, speaker_);
+    merged.SetConfidence(confidence_);
+    return merged;
   } else if (other.end_ < start_ && other.end_ + gap >= start_) {
-    return OfflineSpeakerDiarizationSegment(other.start_, end_, speaker_);
+    OfflineSpeakerDiarizationSegment merged(other.start_, end_, speaker_);
+    merged.SetConfidence(confidence_);
+    return merged;
```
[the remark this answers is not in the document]
matiaslin: Well, I guess the `SetConfidence` is not really needed here. I've initially added it for completion sake.

Since, we **always** want to compute the confidence after merging, both `confidence_` and `other.confidence_` would have been the same (i.e. `NaN`) at this point. 

Either way, I see why it can be misleading to have that call here. I'll remove it.

sherpa-onnx/csrc/offline-speaker-diarization-pyannote-impl.h
```diff
@@ -185,8 +197,10 @@ class OfflineSpeakerDiarizationPyannoteImpl
     }
 
     Timer clustering_timer(config_.segmentation.debug);
+    std::vector<float> silhouettes;
     std::vector<int32_t> cluster_labels = clustering_->Cluster(
-        &embeddings(0, 0), embeddings.rows(), embeddings.cols());
+        &embeddings(0, 0), embeddings.rows(), embeddings.cols(),
+        config_.clustering.compute_confidence ? &silhouettes : nullptr);
```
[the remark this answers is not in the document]
matiaslin: Mathematically, we cannot compute the silhouette coefficient when we only have one cluster, so it's right to leave it as `NaN`. I've updated the single cluster branch in `fast-clustering.cc:145` to return `NaN` instead of `0.0`.

The special case where we return 0.0 is when we have singleton clusters (convention).

Updated the PR description to accurately reflect this.

sherpa-onnx/c-api/c-api.h
```diff
@@ -4012,6 +4026,36 @@ SherpaOnnxOfflineSpeakerDiarizationResultSortByStartTime(
...
+ * SherpaOnnxOfflineSpeakerDiarizationResultSortByStartTime().
+ *
+ * NaN confidence values indicate confidence was not available (e.g. only one
+ * cluster).
+ *
+ * @param r A pointer returned by one of the
+ *          SherpaOnnxOfflineSpeakerDiarizationProcess*() functions.
+ * @return A newly allocated float array. Free it with
+ *         SherpaOnnxOfflineSpeakerDiarizationDestroySegmentConfidences().
+ */
```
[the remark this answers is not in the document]
matiaslin: Done in the latest patch. No need to mention the destroy function accepting `NULL` because every other destroy function does the same.

sherpa-onnx/c-api/c-api.cc
```diff
@@ -3237,6 +3252,11 @@ void SherpaOnnxOfflineSpeakerDiarizationSetConfig(
   sd->impl->SetConfig(sd_config);
 }
 
+void SherpaOnnxOfflineSpeakerDiarizationSetComputeConfidence(
+    const SherpaOnnxOfflineSpeakerDiarization *sd, int32_t enable) {
+  sd->impl->SetComputeConfidence(enable != 0);
+}
```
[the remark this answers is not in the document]
matiaslin: Ops, I missed those `#else` statements. Pushed the fix in the latest patch.

sherpa-onnx/c-api/c-api.h
```diff
@@ -4012,6 +4026,36 @@ SherpaOnnxOfflineSpeakerDiarizationResultSortByStartTime(
...
+ * aligned with the order produced by
+ * SherpaOnnxOfflineSpeakerDiarizationResultSortByStartTime().
+ *
+ * NaN confidence values indicate confidence was not available (e.g. only one
+ * cluster).
+ *
+ * @param r A pointer returned by one of the
+ *          SherpaOnnxOfflineSpeakerDiarizationProcess*() functions.
+ * @return A newly allocated float array, or NULL if there are no segments.
+ *         Free it with SherpaOnnxOfflineSpeakerDiarizationDestroySegmentConfidences().
```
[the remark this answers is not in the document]
matiaslin: I wanted to leave the docstring cofidence level agnostic as much as possible, since a future programmer might want to implement a different confidence value. I've revised the docstring in the latest patch.

sherpa-onnx/c-api/c-api.cc
```diff
@@ -3140,6 +3140,21 @@ struct SherpaOnnxOfflineSpeakerDiarization {
...
+      sorted_segments_cache;
+  mutable bool sorted_cached = false;
+
+  const std::vector<sherpa_onnx::OfflineSpeakerDiarizationSegment> &
+  SortedSegments() const {
+    if (!sorted_cached) {
+      sorted_segments_cache = impl.SortByStartTime();
+      sorted_cached = true;
+    }
+    return sorted_segments_cache;
```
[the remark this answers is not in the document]
matiaslin: I originally opted to have this sorted cache to avoid a double sorting when calling `SherpaOnnxOfflineSpeakerDiarizationResultGetSegmentConfidences`. But in doing so, I'll be adding more complexity to this function that might be unnecessary.

Since, this feature is opt-in, I would prefer the "double sorting" over adding extra complexity to `SherpaOnnxOfflineSpeakerDiarizationResultSortByStartTime`.

Changed in the lastest patch.

matiaslin:
CI passed, ready for review. Thank you.

matiaslin:
Btw, this feature touches 17 files partly because of the toggle to compute or not compute the confidence scores. 

If we decide to always compute the confidence, I can reduce the number of files. As I mentioned before, that comes with a cost that isn't significant, but there will be a slight change of the default performance/behavior.

Personally, I like having the toggle to keep with the legacy behavior of offline speaker diarization, but @csukuangfj, let me know if you have a preference one over the other. Thank you again.

matiaslin:
CI fails are unrelated, a retrigger would probably make them pass. Thanks.

sherpa-onnx/c-api/c-api.h
```diff
@@ -3951,6 +3951,20 @@ SHERPA_ONNX_API void SherpaOnnxOfflineSpeakerDiarizationSetConfig(
...
+ *
+ * When enable is non-zero, subsequent calls to the Process functions will
+ * compute a confidence value for every returned segment. Those values can be
+ * retrieved via SherpaOnnxOfflineSpeakerDiarizationResultGetSegmentConfidences().
+ * Disabled by default.
+ *
+ * @param sd A pointer returned by SherpaOnnxCreateOfflineSpeakerDiarization().
+ * @param enable Non-zero to enable, zero to disable.
+ */
+SHERPA_ONNX_API void SherpaOnnxOfflineSpeakerDiarizationSetComputeConfidence(
```
csukuangfj: Instead of introducing a new function, can you reuse 
https://github.com/k2-fsa/sherpa-onnx/blob/bbd946415d7791a878792790fc94bfcccf2a5ce3/sherpa-onnx/c-api/c-api.h#L3950

---

sherpa-onnx supports 12 programming languages at present. Adding a new public function to `c-api.h` means we need to do that for the remaining languages. In general, we try to use existing APIs.

sherpa-onnx/csrc/fast-clustering-config.h
```diff
@@ -24,6 +24,10 @@ struct FastClusteringConfig {
   // The larger, the fewer clusters it will generate.
   float threshold = 0.5;
 
+  // When non-zero, we compute the confidence value for every clustered
+  // embedding and surface it on each returned diarization segment.
+  int32_t compute_confidence = 0;
```
csukuangfj: Can you use
```
bool compute_confidence = false;
```

In c-api.h, we use int32_t to replace bool since there is no bool in C. In C++, we usually use bool for flag variables.

sherpa-onnx/csrc/fast-clustering-config.cc
```diff
@@ -15,7 +15,8 @@ std::string FastClusteringConfig::ToString() const {
 
   os << "FastClusteringConfig(";
   os << "num_clusters=" << num_clusters << ", ";
-  os << "threshold=" << threshold << ")";
+  os << "threshold=" << threshold << ", ";
+  os << "compute_confidence=" << compute_confidence << ")";
```
csukuangfj: Please print `True` or `False` once you finish changing it to a bool variable, as we do for other bool variables.

sherpa-onnx/csrc/fast-clustering-config.h
csukuangfj: Also, please change the constructor to initialize it.
And change `python/csrc/faster-cluastering.cc` accordingly.

sherpa-onnx/csrc/fast-clustering.cc
```diff
@@ -64,10 +77,88 @@ class FastClustering::Impl {
...
+        int32_t neighbor_count = count[base + cluster_index];
+        if (neighbor_count == 0) continue;
+        double mean = sum[base + cluster_index] / neighbor_count;
+        if (mean < b) b = mean;
+      }
+
+      if (!std::isfinite(b)) {
+        // We only have one cluster overall. We cannot calculate
+        // the silhouette coefficient, so we return NaN.
+        (*silhouettes)[row_index] = std::numeric_limits<float>::quiet_NaN();
```
csukuangfj: Valid silhouette values are in `[-1, 1]`. Can you use a value outside that range, say, -2, to replace `nan`?

csukuangfj (changes requested):

sherpa-onnx/csrc/fast-clustering.cc
matiaslin: I originally wanted to use `nan` to keep the implementation of the confidence computation agnostic. So that in the future if we decide there is a better confidenc value, we can switch out the implementation without having to touch the public API.

That being said, I'm not opposed to using a finite number, like -2, as an unavailable sentinel value. This would just mean that if we ever switch the confidence computing implementation, it would have to be bounded by similar limits as the silhouette coefficient (which any normalized metric should conform to either way). So, we should be relatively safe to change the default value to `nan`.

Updated in the next patch.

sherpa-onnx/csrc/fast-clustering-config.cc
matiaslin: Sounds good. Done in the next patch.

sherpa-onnx/csrc/fast-clustering-config.h
matiaslin: That makes sense. Ok, changed in the next patch.

sherpa-onnx/c-api/c-api.h
matiaslin: I originally was going with this approach, since it was more intuitive to me. But I eventually opted for the new APIs, because touching `SherpaOnnxFastClusteringConfig` would mean breaking the manually mirrored bindgins (namely: C#, rust, dart, wasm), and I wasn't sure how comfortable we were with that decision.

However, if you are ok with it, I'm also more inclined to use the existing APIs too :). This approach would mean adjusting 6 languages vs 12 bindings. We can always add support to the other bindings in future PRs. I made the refactoring in the next patch.

sherpa-onnx/pascal-api/sherpa_onnx.pas
```diff
@@ -1275,6 +1275,7 @@   SherpaOnnxOfflineSpeakerSegmentationModelConfig = record
   SherpaOnnxFastClusteringConfig = record
     NumClusters: cint32;
     Threshold: cfloat;
+    ComputeConfidence: cint32;
```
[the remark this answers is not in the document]
matiaslin: As mentioned in the PR description, the higher-level wrapper are deferred to a different PR.

sherpa-onnx/rust/sherpa-onnx/src/offline_speaker_diarization.rs
```diff
@@ -91,6 +91,7 @@ impl FastClusteringConfig {
         sys::FastClusteringConfig {
             num_clusters: self.num_clusters,
             threshold: self.threshold,
+            compute_confidence: 0,
```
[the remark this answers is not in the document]
matiaslin: As mentioned in the PR description, the higher-level wrapper are deferred to a different PR.

sherpa-onnx/c-api/c-api.h
```diff
@@ -3875,6 +3875,8 @@ typedef struct SherpaOnnxFastClusteringConfig {
   int32_t num_clusters;
   /** Distance threshold used when the number of speakers is unknown. */
   float threshold;
+  /** When non-zero, per-segment confidence values are computed. */
+  int32_t compute_confidence;
```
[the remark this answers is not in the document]
matiaslin: As discussed in https://github.com/k2-fsa/sherpa-onnx/pull/3881#discussion_r3892452119. We prefer the current approach over adding new pub APIs.

matiaslin:
Thank you for the review. I've applied the fixes, and it's ready for review again.

matiaslin:
Uploaded dummy commit to retrigger CI (it seemed to have timed out in the previous patch).

flutter/sherpa_onnx/lib/src/sherpa_onnx_bindings.dart
```diff
@@ -94,6 +97,9 @@ final class SherpaOnnxFastClusteringConfig extends Struct {
 
   @Float()
   external double threshold;
+
+  @Int32()
+  external int computeConfidence;
```
csukuangfj, line 102: Please also update 
https://github.com/k2-fsa/sherpa-onnx/blob/7545823449009d41f36c43466af02fe54531ca15/flutter/sherpa_onnx/lib/src/offline_speaker_diarization_config.dart#L115

I suggest searching for `SherpaOnnxFastClusteringConfig` in dart and changing every place it is used.

flutter/sherpa_onnx/lib/src/sherpa_onnx_bindings.dart
```diff
@@ -66,6 +66,9 @@ final class SherpaOnnxOfflineSpeakerDiarizationSegment extends Struct {
 
   @Int32()
   external int speaker;
+
+  @Float()
```
csukuangfj, line 70: If you change 3 files in `scripts/dotnet`, you should also change the other files in dart, not this single file. 

If you are using AI, please ask AI to change all places that need to be updated.

sherpa-onnx/pascal-api/sherpa_onnx.pas
csukuangfj, line 1280: Please search for where `NumClusters` and `Threshold` are used, and do the same for the newly added field `ComputeConfidence`.

swift-api-examples/SherpaOnnx.swift
```diff
@@ -1780,10 +1780,14 @@ public func sherpaOnnxOfflineSpeakerSegmentationModelConfig(
   )
 }
 
-public func sherpaOnnxFastClusteringConfig(numClusters: Int = -1, threshold: Float = 0.5)
+public func sherpaOnnxFastClusteringConfig(
+  numClusters: Int = -1, threshold: Float = 0.5
+)
```
csukuangfj, line 1785: The function argument should include the newly added `computeConfidence`.

csukuangfj (changes requested):

flutter/sherpa_onnx/lib/src/sherpa_onnx_bindings.dart
matiaslin: Done in the next patch.

flutter/sherpa_onnx/lib/src/sherpa_onnx_bindings.dart
matiaslin: Sounds good. Done in the next patch.

swift-api-examples/SherpaOnnx.swift
matiaslin: Added on the next patch.

sherpa-onnx/pascal-api/sherpa_onnx.pas
matiaslin: Got it. Done in the next patch.

matiaslin (commented):
Thank you again for the review, @csukuangfj! I remember from my previous contributions that we prefer to touch the fewest bindings possible in one PR to make it more reviewable. That's why in the previous patch, I had the mental concept of splitting the bindings changes required into two: the must-have for it to compile, and the binding wrapper layer.

I've done the first part (the must-have for compilation) in the previous patch thinking that I could delegate the second part (wrapper layer) for those languages in subsequent PRs. But after some thought last night, I would agree with you that having a half-wired confidence value in fast clustering config is not the best option. Cool, I'll keep this in mind for future changes.

I made the requested changes, let me know if there is anything else I can do/improve on.

swift-api-examples/SherpaOnnx.swift
```diff
@@ -1780,10 +1780,14 @@ public func sherpaOnnxOfflineSpeakerSegmentationModelConfig(
   )
 }
 
-public func sherpaOnnxFastClusteringConfig(numClusters: Int = -1, threshold: Float = 0.5)
+public func sherpaOnnxFastClusteringConfig(
+  numClusters: Int = -1, threshold: Float = 0.5, computeConfidence: Int = 0
```
csukuangfj: `computeConfidence` is a bool, please follow
https://github.com/k2-fsa/sherpa-onnx/blob/7545823449009d41f36c43466af02fe54531ca15/swift-api-examples/SherpaOnnx.swift#L168

csukuangfj (commented):
Thanks! Left a minor comment.

swift-api-examples/SherpaOnnx.swift
matiaslin: Sounds good, I was following the pattern of:
https://github.com/k2-fsa/sherpa-onnx/blob/7545823449009d41f36c43466af02fe54531ca15/swift-api-examples/SherpaOnnx.swift#L109

But yeah, most of the other flags are `Bool`, changed it in the next patch.

csukuangfj (approved):

matiaslin:
Thank you for reviewing and merging!

PR #3307, opened 2026-03-12
feat: add generic SetOption/GetOption to OnlineStream/OfflineStream (C++ core)

ZhaoChaoqun:
## Summary

Ref #3101 (Part 1a)

Add a generic key-value option mechanism to both `OnlineStream` and `OfflineStream` C++ core classes. This is the foundation for the SetOption API — follow-up PRs add C API, CXX wrapper, language bindings, and Paraformer migration.

## New C++ Methods

```cpp
// OnlineStream & OfflineStream
void SetOption(const std::string &key, const std::string &value);
bool HasOption(const std::string &key) const;
const std::string &GetOption(const std::string &key) const;
int32_t GetOptionInt(const std::string &key, int32_t default_value = 0) const;
float GetOptionFloat(const std::string &key, float default_value = 0.0f) const;
```

## Design

- Options stored as `std::unordered_map<std::string, std::string>` in Pimpl `Impl`
- `GetOption` returns empty string for missing keys (no exceptions)
- Existing APIs are untouched — no breaking changes

## Files Changed (4 files, +138 lines)

| File | Change |
|------|--------|
| `online-stream.h` | Add method declarations, `#include <string>` |
| `online-stream.cc` | Add `options_` map and implementations |
| `offline-stream.h` | Add method declarations |
| `offline-stream.cc` | Add `options_` map and implementations |

## Test Plan

- [ ] Build on macOS/Linux — verify compilation passes
- [ ] Existing tests still pass (no breaking changes)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added configurable per-stream options with a new public API for both offline and online streams. Users can set and retrieve string-based key-value options with typed accessors (string, integer, float) and default fallback values.

PR #3891, opened 2026-08-24
Support per-stream runtime options (setOption/getOption/hasOption) in the JavaScript (node-addon) API for streaming ASR

josephomills:
## What this PR does

The C API has exported `SherpaOnnxOnlineStreamSetOption` / `GetOption` / `HasOption` since the SetOption series (#3307–#3315), and the streaming NeMo (Nemotron) recognizer reads the per-stream option `"language"` on every decode call (`GetLanguagePromptIds` in `online-recognizer-transducer-nemo-impl.h`). #3458 exposed the offline twin (`offlineStreamSetOption`) to JavaScript, but the node addon's `OnlineStream` was skipped — and `SherpaOnnxOnlineModelConfig` has no language field, so `sherpa-onnx-node` users of the multilingual streaming Nemotron models currently cannot select a language at all.

This PR binds the three online-stream option functions in the node addon and surfaces them on `OnlineStream`:

```js
const stream = recognizer.createStream();
stream.setOption('language', 'de');  // pin this stream to German
stream.hasOption('language');        // true
stream.getOption('language');        // 'de'
```

Because the option is re-read on each decode call, a stream can switch language mid-session without recreating the recognizer, and different streams of the same recognizer can be pinned to different languages.

Downstream context: FreeShow, an open-source church presentation app with an AI scripture feature under review (ChurchApps/FreeShow#3579), ships this exact multilingual Nemotron export via sherpa-onnx-node and needs to pin the congregation's language per stream.

## Changes

- `harmony-os/SherpaOnnxHar/sherpa_onnx/src/main/cpp/streaming-asr.cc` (shared with `scripts/node-addon-api/src/` via the existing symlink): add `OnlineStreamSetOptionWrapper` / `OnlineStreamGetOptionWrapper` / `OnlineStreamHasOptionWrapper`, mirroring `OfflineStreamSetOptionWrapper` in `non-streaming-asr.cc`, registered as `onlineStreamSetOption` / `onlineStreamGetOption` / `onlineStreamHasOption`.
- `scripts/node-addon-api/lib/streaming-asr.js`: add `OnlineStream.setOption()` / `getOption()` / `hasOption()`.
- `nodejs-addon-examples/test_asr_streaming_nemotron.js`: new example decoding the multilingual streaming Nemotron model with per-stream language pinning (German and Spanish streams of one recognizer, plus automatic detection).
- `nodejs-addon-examples/README.md`: table entry and run instructions for the new example.
- `.github/scripts/test-nodejs-addon-npm.sh`: run the new example. The tarball is ~453 MB — the same order as the parakeet/canary blocks already in this script — happy to drop this block if you prefer to keep CI lighter.

## Example output

From `node ./test_asr_streaming_nemotron.js` with `sherpa-onnx-nemotron-3.5-asr-streaming-0.6b-1120ms-int8-2026-06-11`:

```
hasOption('language'): true, getOption('language'): 'de'
de pinned: hat ein Ende, nur die Wurst
hasOption('language'): true, getOption('language'): 'es'
es pinned: preguntas que puede hacer tu país por ti.  Pregunta qué puedes hacer tú por tu país
hasOption('language'): false, getOption('language'): ''
auto: hat ein Ende, nur die Wurst
```

## Notes

- Backward compatible: only adds new exports and methods.
- `getOption` returns `''` for an option that has not been set, matching the C/C++ semantics (`OnlineStream::GetOption`); `hasOption` disambiguates. The binding copies the stream-owned pointer into a JS string immediately.
- English-only Nemotron models ignore the `language` option (`IsMultilingual` guard) and unknown keys are decoding no-ops, so the new API is safe on all models.
- The C API example `c-api-examples/streaming-nemotron-c-api.c` already demonstrates the same option in C; this brings the node addon to parity with the Flutter/WASM/.NET/Go/Swift/Rust bindings, which all expose it.
- Tested by building the macOS arm64 addon from this branch (cmake flags from `test-nodejs-addon-api.yaml`, then `cmake-js compile`) and running the new example against the released 1120ms int8 tarball (output above). Additionally verified: per-stream isolation (two interleaved streams of one recognizer pinned to different languages match their single-stream results exactly) and mid-stream language switching. `npm run typecheck` passes. Note the node-addon CI workflows do not run on pull requests.

## Summary by CodeRabbit

* **New Features**
  * Added support for setting, retrieving, and checking streaming recognition options.
  * Added multilingual Nemotron streaming speech recognition support, including optional language selection and automatic language detection.
  * Added a runnable example for German and Spanish transcription.

* **Documentation**
  * Updated usage examples with setup and execution instructions for the multilingual Nemotron model.

* **Tests**
  * Added automated coverage for Nemotron streaming speech recognition.

### DIFF
--- a/.github/scripts/test-nodejs-addon-npm.sh
+++ b/.github/scripts/test-nodejs-addon-npm.sh
@@ -486,6 +486,16 @@ rm sherpa-onnx-streaming-paraformer-bilingual-zh-en.tar.bz2
 node ./test_asr_streaming_paraformer.js
 rm -rf sherpa-onnx-streaming-paraformer-bilingual-zh-en
 
+echo "----------streaming ASR Nemotron multilingual----------"
+
+curl -SL -O https://github.com/k2-fsa/sherpa-onnx/releases/download/asr-models/sherpa-onnx-nemotron-3.5-asr-streaming-0.6b-1120ms-int8-2026-06-11.tar.bz2
+tar xvf sherpa-onnx-nemotron-3.5-asr-streaming-0.6b-1120ms-int8-2026-06-11.tar.bz2
+rm sherpa-onnx-nemotron-3.5-asr-streaming-0.6b-1120ms-int8-2026-06-11.tar.bz2
+
+node ./test_asr_streaming_nemotron.js
+
+rm -rf sherpa-onnx-nemotron-3.5-asr-streaming-0.6b-1120ms-int8-2026-06-11
+
 echo "----------non-streaming asr----------"
 
 curl -SL -O https://github.com/k2-fsa/sherpa-onnx/releases/download/asr-models/sherpa-onnx-zipformer-en-2023-04-01.tar.bz2
--- a/harmony-os/SherpaOnnxHar/sherpa_onnx/src/main/cpp/streaming-asr.cc
+++ b/harmony-os/SherpaOnnxHar/sherpa_onnx/src/main/cpp/streaming-asr.cc
@@ -439,6 +439,116 @@ static void AcceptWaveformWrapper(const Napi::CallbackInfo &info) {
 #endif
 }
 
+static void OnlineStreamSetOptionWrapper(const Napi::CallbackInfo &info) {
+  Napi::Env env = info.Env();
+
+  if (info.Length() != 3) {
+    std::ostringstream os;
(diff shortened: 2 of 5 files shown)


csukuangfj (approved):

PR #3907, opened 2026-08-31
Fix Qwen3-ASR hallucinating text on silent audio when hotwords/language are set

### LINKED ISSUE #3509: [BUG] Qwen3-ASR: hotwords and language parameters affect recognition results on empty audio
### Describe the issue

When using the same empty audio file with different combinations of `--hotwords` and `--language` parameters in Qwen3-ASR, different recognition results are returned. This suggests that the `hotwords` and `language` parameters are unexpectedly affecting the ASR output even when processing identical empty audio content.

### Steps to reproduce

1. Prepare an empty audio file (or any silent audio)
2. Run the following commands with different parameter combinations using the official script `offline-qwen3-asr-decode-files.py`

**Test case 1** (no hotwords, no language):
```bash
python offline-qwen3-asr-decode-files.py \
    --conv-frontend=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/conv_frontend.onnx \
    --encoder=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/encoder.int8.onnx \
    --decoder=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/decoder.int8.onnx \
    --tokenizer=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/tokenizer \
    --max-new-tokens=128 \
    --language="" \
    --hotwords="" \
    --num-threads=2 \
    my_audio_0.wav
```
Result: `""` (Correct)

---

**Test case 2** (with hotwords, no language):
```bash
python offline-qwen3-asr-decode-files.py \
    --conv-frontend=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/conv_frontend.onnx \
    --encoder=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/encoder.int8.onnx \
    --decoder=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/decoder.int8.onnx \
    --tokenizer=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/tokenizer \
    --max-new-tokens=128 \
    --language="" \
    --hotwords="北京,上海,广州,深圳" \
    --num-threads=2 \
    my_audio_0.wav
```
Result: `"system"` (Incorrect)

---

**Test case 3** (with language, no hotwords):
```bash
python offline-qwen3-asr-decode-files.py \
    --conv-frontend=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/conv_frontend.onnx \
    --encoder=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/encoder.int8.onnx \
    --decoder=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/decoder.int8.onnx \
    --tokenizer=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/tokenizer \
    --max-new-tokens=128 \
    --language="Chinese" \
    --hotwords="" \
    --num-threads=2 \
    my_audio_0.wav
```
Result: `"嗯，嗯。"` (Incorrect)

---

**Test case 4** (with both language and hotwords):
```bash
python offline-qwen3-asr-decode-files.py \
    --conv-frontend=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/conv_frontend.onnx \
    --encoder=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/encoder.int8.onnx \
    --decoder=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/decoder.int8.onnx \
    --tokenizer=./sherpa-onnx-qwen3-asr-0.6B-int8-2026-03-25/tokenizer \
    --max-new-tokens=128 \
    --language="Chinese" \
    --hotwords="北京,上海,广州,深圳" \
    --num-threads=2 \
    my_audio_0.wav
```
Result: `"北京 上海 广州 深圳。"` (Incorrect)

### Expected behavior

All test cases should return similar results when processing the same empty/silent audio file, regardless of the `hotwords` and `language` parameter settings. The parameters should only affect recognition quality for actual speech content, not cause different outputs for identical input. The correct result should be an empty string as shown in Test Case 1.

### Environment

- **sherpa-onnx version**: 1.12.38
- **OS**: Linux UOS-PC 4.19.0-arm64-desktop / Windows 11 (reproducible on both)
- **Python version**: 3.10.13
- **Architecture**: ARM64 (Linux) / x86_64 (Windows)

### Logs

| Test Case | hotwords | language | Result |
|-----------|----------|----------|--------|
| 1 | `""` | `""` | `""` (Correct) |
| 2 | `"北京,上海,广州,深圳"` | `""` | `"system"` |
| 3 | `""` | `"Chinese"` | `"嗯，嗯。"` |
| 4 | `"北京,上海,广州,深圳"` | `"Chinese"` | `"北京 上海 广州 深圳。"` |

### Files

- [test.log](https://github.com/user-attachments/files/26694412/test.log) - Full test output log
- [my_audio_0.zip](https://github.com/user-attachments/files/26694443/my_audio_0.zip) - Empty/silent audio test file

fra-shipper:
Fixes #3509

### Root cause

`TrimAudioFeatures()` walks `audio_features` (shape `[1, A, H]`) back-to-front
looking for the last frame whose energy exceeds `eps`. When every frame is
below `eps` (`A_valid <= 0`, i.e. the whole clip is silence), it returns the
original, untrimmed tensor -- the same thing it does when nothing needed
trimming. The caller in `GenerateText()` has no way to tell those two cases
apart: it only shrinks `audio_token_len` when the trimmed tensor is shorter
than the original, so for a nonzero-duration all-silent clip `audio_token_len`
stays at its original nonzero value and the existing `audio_token_len <= 0`
early return never fires.

As a result, decoding proceeds on all-silent audio and the hotwords/language
prompt tokens get built and fed to the LLM decoder. As reported in the issue,
an empty prompt (no hotwords, no language) still happens to produce an empty
result because the decoder naturally emits EOS, but adding `--hotwords` or
`--language` biases it away from that and it hallucinates text (e.g.
`"system"`).

### Fix

`TrimAudioFeatures()` now takes an optional `bool *all_silent` out-parameter
and sets it to `true` when every frame is below the silence threshold.
`GenerateText()` uses that flag to force `audio_token_len` to `0` and return
an empty result before any hotwords/language prompt tokens are built, so
silence is handled the same way regardless of what options are set.

`TrimAudioFeatures()` is moved out of the file-local anonymous namespace and
declared in `offline-recognizer-qwen3-asr-impl.h` so it can be exercised
directly by a unit test.

### Testing

Added `offline-recognizer-qwen3-asr-impl-test.cc` with three gtest cases
against `TrimAudioFeatures()` directly:
- an all-zero (silent) tensor sets `all_silent = true`
- a tensor with real signal in the leading frames and silence in the trailing
  frames trims correctly and leaves `all_silent = false`
- a tensor with no trailing silence is returned unmodified and leaves
  `all_silent = false`

Registered the new test file in `sherpa-onnx/csrc/CMakeLists.txt`'s
`sherpa_onnx_test_srcs` list.

Built and ran the new test against a real CMake configuration
(`-DSHERPA_ONNX_ENABLE_TESTS=ON`, all other optional components off, ONNX
Runtime v1.27.1 fetched by the build):

```
cmake --build . --target offline-recognizer-qwen3-asr-impl-test -j 12
-- [100%] Built target offline-recognizer-qwen3-asr-impl-test

./bin/offline-recognizer-qwen3-asr-impl-test
[==========] Running 3 tests from 1 test suite.
[ OK ] TrimAudioFeatures.AllSilentSetsFlag (0 ms)
[ OK ] TrimAudioFeatures.TrailingSilenceIsTrimmedAndFlagStaysFalse (0 ms)
[ OK ] TrimAudioFeatures.NoTrailingSilenceFlagStaysFalse (0 ms)
[PASSED] 3 tests.

ctest -R offline-recognizer-qwen3-asr-impl-test --output-on-failure
1/1 Test #6: offline-recognizer-qwen3-asr-impl-test ... Passed 0.01 sec
100% tests passed out of 1
```

Also confirmed the test fails to build against the pre-fix code (checked out
`offline-recognizer-qwen3-asr-impl.{cc,h}` from the parent commit): the
build fails with `use of undeclared identifier 'TrimAudioFeatures'` because
the old signature isn't visible outside the anonymous namespace, then passes
again once the fix is restored.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved offline speech recognition handling for audio containing only silence.
  * Trailing silent audio frames are now trimmed while preserving valid speech features.
  * Prevented unnecessary prompt and token processing when an entire audio clip is silent.

* **Tests**
  * Added coverage for all-silent, trailing-silence, and no-silence audio scenarios.

### DIFF
--- a/sherpa-onnx/csrc/CMakeLists.txt
+++ b/sherpa-onnx/csrc/CMakeLists.txt
@@ -830,6 +830,7 @@ if(SHERPA_ONNX_ENABLE_TESTS)
     context-graph-test.cc
     lfr-test.cc
     math-test.cc
+    offline-recognizer-qwen3-asr-impl-test.cc
     offline-whisper-timestamp-rules-test.cc
     packed-sequence-test.cc
     pad-sequence-test.cc
--- a/sherpa-onnx/csrc/offline-recognizer-qwen3-asr-impl-test.cc
+++ b/sherpa-onnx/csrc/offline-recognizer-qwen3-asr-impl-test.cc
@@ -1,0 +1,98 @@
+// sherpa-onnx/csrc/offline-recognizer-qwen3-asr-impl-test.cc
+//
+// Copyright (c)  2026  fra-shipper
+
+#include "sherpa-onnx/csrc/offline-recognizer-qwen3-asr-impl.h"
+
+#include <array>
+#include <cstring>
+
+#include "gtest/gtest.h"
+#include "sherpa-onnx/csrc/onnx-utils.h"
+
+namespace sherpa_onnx {
+
+// Regression test for https://github.com/k2-fsa/sherpa-onnx/issues/3509
+//
+// When every frame of audio_features is silence, TrimAudioFeatures() must
+// report that via |all_silent| so that GenerateText() can short-circuit to
+// an empty result before any hotwords/language prompt tokens are built.
+// Previously the all-silent case was indistinguishable from "nothing needed
+// trimming", so decoding proceeded and the hotwords/language prompt could
+// bias the LLM decoder into hallucinating text for silent audio.
+TEST(TrimAudioFeatures, AllSilentSetsFlag) {
+  Ort::AllocatorWithDefaultOptions allocator;
+
+  constexpr int32_t kFrames = 5;
+  constexpr int32_t kDim = 4;
+  std::array<int64_t, 3> shape{1, kFrames, kDim};
+  Ort::Value audio_features =
+      Ort::Value::CreateTensor<float>(allocator, shape.data(), shape.size());
+
+  float *p = audio_features.GetTensorMutableData<float>();
+  std::memset(p, 0, sizeof(float) * kFrames * kDim);
+
+  bool all_silent = false;
+  Ort::Value trimmed =
+      TrimAudioFeatures(std::move(audio_features), allocator, &all_silent);
+
+  EXPECT_TRUE(all_silent);
+
+  auto trimmed_shape = trimmed.GetTensorTypeAndShapeInfo().GetShape();
+  ASSERT_EQ(trimmed_shape.size(), 3u);
+  EXPECT_EQ(trimmed_shape[1], kFrames);
+}
+
+TEST(TrimAudioFeatures, TrailingSilenceIsTrimmedAndFlagStaysFalse) {
+  Ort::AllocatorWithDefaultOptions allocator;
+
+  constexpr int32_t kFrames = 5;
+  constexpr int32_t kValidFrames = 3;
+  constexpr int32_t kDim = 4;
+  std::array<int64_t, 3> shape{1, kFrames, kDim};
+  Ort::Value audio_features =
+      Ort::Value::CreateTensor<float>(allocator, shape.data(), shape.size());
+
+  float *p = audio_features.GetTensorMutableData<float>();
+  std::memset(p, 0, sizeof(float) * kFrames * kDim);
+  for (int32_t a = 0; a < kValidFrames; ++a) {
+    p[a * kDim] = 1.0f;
+  }
+
+  bool all_silent = false;
+  Ort::Value trimmed =
+      TrimAudioFeatures(std::move(audio_features), allocator, &all_silent);
+
+  EXPECT_FALSE(all_silent);
+
+  auto trimmed_shape = trimmed.GetTensorTypeAndShapeInfo().GetShape();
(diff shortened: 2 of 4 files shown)


csukuangfj (approved):

PR #3912, opened 2026-09-03
Fix data race on shared PRNG in FunASR-Nano and Qwen3-ASR offline recognizers

fra-shipper:
## Root cause

`OfflineRecognizerFunASRNanoImpl` and `OfflineRecognizerQwen3ASRImpl` each hold a single
`mutable std::mt19937 rng_`, seeded once in the constructor
(`offline-recognizer-funasr-nano-impl.h:66`, `offline-recognizer-qwen3-asr-impl.h:75`)
and drawn from inside `SampleTokenWithTemperatureAndTopP()` -- a `const` method reachable
from `DecodeStreams()` via `GenerateText()` -- whenever temperature sampling is enabled
(`temperature > 1e-6`; the default is greedy decoding at `1e-6`, so the bug is dormant
unless the caller opts into sampling via `--funasr-nano-temperature` /
`--qwen3-asr-temperature`, wired in through `OfflineModelConfig::Register`).

`sherpa-onnx-offline-parallel.cc`, the project's own documented way to batch-decode
concurrently with a single loaded model instead of `nj` copies, constructs exactly ONE
`sherpa_onnx::OfflineRecognizer recognizer(config)` and spawns `nj` `std::thread`s that
each call `recognizer->DecodeStreams(...)` on that same shared instance. `DecodeStreams()`
just forwards to `impl_->DecodeStreams()` with no synchronization
(`offline-recognizer.cc`, around line 172). So with temperature sampling on, every decode
thread draws from the same `rng_` with no locking: a data race on the shared `mt19937`
state array (UB), and it silently defeats the documented `seed=<N>` reproducibility knob
under concurrency -- the same audio can decode to different text run-to-run depending on
thread interleaving.

## Fix

Add a `mutable std::mutex rng_mutex_` next to `rng_` in both impl classes, and take a
`std::lock_guard<std::mutex> lock(rng_mutex_)` around each draw from `rng_`
(`SampleTokenWithTemperatureAndTopP()` in both `.cc` files, at the two draw sites in
each). Greedy decoding (the default) never reaches these draw sites and is unaffected;
existing single-threaded callers see no behavior change.

## Testing

`OfflineRecognizerFunASRNanoImpl`/`OfflineRecognizerQwen3ASRImpl` can't be constructed in
a unit test without loading real ONNX model weights (their constructors eagerly open the
model files), so the added test, `offline-recognizer-shared-rng-test.cc`, isolates the
exact `rng_`/`rng_mutex_` synchronization pattern both fixes now share and exercises it
under the same one-instance/`nj`-threads usage pattern as
`sherpa-onnx-offline-parallel.cc` (8 threads, 50 draws each, with an injected delay inside
the critical section to widen the race window), asserting via an atomic
overlap-detector that no two threads are ever inside the critical section at once.

Built and ran locally (macOS arm64, Release):

- With the mutex removed (reproducing the pre-fix unguarded `dist(rng_)` call):
  ```
  [ RUN      ] SharedRngThreadSafety.ConcurrentSamplingIsSerialized
  offline-recognizer-shared-rng-test.cc:103: Failure
  Expected equality of these values:
    sampler.MaxConcurrentEntries()
      Which is: 8
    1
  [  FAILED  ] SharedRngThreadSafety.ConcurrentSamplingIsSerialized (12 ms)
  ```
  (reproduced consistently across 3 runs)

- With the mutex in place (as committed):
  ```
  [ RUN      ] SharedRngThreadSafety.ConcurrentSamplingIsSerialized
  [       OK ] SharedRngThreadSafety.ConcurrentSamplingIsSerialized (106 ms)
  [  PASSED  ] 1 test.
  ```
  (reproduced consistently across 3 runs)

Also ran the existing `offline-recognizer-qwen3-asr-impl-test` (3 tests, all pass,
unaffected by this change) and built `sherpa-onnx-core` in full with no new warnings or
errors on the four touched files.

`clang-format --style=file` reports no diff on any line this PR added.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of random sampling during concurrent speech recognition decoding.
  - Prevented simultaneous access to shared sampling state when multiple decoding operations run in parallel.

- **Tests**
  - Added coverage verifying that concurrent random sampling is properly serialized.

### DIFF
--- a/sherpa-onnx/csrc/CMakeLists.txt
+++ b/sherpa-onnx/csrc/CMakeLists.txt
@@ -831,6 +831,7 @@ if(SHERPA_ONNX_ENABLE_TESTS)
     lfr-test.cc
     math-test.cc
     offline-recognizer-qwen3-asr-impl-test.cc
+    offline-recognizer-shared-rng-test.cc
     offline-whisper-timestamp-rules-test.cc
     packed-sequence-test.cc
     pad-sequence-test.cc
--- a/sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.cc
+++ b/sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.cc
@@ -372,7 +372,11 @@ int64_t OfflineRecognizerFunASRNanoImpl::SampleTokenWithTemperatureAndTopP(
 
   if (top_p >= 1.0f) {
     std::uniform_real_distribution<float> dist(0.0f, 1.0f);
-    float sample = dist(rng_);
+    float sample;
+    {
+      std::lock_guard<std::mutex> lock(rng_mutex_);
+      sample = dist(rng_);
+    }
     float cumsum = 0.0f;
     for (int32_t i = 0; i < vocab_size; ++i) {
       cumsum += probs[i];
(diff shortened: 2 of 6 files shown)


csukuangfj (approved):

PR #3920, opened 2026-09-05
Fix Canary empty transcript when the first decoded token is eos

### LINKED ISSUE #3919: Canary returns an empty result when the first decoded token is `<|endoftext|>`
**Version:** 1.13.7 (`917bed9`)
**Model:** `sherpa-onnx-nemo-canary-180m-flash-en-es-de-fr-int8`
**Platform:** macOS 15, Apple Silicon, CPU

### Symptom

`SherpaOnnxGetOfflineStreamResult` returns an empty `text` for ordinary speech. No error is raised, so a caller cannot distinguish it from silence. The decode is suspiciously fast — ~0.13s for 4s of audio, i.e. encoder time only.

### Cause

`offline-recognizer-canary-impl.h`, `DecodeStream`:

```cpp
int32_t max_token_id = GetMaxTokenId(&logits);   // line 72
std::vector<int32_t> tokens = {max_token_id};    // line 79

for (int32_t i = 1; i <= num_tokens; ++i) {
  if (tokens.back() == eos) break;               // line 87 — fires immediately
  ...
}
tokens.pop_back();                               // line 98 — empties the vector
```

When the argmax of the first decoder logits is `<|endoftext|>`, the loop never runs, `pop_back()` empties `tokens`, and the result is `""`. The first generated position is never a legitimate place for end-of-transcript when the audio contains speech.

### Reproduction

The trigger is not a property of the audio you can measure — it is a knife edge. From one 4.0s English recording (16kHz mono) that decodes to `""`:

| input | result |
|---|---|
| the recording as-is | `""` |
| every sample x 1.001 | `"Hey, let's see if this sucks for eight seconds of audio."` |
| every sample x 0.999 | `""` |
| speech shifted +100ms | full sentence |
| speech shifted +300ms | `""` |
| speech shifted +500ms | full sentence |

A 0.1% amplitude change flips it. Level, duration, DC offset, noise floor and language do not separate failures from successes.

The failing file: [`canary-empty-decode-4s.wav`](https://github.com/Mr-Sunglasses/vocamac/raw/sherpa-canary-empty-decode-sample/canary-empty-decode-4s.wav) (4.0s, 16kHz mono). Decoding it with `sherpa-onnx-offline` and the model above reproduces the empty result directly.

kyo-zzz:
Fixes #3919

### Root cause

In `offline-recognizer-canary-impl.h`, the greedy loop seeds `tokens` with the argmax of the first decoder logits and breaks as soon as the last token is eos. When that first argmax is already `<|endoftext|>`, the loop never runs and the trailing `tokens.pop_back()` — which normally removes the eos — removes the only generated token, so the result is an empty string with no indication that anything went wrong.

Some inputs really are that close to the edge. For the wav attached to the issue (4 s of speech), the first-position logits put `<|endoftext|>` only 0.14 above the best text token, and adding 50 ms of leading silence is enough to flip the argmax to eos on x86 too — the reporter's macOS/ARM run lands on the other side of the same knife edge. Feeding that eos back makes the decoder emit eos again, so continuing after it is not an option and the transcript is lost entirely.

### Fix

`SelectCanaryFirstToken()` now decides the first generated token:

- if the argmax is not eos, nothing changes;
- if the argmax is eos and the audio contains signal, the best non-eos token is used instead — a transcript for non-silent audio is never legitimately empty, and the near-tie means the text token is what the model intended;
- if the argmax is eos and the audio is silent, eos is kept: an empty transcript is the correct answer there.

"Contains signal" is decided by `CanaryHasSignal()` from the normalized fbank features. Those features are scale invariant, so the recording level does not matter: for this model, digital silence measures ~0.09 max |feature| while speech stays above ~4 even when it peaks at -60 dBFS; the threshold sits at 1.0.

Behavior on silent audio is unchanged, including the empty result.

### Verification

`sherpa-onnx-nemo-canary-180m-flash-en-es-de-fr-int8`, CPU, Windows:

| input | before | after |
|---|---|---|
| issue wav, as-is | full text | full text (unchanged) |
| issue wav + 50 ms / 300 ms / 800 ms leading silence | "" | full text |
| issue wav + 50 ms low-level noise lead-in | "" | full text |
| 4 s digital silence | "" | "" (unchanged) |
| official `test_wavs/en.wav`; issue wav at 0.6x-1.4x amplitude | full text | full text (unchanged) |

The new `offline-recognizer-canary-impl-test` covers the first-token selection (eos suppressed on non-silent audio, kept on silence, kept when no non-eos token exists) and the silence detector.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Canary speech recognition for non-silent audio when the decoder initially selects an end-of-speech token.
  * Prevented empty transcripts by selecting the strongest valid speech token instead.
  * Preserved end-of-speech handling for silent audio.
  * Improved handling of tied token scores and vocabularies containing only the end-of-speech token.

* **Tests**
  * Added coverage for first-token selection and silence detection.
  * Expanded offline recognizer test coverage for Canary, Cohere Transcribe, FunASR Nano, and Moonshine V2.


kyo-zzz:
For reference, the Qwen3-ASR offline recognizer already handles this case the same way: `offline-recognizer-qwen3-asr-impl.cc:994` re-samples the first token with eos excluded and only returns an empty result if eos still wins. This PR applies the equivalent greedy-search rule (best non-eos token) to Canary, gated on the audio not being silent.

csukuangfj (approved):

PR #3921, opened 2026-09-05
Fix FunASR-Nano hallucinating text on silent audio

kyo-zzz:
### Root cause

`OfflineRecognizerFunASRNanoImpl::DecodeStreams()` runs the LLM decoder on the audio embeddings regardless of the input content. For a clip of digital silence, the decoder hallucinates fillers (e.g. `"嗯。"`, `"对吧？"`) and, when hotwords are set, the hotword-biased prompt produces unrelated text — with `--funasr-nano-hotwords="人工智能,大模型,推理引擎"`, 4 s of digital silence decodes to `"I am not sure."`.

The Qwen3-ASR recognizer in this repo already short-circuits this case: `TrimAudioFeatures()` reports all-silent clips and `GenerateText()` returns an empty result before any hotwords/language prompt tokens are built (#3907). FunASR-Nano has no equivalent check.

### Fix

`FunASRNanoAudioIsSilent()` detects constant (digital-silence) frames: dither is disabled for this model family and the fbank is deterministic, so constant samples produce bit-identical frames while any real signal varies from frame to frame. `DecodeStreams()` returns an empty transcript for such clips before the prompt is built, skipping the encoder adaptor and the LLM entirely.

### Verification

`sherpa-onnx-funasr-nano-int8-2025-12-30`, CPU, Windows:

| input | hotwords | before | after |
|---|---|---|---|
| 4 s digital silence | none | `"嗯。"` | `""` |
| 4 s digital silence | 孙膑,庞涓 | `"/sil"` | `""` |
| 1 s digital silence | none | `"对吧？"` | `""` |
| 1 s digital silence | 人工智能 | `"嗯"` | `""` |
| real speech (test_wavs, with and without hotwords) | — | full transcript | full transcript (unchanged) |

The new `offline-recognizer-funasr-nano-impl-test` covers the silence detector (constant frames, varying frames, single frame, empty input).

## Summary by CodeRabbit

* **Bug Fixes**
  * FunASR Nano now returns an empty transcription for silent audio instead of performing unnecessary recognition processing.

* **Tests**
  * Added coverage for silent, constant, zero-filled, varying, repeated non-uniform, and empty audio feature inputs to improve recognition reliability.


sherpa-onnx/csrc/offline-recognizer-funasr-nano-impl.cc
```diff
@@ -868,6 +883,16 @@ void OfflineRecognizerFunASRNanoImpl::DecodeStreams(OfflineStream **ss,
       continue;
     }
 
+    if (FunASRNanoAudioIsSilent(f.data(), static_cast<int32_t>(f.size()))) {
+      // The whole clip is silence. Return an empty result now, before any
+      // hotwords/language prompt tokens are built, so they cannot bias the
+      // LLM decoder into hallucinating text for silent audio.
+      OfflineRecognitionResult r;
+      r.text = "";
```
csukuangfj:
```suggestion
```

csukuangfj (approved):

PR #3923, opened 2026-09-05
Fix Moonshine v2 hallucinating text on silent audio

kyo-zzz:
### Root cause

`OfflineRecognizerMoonshineV2Impl::DecodeStreams()` runs the encoder and the decoder regardless of the input content. A clip of digital silence decodes to `"You"` (deterministically), so callers cannot distinguish silence from real speech.

The Qwen3-ASR recognizer in this repo already short-circuits this case: an all-silent clip returns an empty result before any decoding happens (#3907). Moonshine v2 has no equivalent check.

### Fix

`MoonshineV2AudioIsSilent()` checks the raw samples returned by `GetFrames()` — for moonshine these are the unmodified PCM samples, no feature extraction involved. If every sample is zero within `1e-6` (digital silence), `DecodeStreams()` returns an empty transcript before the encoder runs.

### Verification

`sherpa-onnx-moonshine-tiny-en-quantized-2026-02-27`, CPU, Windows:

| input | before | after |
|---|---|---|
| 4 s digital silence | `"You"` | `""` |
| 1 s digital silence | `"You"` | `""` |
| 4 s noise at -60 dBFS | `"You"` | `"You"` (unchanged — noise-floor audio stays decoded, same scope as the Qwen3-ASR fix) |
| `test_wavs/0.wav`; 4 s speech with/without leading silence | full transcript | full transcript (unchanged) |

The new `offline-recognizer-moonshine-v2-impl-test` covers the silence detector (zeros, a single non-zero sample, float noise around zero, single sample, empty input).

## Summary by CodeRabbit

- **Bug Fixes**
  - Silent audio clips now return an empty transcription instead of potentially producing hallucinated text.
  - Near-zero audio is correctly recognized as silence, while clips containing signal continue through transcription.
  - Empty audio inputs remain handled without producing transcription text.

- **Tests**
  - Added coverage for silent, near-zero, non-silent, single-sample, and empty audio inputs.


kyo-zzz:
On the NaN suggestion: keeping the current behavior deliberately. A non-finite sample is invalid data, not a valid signal — treating NaN as non-silent would make DecodeStream run the encoder and decoder on NaN input and produce unpredictable output, which is exactly the kind of garbage this PR is meant to avoid. With the current check, an all-NaN clip is classified as silence and returns an empty transcript (a safe degradation), and a clip that also contains real samples is decoded as before. Happy to revisit if there is a specific upstream case that produces NaNs.

sherpa-onnx/csrc/offline-recognizer-moonshine-v2-impl.h
```diff
@@ -87,6 +108,17 @@ class OfflineRecognizerMoonshineV2Impl : public OfflineRecognizerImpl {
 
     std::vector<float> audio = s->GetFrames();
 
+    if (MoonshineV2AudioIsSilent(audio.data(),
+                                 static_cast<int32_t>(audio.size()))) {
+      // The whole clip is silence. Return an empty result before running the
+      // encoder, so the decoder cannot hallucinate text (e.g. "You") for
+      // silent audio.
+      OfflineRecognitionResult r;
+      r.text = "";
```
csukuangfj:
```suggestion
```

csukuangfj (approved):

PR #3924, opened 2026-09-06
Fix Cohere Transcribe hallucinating sentences on silent audio

kyo-zzz:
### Root cause

`OfflineRecognizerCohereTranscribeImpl::DecodeStreams()` runs the encoder and the decoder regardless of the input content. For a clip of digital silence, most supported languages decode to a hallucinated sentence (deterministically, measured with the 14-language int8 model):

| language | output for 4 s of digital silence |
|---|---|
| en | `""` (the only correct one) |
| de | `"Herr Präsident, meine Damen und Herren!"` |
| fr | `"Sous-titrage Société Radio-Canada"` |
| zh | `"谢谢"` |
| ko | `"으음."` |
| vi | `"ạ"` |
| ar | `"نهايه الدرس"` |
| es | `"El juego se queda en la pista."` |

The Qwen3-ASR recognizer in this repo already short-circuits this case: an all-silent clip returns an empty result before any decoding happens (#3907). Cohere Transcribe has no equivalent check.

### Fix

`CohereHasSignal()` decides whether the clip contains any signal, and `DecodeStreams()` returns an empty transcript when it does not, before the encoder and the decoder run.

The features returned by `GetFrames()` are per-feature normalized fbank values, which are scale invariant — the recording level does not matter. Measured in this model's own feature domain: digital silence gives a max |feature| of about 0.09, while the official test wavs for all 8 supported languages stay between 3.3 and 4.5 regardless of their level; the threshold sits at 1.0.

### Verification

`sherpa-onnx-cohere-transcribe-14-lang-int8-2026-04-01`, CPU, Windows:

| input | before | after |
|---|---|---|
| 4 s digital silence, de/fr/zh/ko/vi/ar/es | hallucinated sentences (table above) | `""` |
| 4 s digital silence, en | `""` | `""` (unchanged) |
| 8 official test_wavs (one per language) | full transcript | full transcript (unchanged) |
| 4 s speech with 50/200/500 ms leading silence, en/de/zh | full transcript | full transcript (unchanged) |

The new `offline-recognizer-cohere-transcribe-impl-test` covers the silence detector (thresholds on both sides, a single loud sample, empty input).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved offline transcription handling for digitally silent audio.
  * Silent clips now return an empty transcript instead of potentially generating hallucinated text.
  * Improved handling of empty and very quiet audio to avoid unintended transcription results.

* **Tests**
  * Added coverage for silence, empty input, quiet audio, and audio containing a detectable signal to help ensure consistent transcription behavior.

### DIFF
--- a/sherpa-onnx/csrc/offline-recognizer-cohere-transcribe-impl.h
+++ b/sherpa-onnx/csrc/offline-recognizer-cohere-transcribe-impl.h
@@ -173,7 +173,6 @@ class OfflineRecognizerCohereTranscribeImpl : public OfflineRecognizerImpl {
       // encoder/decoder run, so they cannot hallucinate text for silent
       // audio.
       OfflineRecognitionResult r;
-      r.text = "";
       s->SetResult(r);
       return;
     }
--- a/sherpa-onnx/csrc/offline-recognizer-cohere-transcribe-impl.h
+++ b/sherpa-onnx/csrc/offline-recognizer-cohere-transcribe-impl.h
@@ -24,6 +24,31 @@
 
 namespace sherpa_onnx {
 
+// The decoder in this file happily transcribes silence into hallucinated
+// sentences for most languages (measured with
+// sherpa-onnx-cohere-transcribe-14-lang-int8-2026-04-01: e.g. "Herr
+// Präsident, meine Damen und Herren!" for de, "Sous-titrage Société
+// Radio-Canada" for fr, "谢谢" for zh; only en returns an empty string), so
+// digitally silent clips must be detected and short-circuited before the
+// decoder runs.
+//
+// Silence is detected on the normalized fbank features returned by
+// OfflineStream::GetFrames(). Those features are scale invariant, so the
+// recording level does not matter. The fbank configuration is the same as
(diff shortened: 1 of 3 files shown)


sherpa-onnx/csrc/offline-recognizer-cohere-transcribe-impl.h
```diff
@@ -143,6 +168,16 @@ class OfflineRecognizerCohereTranscribeImpl : public OfflineRecognizerImpl {
     std::vector<float> f = s->GetFrames();
     int32_t num_frames = f.size() / feat_dim;
 
+    if (!CohereHasSignal(f.data(), static_cast<int32_t>(f.size()))) {
+      // The whole clip is silence. Return an empty result before the
+      // encoder/decoder run, so they cannot hallucinate text for silent
+      // audio.
+      OfflineRecognitionResult r;
+      r.text = "";
```
csukuangfj:
```suggestion
```

This line can be deleted.

csukuangfj (approved):

PR #3931, opened 2026-09-07
Remove eSpeak NG and piper-phonemize for redistribution-safe builds

vicpon:
## Motivation

eSpeak NG and its `piper-phonemize` wrapper cannot ship in closed-source redistributions. This PR removes both components entirely while keeping TTS fully functional for **lexicon-based** models — Kokoro v1.0+ (`--kokoro-lexicon`) and Matcha (dict + fst). Based on tag `v1.13.7`.

## What changes

**Core behavior** (`sherpa-onnx/csrc/`):
- `piper-phonemize-lexicon.cc/.h` — all espeak integration removed (~643 lines). Token loading retained. Models that *require* phonemization (Piper/Coqui/Icefall/Inflect VITS, Kokoro v0.19, Kitten) now **fail loudly at construction** with an actionable message instead of silently producing empty audio.
- `kokoro-multi-lang-lexicon.cc` — lexicon-only per-word lookup; out-of-lexicon words are **dropped** (surrounding speech unaffected) with a once-per-process warning via `std::call_once`.
- `matcha-tts-lexicon.cc` — espeak fallback pipeline removed; the CJK path never used eSpeak and is untouched.
- `offline-tts-kokoro-impl.h` — multi-lingual Kokoro (`meta_data.version >= 2`) now **requires `--kokoro-lexicon`**; a `--kokoro-lang`-only config fails at construction instead of returning "Failed to convert" with no audio and no explanation.
- `offline-tts-kokoro-model-config.cc` — `--kokoro-data-dir` is now **optional** (validated only when provided), matching the Matcha/VITS pattern.

**Build surface**: cmake fetch recipes for both packages deleted; `piper_phonemize` dropped from core link and tests; all 19 `build-*.sh` scripts cleaned of espeak/piper flags; iOS fat-lib member lists, `sherpa-onnx-static.pc.in`, Rust `build.rs`, Pascal API, and the three MFC `.props` all updated. Post-patch grep for `piper_phonemize|espeak-ng|ucd` across all link manifests: **0 hits**.

## Model support matrix (vs upstream v1.13.7)

| Model | Upstream | This PR |
|---|---|---|
| Kokoro v1.0+ multi-lingual + lexicon | ✅ | ✅ (lexicon required; OOV words dropped + warned) |
| Kokoro v1.0+ lang-only (`--kokoro-lang`) | ✅ (via eSpeak) | ❌ loud failure at construction (Message C) |
| Matcha zh-en (dict + fst) | ✅ | ✅ |
| Piper / Coqui / Icefall / Inflect VITS | ✅ | ❌ loud failure at construction (Message A) |
| Kokoro v0.19, Kitten | ✅ | ❌ loud failure at construction (Message A) |

## User-facing messages (new strings, for review)

**Message A — fatal at model load** (models needing phonemization; process exits nonzero, 255 on Unix), from `piper-phonemize-lexicon.cc`:

> This model needs the eSpeak text-to-phoneme engine to convert words into sounds, but this build of sherpa-onnx does not include it (it was removed for licensing reasons). Affected models: Piper / Coqui / Icefall / Inflect VITS voices (for example en_US-amy-low), Kokoro v0.19, and Kitten TTS models. Use a model with a lexicon instead: Kokoro v1.0 or newer with --kokoro-lexicon set to the matching lexicon file (for example lexicon-us-en.txt), or Matcha zh-en.

**Message B — non-fatal, once per process** (first out-of-lexicon word; from `offline-tts-frontend.h`):

> This build of sherpa-onnx has no built-in text-to-phoneme engine (eSpeak removed); dropping out-of-lexicon words instead of phonemizing them. Provide a lexicon covering the spoken text, or use an upstream sherpa-onnx build.

**Message C — fatal at model load** (Kokoro multi-lingual configured with `--kokoro-lang` but no `--kokoro-lexicon`; exits nonzero), from `offline-tts-kokoro-impl.h` (both `InitFrontend()` overloads):

> This Kokoro multi-lingual model needs --kokoro-lexicon in this build of sherpa-onnx; the language-based phonemization that --kokoro-lang relied on (eSpeak) has been removed. Example: --kokoro-lexicon lexicon-us-en.txt

## Verification

- Host x86_64 build with TTS=ON (no espeak installed): clean; `nm`/`strings` scans of core + c-api + cxx-api libs for `espeak_ng | speak-lib | CallPhonemizeEspeak | piper_phonemize`: **0 matches**.
- Positive: Kokoro v1.0 + `lexicon-us-en.txt` with invented OOV words → valid non-empty WAV (~5.4 s, RTF 0.92), Message B logged exactly once and never again.
- Positive: same run **without any `--kokoro-data-dir`** → valid WAV.
- Negative: piper `en_US-amy-low` → Message A, exit 255, no WAV.
- Negative: Kokoro lang-only → Message C, exit 255, no WAV.
- Android arm64-v8a (NDK r27, ONNX Runtime 1.27.1 shared): built, stripped, scans clean, deployed to a physical device (SM-S901U, Android 16) — Kokoro v1.0 loads with `data_dir=""`, synthesizes and audibly plays speech; device logcat shows Message B exactly once, named OOV drops (`onnx`, `espeak`), and zero espeak strings.
- Android x86_64 emulator (API 36): same behavior confirmed.

## UAT (user-acceptance steps)

1. Build any Android target: `./build-android-arm64-v8a.sh` (no espeak flags needed).
2. Run any Kokoro v1.0 model with `--kokoro-lexicon lexicon-us-en.txt` (no data dir) → speech plays; words missing from the lexicon are skipped, warned once.
3. Run a Piper voice → immediate exit with Message A above; no empty/garbage WAV.
4. Run Kokoro v1.0 with only `--kokoro-lang en-us` → immediate exit with Message C above.

## Notes for maintainers

Understand if this is too aggressive for mainline — it removes functionality upstream users rely on. An alternative shape would be gating eSpeak behind a CMake option (`SHERPA_ONNX_ENABLE_ESPEAK`, default ON) so distribution builders can opt out; happy to rework in that direction if preferred. The loud-failure messages and optional data-dir validation are useful independently of the removal.

## Summary by CodeRabbit

- **Breaking Changes**
  - Removed eSpeak-based phonemization from supported TTS frontends.
  - Piper phonemizer usage now reports an actionable runtime error instead of generating speech.
  - Kokoro and Matcha frontends drop words missing from their lexicons and issue a one-time warning.
  - Multilingual Kokoro models now require a lexicon; language-only configuration is no longer accepted.
  - The Kokoro data directory is optional when not used for phonemization.

- **Build and Packaging**
  - Removed related phonemizer libraries from static linking and platform build archives.


vicpon:
Closing — raised in error, this change is intended for our internal distribution rather than upstream. Thanks!

PR #3950, opened 2026-09-11
Add GetEmbedding to SpeakerEmbeddingManager

XDcobra:
## Summary

Right now `SpeakerEmbeddingManager` is mostly write + match: you can `Add` a speaker, search, verify, list names, and so on, but once an embedding is enrolled there is no way to read that vector back out. While implementing my app, where I want to persist enrollments across sessions, I usually end up keeping a second copy of the same floats next to the manager, which is easy to get out of sync and wastes memory.

This PR adds a small readout path:

- C++: `SpeakerEmbeddingManager::GetEmbedding(name, out)` fills a caller buffer of length `Dim()`, or returns false if the name is missing
- C-API: `SherpaOnnxSpeakerEmbeddingManagerGetEmbedding` + `DestroyEmbedding`, same ownership pattern as `SpeakerEmbeddingExtractorComputeEmbedding`
- C-API: `SherpaOnnxSpeakerEmbeddingManagerDim` so callers know how long the vector is (already available in C++, just missing from the C bindings)
- cxx-API: `GetEmbedding(name)` returning `std::vector<float>` (empty if missing)

What you get back is the embedding as the manager already stores it after `Add` / list average + L2 normalize. That is enough to rebuild a portable enrollment export with `GetAllSpeakers` + `GetEmbedding` per name, without inventing a file format and without mirroring embeddings in the app layer.

## How to test

```bash
cmake -DSHERPA_ONNX_ENABLE_TESTS=ON ...
make speaker-embedding-manager-test
./bin/speaker-embedding-manager-test --gtest_filter='SpeakerEmbeddingManager.GetEmbedding*'
```

## Summary by CodeRabbit

- **New Features**
  - Added access to speaker embedding dimensions through the C and C++ APIs.
  - Added retrieval of named speaker embeddings through the C and C++ APIs.
  - Missing speakers now return an empty result.
  - Added support for safely releasing retrieved embedding data.

- **Tests**
  - Added coverage for embedding retrieval, missing speakers, normalization, similarity calculations, averaging, and lookups after speaker removal.


sherpa-onnx/csrc/speaker-embedding-manager.h
```diff
@@ -110,6 +110,13 @@ class SpeakerEmbeddingManager {
   // Return a list of speaker names
   std::vector<std::string> GetAllSpeakers() const;
 
+  /* Copy the embedding of a speaker into `out`.
+   *
+   * Return true if the speaker exists; return false otherwise.
+   * `out` must point to an array of Dim() elements.
+   */
+  bool GetEmbedding(const std::string &name, float *out) const;
```
csukuangfj:
```suggestion
  std::vector<float> GetEmbedding(const std::string &name) const;
```

where an empty vector means the speaker does not exist.

csukuangfj (changes requested):

XDcobra:
Updated `GetEmbedding` to return `std::vector<float>` (empty if the speaker is missing), and added a test that still reads embeddings correctly after removing a speaker in the middle of the gallery.

sherpa-onnx/csrc/speaker-embedding-manager.h
XDcobra: Updated `GetEmbedding`to represent your requested change

csukuangfj (approved):

PR #3951, opened 2026-09-11
Support HLG decoding for offline CTC models in the C, CXX, JNI/Kotlin/Java and WASM APIs

baszczewski:
`sherpa_onnx::OfflineCtcFstDecoderConfig` has always existed in C++ and is
exposed by the Python bindings, but it is not reachable through the C API, so
HLG decoding for non-streaming CTC models is Python-only today. The streaming
counterpart, `SherpaOnnxOnlineCtcFstDecoderConfig`, has been in the C API, the
JNI and the WASM layer all along.

With a graph attached the decoder can only return paths the graph contains,
which is what a closed-grammar recognizer needs on a phone or in a browser.

### Commits

1. **C and CXX API** - adds `SherpaOnnxOfflineCtcFstDecoderConfig` and a
   `ctc_fst_decoder_config` field on `SherpaOnnxOfflineRecognizerConfig`, the
   same pair in `cxx-api`, plus a C and a C++ example and a CI step for each.
   This follows the shape of #2156, which added the last field to this struct.
2. **JNI, Kotlin and Java API** - reads the new config, and returns the graph's
   output labels as `words`.
3. **WebAssembly** - serialises the new field in `wasm/asr/sherpa-onnx-asr.js`
   and extends the struct-layout `static_assert`.

### Reading the graph's output labels

Only the JNI needed a result-side change. The C API, the CXX API and the WASM
layer all return the result as json, and `OfflineRecognitionResult::words` is
already serialised into it, so a caller reads the labels with no new API. The
JNI builds the Kotlin/Java object field by field, so there `words` had to be
added explicitly or it would be dropped.

`OfflineRecognizerResult` is declared twice, in `kotlin-api` and in `java-api`,
and the JNI looks its constructor up by descriptor. Both declarations are
updated, so the Java API keeps working too. In Kotlin the field defaults to an
empty array, which keeps existing callers compiling.

### ABI

`ctc_fst_decoder_config` goes at the end of
`SherpaOnnxOfflineRecognizerConfig`, so the offset of every existing field is
unchanged. A caller that zero-initialises the struct gets an empty graph, which
means no graph, so nothing changes for existing code.

Bindings that mirror the struct by hand rather than reading the C header
(Dart/Flutter, C#, Pascal, Rust, node-addon) are unchanged here and keep
working; they cannot set the new field yet. This follows how the last
field added to this struct was handled: #2156 for the C and CXX API, then
#2167 for Dart. I am happy to send the same follow-ups, either as one sweep
across the remaining bindings or one PR per binding - whichever you prefer.

The Go binding needs nothing: cgo generates the struct from `c-api.h`.

### Testing

- Both examples decode with the `HLG.fst` that ships with
  `sherpa-onnx-zipformer-ctc-en-2023-10-02`, the model
  `.github/scripts/test-offline-ctc.sh` already downloads for graph decoding,
  so no new asset is involved. A step in `c-api.yaml` and one in `cxx-api.yaml`
  build and run them, following the neighbouring steps.
- Both examples are also built by the ordinary CMake build
  (`SHERPA_ONNX_BUILD_C_API_EXAMPLES` and `SHERPA_ONNX_ENABLE_BINARY` are both
  on by default for a top-level build), so `linux.yaml` and `macos.yaml`
  compile them on this PR.
- Both were run locally against that model. The config reaches the core
  (`ctc_fst_decoder_config=OfflineCtcFstDecoderConfig(graph="./sherpa-onnx-zipformer-ctc-en-2023-10-02/HLG.fst", max_active=3000)`
  in the debug dump) and the result json carries the graph's labels:

  ```
  text: AFTER EARLY NIGHTFALL THE YELLOW LAMPS WOULD LIGHT UP HERE AND THERE THE SQUALID QUARTER OF THE BROTHELS
  "words": [2191, 52770, 121894, 175861, 198295, 98505, 197114, 101964, 186768, 80190, 5411, 176144, 175861, 166824, 142471, 124782, 175861, 22700]
  ```

  Eighteen labels for eighteen words, with 175861 in each of the three places
  the sentence says "THE".
- `./scripts/check_style_cpplint.sh`: the series adds no new cpplint findings;
  every file it touches reports exactly what it reports on master.
- The JNI translation unit compiles with the Android NDK clang (aarch64), and
  the whole `java-api` source set compiles with `javac -Xlint:all`, so the
  descriptor the JNI looks up matches both declarations of
  `OfflineRecognizerResult`.
- `wasm/nodejs/sherpa-onnx-wasm-nodejs.cc` compiles under `emcc`. That is the
  target where its `static_assert` means something: it fails the build if the
  JavaScript layout and the C header disagree.
- The three commits have been running in a closed-grammar voice-command
  project on desktop (Python), Android (Kotlin) and in the browser (WASM),
  against NeMo FastConformer CTC models with HLG graphs.

## Summary by CodeRabbit

* **New Features**
  * Added offline CTC decoding with HLG/FST graphs across the C, C++, Java, Kotlin, and WebAssembly APIs.
  * Added configuration for decoding graphs and maximum active paths.
  * Recognition results now expose decoded word information where supported.
  * Added C and C++ examples demonstrating Zipformer CTC recognition with HLG graphs.

* **Tests**
  * Added automated coverage for Zipformer CTC HLG recognition in the C and C++ API workflows.

### DIFF
--- a/sherpa-onnx/java-api/src/main/java/com/k2fsa/sherpa/onnx/OfflineRecognizerResult.java
+++ b/sherpa-onnx/java-api/src/main/java/com/k2fsa/sherpa/onnx/OfflineRecognizerResult.java
@@ -10,8 +10,9 @@ public class OfflineRecognizerResult {
     private final String emotion;
     private final String event;
     private final float[] durations;
+    private final int[] words;
 
-    public OfflineRecognizerResult(String text, String[] tokens, float[] timestamps, String lang, String emotion, String event, float[] durations) {
+    public OfflineRecognizerResult(String text, String[] tokens, float[] timestamps, String lang, String emotion, String event, float[] durations, int[] words) {
         this.text = text;
         this.tokens = tokens;
         this.timestamps = timestamps;
@@ -19,6 +20,7 @@ public class OfflineRecognizerResult {
         this.emotion = emotion;
         this.event = event;
         this.durations = durations;
+        this.words = words;
     }
 
     public String getText() {
@@ -47,5 +49,9 @@ public class OfflineRecognizerResult {
 
     public float[] getDurations() {
         return durations;
+    }
+
+    public int[] getWords() {
+        return words;
     }
 }
--- a/wasm/asr/sherpa-onnx-asr.js
+++ b/wasm/asr/sherpa-onnx-asr.js
@@ -442,6 +442,19 @@ function initSherpaOnnxOnlineCtcFstDecoderConfig(config, Module) {
   return {ptr: ptr, len: len, buffer: buffer};
 }
 
+function initSherpaOnnxOfflineCtcFstDecoderConfig(config, Module) {
+  const len = 2 * 4;
+  const ptr = Module._malloc(len);
+
+  const graphLen = Module.lengthBytesUTF8(config.graph || '') + 1;
+  const buffer = Module._malloc(graphLen);
+  Module.stringToUTF8(config.graph || '', buffer, graphLen);
+
+  Module.setValue(ptr, buffer, 'i8*');
+  Module.setValue(ptr + 4, config.maxActive || 3000, 'i32');
+  return {ptr: ptr, len: len, buffer: buffer};
+}
+
 function initSherpaOnnxOnlineRecognizerConfig(config, Module) {
   if (!('featConfig' in config)) {
     config.featConfig = {
@@ -1697,12 +1710,21 @@ function initSherpaOnnxOfflineRecognizerConfig(config, Module) {
     };
   }
 
+  if (!('ctcFstDecoderConfig' in config)) {
+    config.ctcFstDecoderConfig = {
+      graph: '',
+      maxActive: 3000,
+    };
+  }
+
   const feat = initSherpaOnnxFeatureConfig(config.featConfig, Module);
   const model = initSherpaOnnxOfflineModelConfig(config.modelConfig, Module);
   const lm = initSherpaOnnxOfflineLMConfig(config.lmConfig, Module);
   const hr = initSherpaOnnxHomophoneReplacerConfig(config.hr, Module);
+  const ctcFstDecoder = initSherpaOnnxOfflineCtcFstDecoderConfig(
+      config.ctcFstDecoderConfig, Module);
 
-  const len = feat.len + model.len + lm.len + 7 * 4 + hr.len;
+  const len = feat.len + model.len + lm.len + 7 * 4 + hr.len + ctcFstDecoder.len;
   const ptr = Module._malloc(len);
 
   let offset = 0;
@@ -1768,6 +1790,9 @@ function initSherpaOnnxOfflineRecognizerConfig(config, Module) {
   Module._CopyHeap(hr.ptr, hr.len, ptr + offset);
   offset += hr.len;
 
+  Module._CopyHeap(ctcFstDecoder.ptr, ctcFstDecoder.len, ptr + offset);
+  offset += ctcFstDecoder.len;
+
   return {
     buffer: buffer,
     ptr: ptr,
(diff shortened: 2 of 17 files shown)


baszczewski:
Pushed an update after the automated review:

- **The Java API was broken by the JNI change.** `java-api` declares its own
  `OfflineRecognizerConfig`, so reading `ctcFstDecoderConfig` off it would have
  raised `NoSuchFieldError` on every Java recognizer. Added
  `OfflineCtcFstDecoderConfig.java` next to the existing
  `OnlineCtcFstDecoderConfig`, and the matching field and builder setter.
  Kotlin was already fine. `javac -Xlint:all` over the whole `java-api` source
  set is clean.
- **The CI steps now assert the graph was used.** Decoding succeeds with or
  without a graph, so a clean exit proved nothing. Both steps grep the result
  json for word ids and fail if they are missing.

No change to the C API or the WASM commit.

sherpa-onnx/java-api/src/main/java/com/k2fsa/sherpa/onnx/OfflineRecognizerResult.java
```diff
@@ -10,15 +10,17 @@ public class OfflineRecognizerResult {
     private final String emotion;
     private final String event;
     private final float[] durations;
+    private final int[] words;
 
-    public OfflineRecognizerResult(String text, String[] tokens, float[] timestamps, String lang, String emotion, String event, float[] durations) {
+    public OfflineRecognizerResult(String text, String[] tokens, float[] timestamps, String lang, String emotion, String event, float[] durations, int[] words) {
```
csukuangfj, line 17: Can you document why to add this field?

wasm/asr/sherpa-onnx-asr.js
```diff
@@ -1776,6 +1801,7 @@ function initSherpaOnnxOfflineRecognizerConfig(config, Module) {
     model: model,
     lm: lm,
     hr: hr,
+    ctcFstDecoder: ctcFstDecoder,
```
csukuangfj, line 1804: Please also free it in
https://github.com/k2-fsa/sherpa-onnx/blob/f007b559b75c45c04f246245ef7bbc5cd193307d/wasm/asr/sherpa-onnx-asr.js#L1

csukuangfj (changes requested):

sherpa-onnx/java-api/src/main/java/com/k2fsa/sherpa/onnx/OfflineRecognizerResult.java
baszczewski: Done, the field now has a comment in both the Java and the Kotlin result class:

```
// The decoded word IDs. Empty for greedy search decoding, non-empty when
// an HLG graph is used, i.e. when ctcFstDecoderConfig.graph is set.
```

It condenses the note already on `OfflineCtcDecoderResult` and adds the name of
the config field that turns the ids on, since someone reading a binding has no
decoder header in front of them.

As for why the field is in this commit at all: `OfflineRecognitionResult::words`
already exists and is already returned by `AsJsonString` and by the Python
bindings, so the C API, the CXX API and the WASM layer needed nothing on the
result side. The JNI is the one layer that builds the result object field by
field, so `words` had to be named there or it would be dropped for Kotlin and
Java callers.

What I use them for: with a graph attached they are the graph's own output
labels, so they can be looked up in its `words.txt` directly, instead of
parsing the text back into words.

wasm/asr/sherpa-onnx-asr.js
baszczewski: I believe it already is, though it is easy to miss, because the free site is
keyed by name and says nothing about offline or online: `freeConfig` recurses
into `config.ctcFstDecoder` (line 102), which the streaming config already
relies on, and the offline config returns its sub-config under that same key.
So the graph string and the struct both go through it.

Verified with a stub `Module` that counts `_malloc`/`_free`: nothing is left
allocated after `freeConfig`, and removing that branch leaks exactly two
allocations.

If you would rather have the offline sub-config under its own key with its own
branch, say so and I will add it.

csukuangfj (approved):

PR #3953, opened 2026-09-12
feat(offline-transducer): expose n-best hypotheses from modified beam search

### LINKED ISSUE #3858: [FEATURE] Expose N-best hypotheses from offline transducer modified beam search
### Describe the feature

I would like `sherpa-onnx` to expose N-best hypotheses for offline transducer models when using `modified_beam_search`.

Currently, `modified_beam_search` keeps multiple hypotheses internally according to `max_active_paths`, but only the single best hypothesis is returned to the user.

A possible API would be to add an option such as:

```python
num_return_paths=5
```

separately from:

```python
max_active_paths=8
```

where:

- `max_active_paths` controls the search width
- `num_return_paths` controls how many final hypotheses are returned

The default could remain `num_return_paths=1` to preserve the current behavior.

For backward compatibility, existing fields such as:

```python
result.text
result.tokens
result.timestamps
```

could continue to represent the 1-best result, while N-best results could be exposed separately, for example:

```python
result.hypotheses
```

with each hypothesis containing fields such as:

```python
hyp.text
hyp.tokens
hyp.timestamps
hyp.ys_log_probs
hyp.score
```

The exact public API is open to discussion. In particular, I would appreciate guidance on the preferred representation across the different language bindings.

For the initial implementation, I suggest limiting the scope to:

- offline recognition
- transducer models
- `modified_beam_search`

This proposal is not intended to add a new lattice-based decoding algorithm such as icefall's `fast_beam_search_nbest`.

### Motivation / Use Case

The main use case is downstream ASR post-processing where having several plausible hypotheses is significantly more useful than having only the 1-best result.

Examples include:

- ASR error correction
- named entity correction
- external LM / LLM reranking
- domain-specific post-processing

For example, the correct named entity may already exist as the 2nd or 3rd hypothesis even when the 1-best hypothesis is incorrect. If only the 1-best result is exposed, downstream correction systems cannot make use of that information.

This would be particularly useful for deployment scenarios where `sherpa-onnx` is used as the inference runtime, while additional lightweight correction or reranking is performed outside the ASR decoder.

### Additional context

Looking at the current implementation, this seems potentially smaller than implementing a new N-best decoder.

`OfflineTransducerModifiedBeamSearchDecoder::Decode()` currently selects the final result with:

```cpp
Hypothesis hyp = cur[i].GetMostProbable(true);
```

However, `Hypotheses` already implements:

```cpp
std::vector<Hypothesis> GetTopK(int32_t k, bool length_norm) const;
```

so the decoder already has a mechanism for retrieving the top-K final hypotheses.

Also, `Hypotheses::Add()` merges hypotheses with identical token sequences using log-sum-exp, so these would naturally be N-best distinct token sequences rather than duplicate search paths.

Related issue:

- #465 — Availability of different beam search as icefall

That issue also mentions `fast_beam_search_nbest`, but this request is narrower: expose the final hypotheses already retained by `modified_beam_search`, without introducing a new decoding method or lattice dependency.

It may also be useful to expose the ranking score for each hypothesis. Since `Hypothesis` already contains `log_prob`, `lm_log_prob`, and `TotalLogProb()`, I would be interested in the maintainers' preference for the public score semantics, especially when LM or context-biasing scores are involved.

If this direction makes sense, I would be happy to work on an implementation.

shahrryyar:
Closes #3858.

`modified_beam_search` already keeps `max_active_paths` hypotheses internally but throws all but the best one away. This adds a `num_return_paths` option so callers can read the whole n-best list.

Scope is what the issue proposed: offline recognition, transducer models, `modified_beam_search`. No new decoding algorithm.

### API

```python
recognizer = sherpa_onnx.OfflineRecognizer.from_transducer(
    ...,
    decoding_method="modified_beam_search",
    max_active_paths=8,   # search width
    num_return_paths=5,   # how many to return
)

result = stream.result
result.text          # 1-best, unchanged
for hyp in result.hypotheses:   # ordered best first
    hyp.text, hyp.tokens, hyp.timestamps, hyp.ys_log_probs, hyp.score
```

### Behavior

- `num_return_paths` defaults to `1`, which keeps the current behavior exactly and leaves `hypotheses` empty. Nothing changes for existing users.
- When `> 1`, `hypotheses[0]` is the 1-best and matches the top-level `text` / `tokens` / `timestamps`.
- It is clamped to `max_active_paths`, and `GetTopK()` clamps again to how many hypotheses the beam actually holds, so a small beam simply returns fewer.
- `Validate()` rejects `num_return_paths < 1`, and `num_return_paths > 1` with any decoding method other than `modified_beam_search`.
- Each hypothesis gets the same text post-processing as the 1-best (inverse text normalization, homophone replacer, leading-space removal), so `hypotheses[0].text == result.text` holds.
- `score` is the total log-prob (acoustic + LM when an LM is used), not length-normalized. Ranking uses the same length-normalized comparison as `GetMostProbable(true)`, so the 1-best is unchanged.

### Implementation

`Hypotheses::GetTopK()` already existed, so the decoder change is small: when `num_return_paths_ > 1`, take the top-k instead of just the most probable. The token→text conversion in `Convert()` was factored into a helper (`ConvertTokens` / `ConvertTimestamps`) and reused for each hypothesis rather than duplicated.

### Verified

Built `sherpa-onnx-core` and `_sherpa_onnx` cleanly (gcc 13.3, Release), then ran `sherpa-onnx-zipformer-small-en-2023-06-26` over `test_wavs/0.wav` and `1.wav`:

```
1-best:
  AFTER EARLY NIGHTFALL THE YELLOW LAMPS WOULD LIGHT UP HERE AND THERE THE SQUALID QUARTER OF THE BROTHELS

n-best (5 hypotheses):
  [0] score=-1.0923   ... OF THE BROTHELS
  [1] score=-2.3082   ... OF THE BROTHALS
  [2] score=-2.5559   ... OF THE BROFFLES
  [3] score=-2.9750   ... OF THE BRIFFLES
  [4] score=-12.2809  ... OF THE BROFFLES'
```

Checked on both wavs: 1-best text is byte-identical between `num_return_paths=1` and `=5`; `hypotheses[0]` matches the top-level text, tokens and timestamps; scores are non-increasing; `len(timestamps) == len(tokens) == len(ys_log_probs)` per hypothesis; `num_return_paths=50` with `max_active_paths=4` returns 4; `greedy_search` is unaffected and returns an empty list.

### Notes / open questions

- Only the Python bindings are wired up here. Happy to add C / Go / Kotlin / Swift / C# in this PR or a follow-up — let me know which you want and how you would prefer the n-best represented in the C API.
- `AsJsonString()` does not include `hypotheses` yet. I left it out to avoid changing existing JSON output; glad to add it behind the same flag if you would like it.

## Summary by CodeRabbit

- **New Features**
  - Added n-best decoding for offline transducer recognition using modified beam search.
  - Configure the number of returned hypotheses, with results including text, tokens, timestamps, log probabilities, and scores.
  - Added Python API support through `num_return_paths` and `result.hypotheses`.
  - Hypotheses receive the same text normalization and formatting as the top result.
- **Documentation**
  - Added a Python example demonstrating offline n-best decoding and displaying ranked hypotheses with scores.

### DIFF
--- a/sherpa-onnx/csrc/offline-recognizer-transducer-impl.h
+++ b/sherpa-onnx/csrc/offline-recognizer-transducer-impl.h
@@ -5,6 +5,7 @@
 #ifndef SHERPA_ONNX_CSRC_OFFLINE_RECOGNIZER_TRANSDUCER_IMPL_H_
 #define SHERPA_ONNX_CSRC_OFFLINE_RECOGNIZER_TRANSDUCER_IMPL_H_
 
+#include <algorithm>
 #include <fstream>
 #include <ios>
 #include <memory>
@@ -31,18 +32,18 @@
 
 namespace sherpa_onnx {
 
-static OfflineRecognitionResult Convert(
-    const OfflineTransducerDecoderResult &src, const SymbolTable &sym_table,
-    int32_t frame_shift_ms, int32_t subsampling_factor) {
-  OfflineRecognitionResult r;
-  r.tokens.reserve(src.tokens.size());
-  r.timestamps.reserve(src.timestamps.size());
-  r.durations.reserve(src.durations.size());
+// Converts token IDs to their symbols, appending the detokenized text to
+// *text. Shared by the 1-best result and every entry of the n-best list.
+static std::vector<std::string> ConvertTokens(
+    const std::vector<int64_t> &token_ids, const SymbolTable &sym_table,
+    std::string *text) {
+  std::vector<std::string> tokens;
+  tokens.reserve(token_ids.size());
 
-  std::string text;
-  for (auto i : src.tokens) {
+  std::string t;
+  for (auto i : token_ids) {
     auto sym = sym_table[i];
-    text.append(sym);
+    t.append(sym);
 
     if (sym.size() == 1 && (sym[0] < 0x20 || sym[0] > 0x7e)) {
       // for bpe models with byte_fallback,
@@ -54,21 +55,38 @@ static OfflineRecognitionResult Convert(
       sym = os.str();
     }
 
-    r.tokens.push_back(std::move(sym));
+    tokens.push_back(std::move(sym));
   }
+
   if (sym_table.IsByteBpe()) {
-    text = sym_table.DecodeByteBpe(text);
+    t = sym_table.DecodeByteBpe(t);
   }
 
-  text = RemoveSpaceBetweenCjk(text);
+  *text = RemoveSpaceBetweenCjk(t);
 
-  r.text = std::move(text);
+  return tokens;
+}
+
+static std::vector<float> ConvertTimestamps(
+    const std::vector<int32_t> &timestamps, float frame_shift_s) {
+  std::vector<float> ans;
+  ans.reserve(timestamps.size());
+  for (auto t : timestamps) {
+    ans.push_back(frame_shift_s * t);
+  }
+  return ans;
+}
+
+static OfflineRecognitionResult Convert(
+    const OfflineTransducerDecoderResult &src, const SymbolTable &sym_table,
+    int32_t frame_shift_ms, int32_t subsampling_factor) {
+  OfflineRecognitionResult r;
+  r.durations.reserve(src.durations.size());
+
+  r.tokens = ConvertTokens(src.tokens, sym_table, &r.text);
 
   float frame_shift_s = frame_shift_ms / 1000. * subsampling_factor;
-  for (auto t : src.timestamps) {
-    float time = frame_shift_s * t;
-    r.timestamps.push_back(time);
-  }
+  r.timestamps = ConvertTimestamps(src.timestamps, frame_shift_s);
 
   // Copy durations (if present)
   for (auto d : src.durations) {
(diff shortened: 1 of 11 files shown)


csukuangfj:

Please use
https://github.com/k2-fsa/sherpa-onnx/blob/f007b559b75c45c04f246245ef7bbc5cd193307d/sherpa-onnx/python/csrc/offline-stream.cc#L119-L126
to pass the argument `num_return_paths`. If you like, you can also pass `max_active_paths` via the set option so that it overrides the default value from the recongizer config.

sherpa-onnx/csrc/offline-stream.h
```diff
@@ -16,6 +16,27 @@
 
 namespace sherpa_onnx {
 
+/// One entry of an n-best list. Produced only by offline transducer models
+/// decoded with modified_beam_search and num_return_paths > 1.
+struct OfflineRecognitionHypothesis {
```
csukuangfj, line 21: Please also add a method  `AsJsonString` to it and call it inside `OfflineRecognitionResult::AsJsonString`.

sherpa-onnx/csrc/offline-recognizer-transducer-impl.h
```diff
@@ -31,18 +32,18 @@
...
 namespace sherpa_onnx {
 
-static OfflineRecognitionResult Convert(
-    const OfflineTransducerDecoderResult &src, const SymbolTable &sym_table,
-    int32_t frame_shift_ms, int32_t subsampling_factor) {
-  OfflineRecognitionResult r;
-  r.tokens.reserve(src.tokens.size());
-  r.timestamps.reserve(src.timestamps.size());
-  r.durations.reserve(src.durations.size());
+// Converts token IDs to their symbols, appending the detokenized text to
```
csukuangfj, line 35: Is `*text` overwritten, not appended?

csukuangfj (changes requested):

PR #3898, opened 2026-08-26
Add Nabra-82M Arabic TTS conversion scripts for Kokoro runtime

marwanelamami:
Follows issue #3897.

[oddadmix](https://huggingface.co/oddadmix) trained Nabra-82M ([oddadmix/Nabra-82M-v0.1](https://huggingface.co/oddadmix/Nabra-82M-v0.1)), a Kokoro/StyleTTS2 model fine-tuned for Modern Standard Arabic. I quantized and packaged it. Its ONNX export shares the Kokoro v1.0 input signature (`tokens` int64[1,T], `style` float32[1,256], `speed` float32[1]), so it runs on the existing kokoro runtime with no code changes. I verified sherpa's espeak-ng token output matches the training pipeline's symbol-for-symbol on several sentences.

This PR adds conversion scripts under `scripts/kokoro/nabra/`:

- `add_meta_data.py`: stamps sherpa metadata (`model_type=kokoro`, `voice=ar`, `style_dim=510,1,256`, etc.) and bakes an FIR notch (Conv1d, 129 taps) at 4800/9600 Hz into the graph tail to remove iSTFT image tones
- `run.sh`: end-to-end conversion (metadata + tokens.txt + voices.bin)
- `README.md`: usage example with `OfflineTtsKokoroModelConfig`

Pre-packaged models (int4/int8/fp16) are at https://huggingface.co/marwanelamami/nabra-82m-sherpa-onnx if you'd rather link than host.

Tested with sherpa-onnx 1.13.6 Python API; RTF ~0.66-0.68 @ 4 threads on Snapdragon 865/870.

## Summary by CodeRabbit

* **New Features**
  * Added documentation for converting Nabra-82M Arabic speech models for use with the Kokoro runtime.
  * Added a tool to prepare Nabra-82M ONNX models with the required Arabic voice metadata and audio filtering.
  * Documented how to generate optimized model assets, configure Arabic text-to-speech, and test speech generation with sherpa-onnx.
  * Improved support for producing high-quality Arabic speech output at supported audio sample rates.

PR #3954, opened 2026-09-15
Add conversion scripts for 7M Kokoro distills (Arabic + English)

marwanelamami:
## What this adds

Conversion scripts for two new 7M-parameter Kokoro distills by [oddadmix](https://huggingface.co/oddadmix), in `scripts/kokoro/7m-distill/`:

| upstream | language | converted bundle |
|---|---|---|
| [Nabra-7M-Distill](https://huggingface.co/oddadmix/Nabra-7M-Distill) | Arabic (MSA) | [nabra-7m-distill-sherpa-onnx](https://huggingface.co/marwanelamami/nabra-7m-distill-sherpa-onnx) |
| [Kokoro-7M-Distill](https://huggingface.co/oddadmix/Kokoro-7M-Distill) | English (US) | [kokoro-7m-distill-sherpa-onnx](https://huggingface.co/marwanelamami/kokoro-7m-distill-sherpa-onnx) |

Both use the standard Kokoro I/O contract and load with `OfflineTtsKokoroModelConfig` unchanged — no runtime changes in this PR. Sizes are ~30 MB FP32 and ~8.7 MB INT8, against 82M / 325 MB for stock Kokoro, which makes them interesting for the mobile targets.

This follows the same structure as #3898 (Nabra-82M), which is still open.

## Samples

Arabic — [sala](https://huggingface.co/marwanelamami/nabra-7m-distill-sherpa-onnx/resolve/main/samples/sala_fp32.wav) · [ilm](https://huggingface.co/marwanelamami/nabra-7m-distill-sherpa-onnx/resolve/main/samples/ilm_fp32.wav) · [short](https://huggingface.co/marwanelamami/nabra-7m-distill-sherpa-onnx/resolve/main/samples/short_fp32.wav)
English — [fox](https://huggingface.co/marwanelamami/kokoro-7m-distill-sherpa-onnx/resolve/main/samples/fox_fp32.wav) · [greet](https://huggingface.co/marwanelamami/kokoro-7m-distill-sherpa-onnx/resolve/main/samples/greet_fp32.wav) · [short](https://huggingface.co/marwanelamami/kokoro-7m-distill-sherpa-onnx/resolve/main/samples/short_fp32.wav)

INT8 renders of the same sentences are in each repo alongside.

## Conversion notes

**iSTFT image tones.** The ISTFTNet decoder leaves narrow tones at 4.8 kHz and 9.6 kHz (sr/5 and 2·sr/5 for its 20-point inverse STFT). `add_meta_data.py` bakes a 65-tap FIR notch into the graph, so every runtime gets filtered audio with no post-processing. Band ratios drop from 1.46 / 4.71 to 0.005 / 0.032. The Conv is inserted at the consumer's index rather than appended — appending violates ONNX topological ordering and `onnx.checker` rejects the result.

**tokens.txt comes from the trained vocab, not the official table.** The two disagree: this English distill maps `A` → 24 where the official 178-token table uses 17, and the Arabic distill parks the MSA pharyngeals ʕ (ع) and ħ (ح) on free embedding slots 7 and 8. Using the official file silently mismaps embeddings.

**voices.bin is af_msa for both.** Upstream distilled the English student against `af_msa` as well, and reports WER 0.0525 → 0.0701 when Kokoro's own `af_heart` is substituted.

**INT8 quality.** Log-mel distance from FP32 is 0.069 (Arabic) and 0.074 (English). For calibration on the same metric: the 82M INT8 build measures 0.124 and the 82M INT4 build measures 0.284.

## One runtime limitation worth flagging

Arabic works with the built-in espeak front-end — upstream's `arabic_g2p.py` is itself an espeak wrapper, so sherpa lands in the same place (duration delta 0.0–2.3% against the reference pipeline).

English does not. misaki, Kokoro's own front-end, uses single-character diphthongs — `W` = /aʊ/, `A` = /eɪ/, `O` = /oʊ/ — that espeak-ng never emits. Sherpa's built-in path therefore feeds the model a token stream it was never trained on, and durations come out up to 37% short. At 82M this is absorbed; at 7M it is not.

The official `lexicon-us-en.txt` from `kokoro-multi-lang-v1_0` *does* use those symbols, and all 178,401 entries resolve cleanly against this model's `tokens.txt` with zero unknown phonemes. But the lexicon branch is unreachable for a single-language model:

```c++
// kokoro-multi-lang-lexicon.cc — lexicon only consulted when voice is empty
if (!voice.empty()) {
  return ConvertTextToTokenIDsWithEspeak(text, voice);
}
```

```c++
// offline-tts-kokoro-model.cc — an absent voice key becomes "en-us"
SHERPA_ONNX_READ_META_DATA_STR_WITH_DEFAULT(meta_data_.voice, "voice", "en-us");
```

An absent `voice` defaults to `"en-us"`; a present-but-empty one trips the macro's validity check and exits. So there is no metadata combination that reaches the lexicon path unless the model is multi-lingual and no voice is passed.

I did not change runtime behaviour in this PR since it affects the shared Kokoro path. Options, if you'd like me to follow up:

1. Let `version >= 2` single-language models opt into the lexicon when `--kokoro-lexicon` is passed explicitly.
2. Add a metadata flag (e.g. `use_lexicon=1`) that takes precedence over the voice default.
3. Treat an explicitly-passed `--kokoro-lexicon` as intent to use the lexicon regardless of voice.

Happy to implement whichever you prefer. In the meantime both bundles ship `g2p_frontend.py` as the reference path, and the English bundle also ships `lexicon-us-en.txt` so it's ready the moment the runtime can read it.

## Testing

- Both models load and generate through `sherpa_onnx.OfflineTts` in FP32 and INT8 (v1.13.8)
- ONNX export verified against the PyTorch pipeline: identical output lengths (3.175 s both, English fox sentence); residual waveform difference is SineGen's stochastic excitation, which correlates at 0.979 with the RNG zeroed
- `generate_assets.py` reproduces the published `tokens.txt` and `voices.bin` byte-for-byte for both models
- `add_meta_data.py` reproduces the published notch-baked graph; `onnx.checker.check_model` passes

Credit for the models goes to oddadmix; these scripts only convert and package them.

## Summary by CodeRabbit

- **New Features**
  - Added support for converting Arabic and English 7M Kokoro distill models into sherpa-onnx bundles.
  - Added Arabic and English text-to-speech front ends.
  - Added FP32 and INT8 bundle generation, metadata, vocabularies, voice assets, and English pronunciation resources.
  - Added audio filtering to remove narrow-band artifacts from converted models.

- **Documentation**
  - Added setup and usage instructions for the complete conversion workflow.

- **Tests**
  - Added export validation and smoke tests that generate WAV samples and report audio output details.


marwanelamami:
Thanks — all four were real. Fixed in d7f35d6.

**`add_meta_data.py` — rerun corrupts the model (and it was worse than described).** Confirmed by running the script twice on the same file: the second pass wires the Conv's output back in as its own input *and* duplicates the initializer. Since `onnx.save()` ran before `check_model()`, the corrupt graph was already on disk when the checker raised:

```
notch convs: 2
   in: ['/decoder/generator/Slice_3_output_0', 'notch_fir_w'] out: ['audio_notched']
   in: ['audio_notched', 'notch_fir_w'] out: ['audio_notched']   <- self-loop
ValidationError: notch_fir_w initializer name is not unique
```

`bake_notch()` is now idempotent, a partially-notched graph is rejected before any mutation, and `check_model()` runs *before* `save()` so a failure can't destroy a working file. Three consecutive runs now leave exactly one Conv and one initializer, checker passing.

**`add_meta_data.py` — silent failure on missing `Squeeze_3`.** Correct, and I'd hit this myself during development without noticing the implication: it printed `notch baked: False` and then `saved + checked`, exit 0. Now aborts before stamping. Verified against a model with the node renamed — exits 1 with a clear message.

**`export_onnx.py` — validation didn't gate the artifact.** Fixed, though not quite as suggested. Empty output and length mismatch now abort. For the numerical check, an absolute/relative tolerance isn't meaningful for this decoder — SineGen draws a fresh random excitation per run, so two *torch* runs of the same input differ more than torch differs from ONNX:

```
torch vs torch run1: max_abs=7.665e-01 rel=1.594e+00
torch vs torch run2: max_abs=5.540e-01 rel=1.152e+00
torch vs onnx      : max_abs=7.127e-01 rel=1.482e+00
```

Any tolerance loose enough to pass a correct export would pass almost anything. I used normalised-spectrum cosine instead, which is stable under that randomness:

```
16tok  torch/torch=0.9137  torch/onnx=0.9404
37tok  torch/torch=0.8877  torch/onnx=0.9153
negative control (noise):   0.2262
```

Gate is set at 0.70, clear of both. A real export scores 0.9432.

**`g2p_frontend.py` — wrong default path.** Correct, `parent.parent` resolved to `scripts/kokoro` while `run.sh` downloads into `scripts/kokoro/7m-distill`. Applied your suggestion; both zero-argument constructors now work from the script directory.

csukuangfj:
> I did not change runtime behaviour in this PR since it affects the shared Kokoro path.

We are removing espeak-ng and users can use a external phonemization tool to pass tokens directly to sherpa-onnx.

csukuangfj (approved):

PR #3957, opened 2026-09-17
Ship arm64 libraries in the sherpa_onnx_windows Flutter package

Ranjan-Bhagat:
## Summary

The `sherpa_onnx_windows` Flutter package bundles x64 DLLs only, so a Flutter app built for Windows on ARM (`flutter build windows` on an arm64 machine) ships x64 speech libraries that cannot load in an arm64 process. sherpa-onnx already builds and releases Windows arm64 libraries on every tag; they were just not in the package.

This PR:

- adds a `sherpa_onnx_windows_arm64` job to `release-dart-package.yaml` that builds the arm64 libraries on the `windows-11-arm` runner with the same options as the x64 build (MSVC setup and `-A ARM64` copied from `windows-arm64.yaml`), and has the existing `sherpa_onnx_windows` job download them into `windows/arm64/` before publishing
- makes the package's `windows/CMakeLists.txt` pick `windows/arm64/` when building for arm64 (`FLUTTER_TARGET_PLATFORM=windows-arm64`, or the `-A ARM64` generator platform Flutter passes) and `windows/` otherwise
- documents the layout in `windows/README.md`

x64 stays where it is, so nothing changes for existing users, and the test workflows that copy freshly built DLLs into `windows/` (`test-flutter*.yaml`, `test-dart.yaml`) keep working unchanged. The CMake floor stays at 3.14 as the file asks. `CMAKE_SYSTEM_PROCESSOR` is deliberately not used: it describes the machine running CMake, so an x64 build on an arm64 machine would pick the wrong folder.

Package size: the published archive grows from about 8.7 MB to about 17 MB (the arm64 pair is 6.6 MB compressed), well inside the pub.dev limit.

## Testing

Assembled the package the way the release job will (v1.13.4 x64 DLLs from the pub.dev archive at `windows/`, the v1.13.4 `win-arm64-shared-MD-Release` DLLs at `windows/arm64/`) and used it as a path dependency of a real Flutter app:

- `flutter build windows --release` on a Windows 11 arm64 machine (Flutter 3.47.0, VS Build Tools 2022 17.14): builds, and the bundled `sherpa-onnx-c-api.dll` and `onnxruntime.dll` in `build/windows/arm64/runner/Release` have PE machine `0xAA64` (ARM64).
- The selection logic, configured directly with CMake (a tiny project that `add_subdirectory`s the package's `windows/` and prints `sherpa_onnx_windows_bundled_libraries`):
  - Visual Studio generator on the arm64 machine: `-A ARM64` selects `windows/arm64/*.dll`, `-A x64` selects `windows/*.dll`
  - `-DFLUTTER_TARGET_PLATFORM=windows-arm64` selects `windows/arm64/*.dll`; `windows-x64`, or the variable unset, selects `windows/*.dll`
- The same arm64 DLL pair has been running in a shipped Flutter app on Windows on ARM since its 1.10.0 release, through a local override of this package.

Not tested here: the new release job itself, which only runs on a tag. It is the configure and build block from `windows-arm64.yaml`, which runs on every push.

## Summary by CodeRabbit

* **New Features**
  * Added Windows ARM64 support for the `sherpa_onnx_windows` package.
  * Published packages now include architecture-specific DLLs for x64 and ARM64.
  * Builds automatically select the libraries matching the Flutter app’s target architecture.

* **Documentation**
  * Updated Windows package documentation to describe x64 and ARM64 library support.


csukuangfj (approved):

PR #3958, opened 2026-09-17
fix(supertonic): allow the "na" language tag

utrumo:
`kSupertonicAvailableLangs` lists 31 language tags, but the Supertonic model also accepts `na`.

The upstream Supertonic reference implementation includes `na` in its language list (`AVAILABLE_LANGS` in `rust/src/helper.rs`) and handles it like any other tag: the same `<lang>text</lang>` wrapping and tokenizer. Only sherpa-onnx's validation list rejected it.

This matters for mixed-language text, e.g. Russian sentences with English words in them: with `ru` forced, the English words are read poorly; with `na` the model handles both.

Tested on Android (arm64) with the Supertonic 3 int8 model: synthesis succeeds with `na`, `ru` and `en`.

## Summary by CodeRabbit

- **New Features**
  - Added support for the `na` language identifier in offline Supertonic text-to-speech generation.
  - The supported language list now includes 32 language identifiers.


csukuangfj (approved):
Thank you for your contribution.

`na` was added in https://github.com/k2-fsa/sherpa-onnx/pull/3958/changes, after we have merged  the PR for supertonic 3 tts.

PR #3962, opened 2026-09-18
Canary: derive lang2id from the model vocab, warn instead of silently falling back to English

### LINKED ISSUE #3961: Canary: lang2id is hardcoded to en/es/de/fr, so any other language silently falls back to English
sherpa-onnx v1.13.8 (and master as of 2026-09-17), Python API, CPU.

`OfflineRecognizer.from_nemo_canary` only honors four languages: `en`, `es`, `de`, `fr`. For any other `src_lang`/`tgt_lang` the initial decoder input silently falls back to the `en` token, so multilingual canary exports such as nvidia/canary-1b-v2 (25 European languages, sherpa-layout int8 export exists at Sarphix/canary-1b-v2-sherpa-onnx-int8) transcribe or translate into English regardless of the requested language.

Repro (Italian voice message, canary-1b-v2 int8 export):

```python
import sherpa_onnx
r = sherpa_onnx.OfflineRecognizer.from_nemo_canary(
    encoder="encoder.int8.onnx", decoder="decoder.int8.onnx", tokens="tokens.txt",
    src_lang="it", tgt_lang="it", feature_dim=128, num_threads=4)
# -> English translation-style output ("Hello, my name is Carlo ..."),
#    identical to src_lang="it", tgt_lang="en" and to src_lang="en".
# fr/fr, es/es and de/de decode correctly on the same export.
```

Root cause: `sherpa-onnx/csrc/offline-recognizer-canary-impl.h` builds `meta.lang2id` from a hardcoded four-entry map (lines 303-306 at v1.13.8), and `GetInitialDecoderInput` falls back to `lang2id.at("en")` for any src/tgt missing from it (lines 263-275). The model's own `tokens.txt` carries the other language tokens (`<|it|>` etc.), and the encoder metadata is otherwise sound.

Suggested fix: derive `lang2id` from the model's token table (or an encoder-metadata language list) instead of the hardcoded map, so each export declares the languages it supports. At minimum, an unknown `src_lang`/`tgt_lang` could log a warning instead of silently switching the target language.

We ship sherpa-onnx on Android for on-device transcription and hit this while evaluating canary-1b-v2 for Italian; I can test a patch against the int8 export.

## Environment

- sherpa-onnx 1.13.8 (Python, CPU), also present on master (2026-09-17)
- Model: Sarphix/canary-1b-v2-sherpa-onnx-int8 (nvidia/canary-1b-v2 base)

paoloantinori:
## What

The canary recognizer hardcoded `lang2id` to the four languages of canary-180m-flash (en, es, de, fr). Any other `src_lang`/`tgt_lang` silently fell back to English: multilingual exports such as [canary-1b-v2](https://huggingface.co/nvidia/canary-1b-v2) (25 European languages) could not transcribe in their languages through sherpa-onnx.

`PostInit()` now derives `lang2id` from the loaded vocabulary: every `<|xx|>` token with a two-letter code becomes a language entry. The hardcoded language whitelist in `OfflineCanaryModelConfig::Validate()` is gone, so CLI and JNI callers accept every vocabulary-defined language.

Languages are resolved per stream: a stream option `src_lang`/`tgt_lang` (via the existing `OfflineStream` option mechanism) overrides the recognizer-level config, so one recognizer can decode streams with different languages. Unknown codes warn and fall back to `en` (or, without `<|en|>`, the language with the lowest token id) instead of switching silently; an empty code keeps the historical silent `en` behavior. A vocab with no language tokens fails at load time.

## Why

Reported in #3961 with a repro on the Sarphix canary-1b-v2 int8 export: `src_lang=it, tgt_lang=it` produced English translation-style output, and the model's own `tokens.txt` already carries `<|it|>` and the other language tokens.

## Backward compatibility

canary-180m-flash exports carry exactly `<|en|>`, `<|es|>`, `<|de|>`, `<|fr|>`, so the derived map is identical to the previous hardcoded map. Nothing changes for existing 180m users.

## Tests

- `offline-recognizer-canary-impl-test`: unit tests for the derivation (language tokens captured, bracketed specials like `<|pnc|>` and timestamp tokens rejected) and for the resolution paths (passthrough, empty code, unknown code, deterministic no-en fallback).
- End to end with the Sarphix canary-1b-v2 int8 export and a 6-second Italian voice message:

| src/tgt | before | after |
|---|---|---|
| it, it | English translation output | "Ciao. Il mio nome è Carlo Buoncompagni e vivo a Castelazzo Novarese..." |
| de, de | correct German | correct German (unchanged) |
| xx, xx | silent English | warning logged, English fallback |
| de config + it stream | (no per-stream API) | Italian transcript, same recognizer |

Fixes #3961

## Summary by CodeRabbit

- **New Features**
  - Canary speech recognition now derives supported languages from each model’s vocabulary.
  - Custom language codes are accepted when available in the model vocabulary.
  - Per-stream source and target language settings are supported, with deterministic fallback for unsupported or unspecified languages.

- **Documentation**
  - Added guidance for producing and verifying reproducible F-Droid Android builds.
  - Updated Canary language configuration documentation.

- **Tests**
  - Added coverage for language discovery, resolution, and fallback scenarios.


sherpa-onnx/csrc/offline-recognizer-canary-impl.h
```diff
@@ -260,19 +325,10 @@ class OfflineRecognizerCanaryImpl : public OfflineRecognizerImpl {
...
-      decoder_input[3] = meta.lang2id.at(canary_config.src_lang);
-    }
-
-    if (canary_config.tgt_lang.empty() ||
-        !meta.lang2id.count(canary_config.tgt_lang)) {
-      decoder_input[4] = meta.lang2id.at("en");
-    } else {
-      decoder_input[4] = meta.lang2id.at(canary_config.tgt_lang);
-    }
+    decoder_input[3] =
```
csukuangfj, line 341: Can you use https://github.com/k2-fsa/sherpa-onnx/blob/a5b4a944c5186a68bcdc0ac3011e4c541781ac84/sherpa-onnx/csrc/offline-stream.h#L131
to test if a stream has set `src_lang` and `tgt_lang`, if yes, use
https://github.com/k2-fsa/sherpa-onnx/blob/a5b4a944c5186a68bcdc0ac3011e4c541781ac84/sherpa-onnx/csrc/offline-stream.h#L135
to get it, otherwise, fallback to the one in canary_config?

In this way, users can set different src_lang and target_lang in different streams, with the same recognizer.

csukuangfj (changes requested):

sherpa-onnx/csrc/offline-recognizer-canary-impl.h
```diff
@@ -300,10 +356,9 @@ class OfflineRecognizerCanaryImpl : public OfflineRecognizerImpl {
     config_.feat_config.window_type = "hann";
     config_.feat_config.is_librosa = true;
 
-    meta.lang2id["en"] = symbol_table_["<|en|>"];
-    meta.lang2id["es"] = symbol_table_["<|es|>"];
-    meta.lang2id["de"] = symbol_table_["<|de|>"];
-    meta.lang2id["fr"] = symbol_table_["<|fr|>"];
+    // Derived from the vocab (canary-1b-v2 and future multilingual exports
+    // carry more than the four languages the 180m-flash port hardcoded).
+    meta.lang2id = DeriveCanaryLang2Id(symbol_table_.sym2id());
```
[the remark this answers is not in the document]
csukuangfj: please fix it.

paoloantinori:
Thank you for the review. Addressed in 44e8d9f5, verified end to end with the Sarphix canary-1b-v2 int8 export:

- **Per-stream languages**: `GetInitialDecoderInput` now reads the stream's `src_lang`/`tgt_lang` options via `HasOption`/`GetOption` and falls back to `canary_config` when unset. Verified with one recognizer built with `de`/`de`: a stream with `src_lang=it, tgt_lang=it` on Italian audio produces the Italian transcript, a plain stream on German audio produces German.
- **Hardcoded validation removed** from `OfflineCanaryModelConfig::Validate()` so CLI/JNI callers accept every vocabulary-defined language (Copilot's finding). Unsupported recognizer-level languages now warn once at load time; unknown per-stream codes warn at decode time and fall back. A vocab with no language tokens fails at load instead of exiting from the decode path.
- **Deterministic fallback**: without `<|en|>` the fallback is the language with the lowest token id (Copilot's finding), pinned by a unit test.

On the CodeRabbit suggestion to use a warning-level logger: the project's `macros.h` defines only `SHERPA_ONNX_LOGE`, so there is no warning logger to switch to; adding one would touch all three platform branches. I kept `SHERPA_ONNX_LOGE`, which is also what the neighboring recoverable condition in `SelectCanaryFirstToken` uses. Happy to add a `SHERPA_ONNX_LOGW` family in a separate PR if you want it.

`ParseCanaryLangToken` intentionally accepts lowercase two-letter codes only: the canary vocabularies use lowercase ISO 639-1 codes, and `<|pnc|>`-style specials must stay excluded.

paoloantinori:
The CodeRabbit caching finding is fixed in fa51a118: `PostInit` and `SetConfig` resolve `src_lang`/`tgt_lang` once and cache the ids; `GetInitialDecoderInput` reuses the cached ids and only resolves options the stream actually set, so an unsupported recognizer-level language warns at load time instead of on every stream. Verified with the canary-1b-v2 export: the per-stream matrix is unchanged and the fallback warning now appears exactly once per unknown code.

paoloantinori:
You are right, I had missed the "and its option help/docs" part of that finding. Fixed in c62d2618: the `--canary-src-lang`/`--canary-tgt-lang` CLI help and the `from_nemo_canary` docstrings (both `src_lang` and `tgt_lang`) now say the valid codes come from the model's `tokens.txt`, and describe the unknown-code warn-and-fallback behavior instead of listing the old four codes.

android/fdroid-reproducible-builds.md
```diff
@@ -0,0 +1,107 @@
+# Reproducible F-Droid builds of the Android library (AAR)
```
csukuangfj: Can you create a separate PR for this file?
Also, can you create a GitHub Actions YAML file for it?

paoloantinori:
Done in 3e0b554e. android/fdroid-reproducible-builds.md is removed from this PR. The page lives in #3963 together with a GitHub Actions workflow that builds the AAR with the flags it documents and uploads it as an artifact. This PR is canary-only again.

csukuangfj (approved):

PR #3963, opened 2026-09-18
Docs: reproducible F-Droid builds of the Android library (AAR)

### LINKED ISSUE #3888: Docs/recipe contribution: reproducible F-Droid source builds of the Android library (srclib + NDK + flags pitfalls)
We maintain [Anti-Vocale](https://github.com/RisorseArtificiali/anti-vocale), an on-device ASR app on F-Droid that consumes sherpa-onnx as a source-built library through an fdroiddata srclib. Getting that path to a **byte-reproducible** build cost us several iterations (thank you @linsui for the review that set us on the right track), and none of the pitfalls are documented anywhere. We would like to contribute the recipe back - as a docs page, a script under `scripts/`, or a CI workflow, whichever shape you prefer. The content below is ready to be adapted; everything in it is battle-tested (our F-Droid releases verify byte-identical against our reference builds).

## Building the sherpa-onnx AAR for a reproducible F-Droid recipe

Context: F-Droid builds apps from source; a prebuilt AAR dependency must be replaced by an `srclib` entry checked out at a pinned commit, and the resulting APK must be byte-identical to the developer's reference build (for `AllowedAPKSigningKeys`/reproducible verification).

1. **Pin the srclib commit exactly, and keep it in sync with your release**. Our app pins `SHERPA_ONNX_VERSION` in its fetch script, the fdroiddata recipe pins the same commit in the srclib block, and a version-marker file ties both to the tag. Desync here is the #1 cause of verification failures.

2. **Build flags that matter for consumers**: `SHERPA_ONNX_ENABLE_C_API=OFF` + `SHERPA_ONNX_ENABLE_JNI=ON`. Without the first, the AAR ships extra `libsherpa-onnx-c-api.so`/`libsherpa-onnx-cxx-api.so` that a prebuilt AAR does not contain, and the byte-comparison fails on unexpected files. This bit us as "everything differs byte-for-byte" until we diffed the file lists.

3. **NDK resolution**: let fdroidserver install the NDK itself (the runner resolves `ndk: r27` to 27.0.12077973 at the time of writing). Do NOT preinstall or pin a different NDK revision in your reference build to "match what CI used": the resolution drifts, and the toolchain difference changes every `.so`. Pin the *recipe line*, build with what fdroid installs, and never hardcode an observed resolution.

4. **Disable AGP's baseline-profile generation** (`android { buildFeatures { } }` / the ArtProfile task): `baseline.prof`/`baseline.profm` are nondeterministic across builds (ordering variation), which breaks byte-reproducibility for no functional loss on a library consumer.

5. **Alignment/signature parity for the verification**: consumers doing reproducible verification use `reproducible-apk-tools` (`zipalign --page-size 16 --pad-like-apksigner`) + `apksigcopier` on the APK side; the library side needs nothing special once 2-4 hold.

6. **ABI splits**: the fdroiddata recipe builds per-ABI (separate versionCodes); make sure the sherpa build produces per-ABI artifacts rather than a fat AAR, or the per-ABI APKs will differ only in unused bytes.

Our working references: the fdroiddata recipe history for `com.antivocale.app` (commit "Update Anti-Vocale to 374") and our [release runbook](https://github.com/RisorseArtificiali/anti-vocale/blob/main/docs/release-runbook.md).

Happy to turn this into a PR against `docs/` (e.g. a "Reproducible F-Droid builds" page) or a `scripts/build-android-fdroid.sh` - tell us the preferred shape.

paoloantinori:
Adds `android/fdroid-reproducible-builds.md`: how to build the Android library (AAR) so that an F-Droid srclib recipe produces APKs byte-identical to a reference build.

Content, from our battle-tested Anti-Vocale packaging (F-Droid releases verified byte-identical):

- commit pinning across the app fetch script, the fdroiddata srclib block, and version markers
- build flags: `SHERPA_ONNX_ENABLE_C_API=OFF`, `SHERPA_ONNX_ENABLE_JNI=ON`
- NDK pinning (llvm-strip output varies across NDK versions)
- disabling the consumer-side baseline-profile generation (ordering variation breaks byte comparison)
- per-ABI artifacts
- the `reproducible-apk-tools` verification recipe

Cross-linked from `android/README.md`. Closes #3888.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for producing and verifying reproducible F-Droid Android builds.
  * Documented pinned source and NDK versions, required build settings, APK verification steps, signing-key handling, and multi-ABI packaging.

* **Build Automation**
  * Added a workflow that builds the Android AAR for supported CPU architectures and uploads a reproducible release artifact.
  * Updated the workflow to use the pinned Android NDK version and require manual triggering.


paoloantinori:
Thank you. Addressed in 41fc756:

- **Signing key vs reproducibility**: the intro now keeps the two checks apart. `AllowedAPKSigningKeys` pins the accepted signing key; the byte-for-byte comparison is fdroidserver's verification of the rebuild against the reference APK, with the commands in section 6.
- **Exact NDK revision**: section 3 now pins the full revision (`ndkVersion = "27.0.12077973"`, r27b/c/d differ in llvm-strip) on both the app side and the recipe's `ndk` field.
- **Executable baseline-profile disable**: section 4 gives the exact Gradle snippet we ship (`whenTaskAdded` disabling `ArtProfile` tasks) with the `afterEvaluate` trap spelled out.
- **Per-ABI**: section 5 corrected. A fat library AAR is fine; the per-ABI artifacts come from the consumer's APK splits (`splits { abi { ... } }`, no universal APK) plus the base*10+arch versionCode mapping, and AGP strips unused library ABIs per split.
- **Complete verification commands**: section 6 now lists `zipalign.py`, `apksigcopier copy`, and `apksigcopier compare` with their inputs, and points at each tool's README for the full flag set.

On the last comment (deterministic no-`en` fallback in `ResolveCanaryLang`): that is code from #3962, not this docs page; it is already fixed there (fallback = the language with the lowest token id, pinned by a unit test).

paoloantinori:
Fixed in cace60f, with both commands verified against the tools' READMEs this time. Two corrections to my original block, and one to the suggested fix:

- `zipalign.py` requires an output APK, and the page size is `-P` with the size in bytes (`-p` alone means 4096). The corrected call is `zipalign.py -P 16384 --pad-like-apksigner rebuilt-unsigned.apk rebuilt-aligned-unsigned.apk`. The `--page-size 16` form proposed in the review does not exist in `zipalign.py`'s usage line.
- `apksigcopier copy` takes three positional APKs (`signed unsigned output`, per its README); the `--v1-only` flag is dropped from the example since the plain form is what the tool documents. `compare` unchanged.

android/fdroid-reproducible-builds.md
```diff
@@ -0,0 +1,107 @@
+# Reproducible F-Droid builds of the Android library (AAR)
```
csukuangfj, line 1: Can you create a GitHub Actions yaml file for it?

You can build an AAR file and upload it as an artifact.

.github/workflows/android-fdroid-reproducible-aar.yaml
```diff
@@ -0,0 +1,138 @@
+name: android-fdroid-reproducible-aar
```
csukuangfj, line 1: Can you post the link to this workflow run in your repository so we can check the logs?

paoloantinori:
Added `.github/workflows/android-fdroid-reproducible-aar.yaml`. It builds the four ABIs with the flags from the page (C API off, JNI on, which are the scripts' defaults), pins NDK 27.2.12479018 (r27c, the revision the fdroiddata blocks pin) instead of the floating latest, assembles the AAR and uploads it as an artifact. The trigger is workflow_dispatch only, so it does not add a build to every master push; anyone can dispatch it, forks included. Run: https://github.com/paoloantinori/sherpa-onnx/actions/runs/35506035680

I also rebased the branch on master, so the PR no longer carries the canary commits from #3962.

One correction to my earlier comment. zipalign.py takes the page size in KiB (`-P, --page-size` is documented as "use N*1024-byte memory page alignment for .so files"), so the value is 16, and the --page-size form does exist; the usage line just abbreviates it to [-P N]. The page now reads `zipalign.py --page-size 16 --pad-like-apksigner ...`, the form suggested in the review. My "does not exist" claim was wrong.

csukuangfj:
> https://github.com/paoloantinori/sherpa-onnx/actions/runs/35506035680

@paoloantinori 

You can ask AI to cancel all actions at https://github.com/paoloantinori/sherpa-onnx/actions
Otherwise, it may take a long time to run https://github.com/paoloantinori/sherpa-onnx/actions/runs/35506035680

You can change the trigger condition to a specific branch, e.g., docs-fdroid-reproducible-builds in your case.

paoloantinori:
An update on the run: the dispatch is stuck in queued state on my fork (https://github.com/paoloantinori/sherpa-onnx/actions/runs/35514677830), and a minimal echo workflow queues the same way, so the runner allocation on my account is the problem, not the workflow. GitHub's status page reports Actions operational. I am resolving it on my side and will post the completed run link here.

csukuangfj:
You can cancel  all other running  actions to run your last job.

.github/workflows/android-fdroid-reproducible-aar.yaml
```diff
@@ -0,0 +1,138 @@
...
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - '.github/workflows/android-fdroid-reproducible-aar.yaml'
+      - 'android/fdroid-reproducible-builds.md'
+      - 'cmake/**'
+      - 'sherpa-onnx/csrc/*'
+      - 'sherpa-onnx/jni/*'
```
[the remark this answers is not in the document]
paoloantinori: Addressed by removal: the workflow at the current head is workflow_dispatch only (see the comment block in the file), so this paths filter no longer exists. Refreshing the reference AAR is a manual dispatch by design, which also keeps the runner cost off every master push.

.github/workflows/android-fdroid-reproducible-aar.yaml
```diff
@@ -0,0 +1,138 @@
...
+concurrency:
+  group: android-fdroid-reproducible-aar-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-fdroid-reproducible-aar:
+    name: F-Droid reproducible AAR
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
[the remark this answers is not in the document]
paoloantinori: Addressed by removal: the workflow at the current head is workflow_dispatch only (see the comment block in the file), so this paths filter no longer exists. Refreshing the reference AAR is a manual dispatch by design, which also keeps the runner cost off every master push.

android/fdroid-reproducible-builds.md
```diff
@@ -0,0 +1,108 @@
...
+
+## 6. Verification tooling
+
+With points 1 to 5 in place, compare the rebuilt APK against the signed
+reference with
+[reproducible-apk-tools](https://github.com/obfusk/reproducible-apk-tools):
+
+```bash
+# align the rebuilt unsigned APK the way apksigner pads, 16 KiB pages
+zipalign.py --page-size 16 --pad-like-apksigner rebuilt-unsigned.apk rebuilt-aligned-unsigned.apk
```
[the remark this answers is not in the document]
paoloantinori: Applied in 8e9b4926, thank you. Without --replace an already-aligned rebuild keeps whatever padding it arrived with, which is exactly the mismatch this step exists to remove.

paoloantinori:
The run completed: https://github.com/RisorseArtificiali/sherpa-onnx/actions/runs/35517793437 (green, artifact sherpa-onnx-fdroid-reproducible-aar attached, 41.7 MB). It ran on an org fork because runner allocation on my personal fork was stuck in queued for hours that morning. The two fixes it needed along the way are already on the PR branch: the NDK install step now shields the yes SIGPIPE from pipefail, and the zipalign step in the docs page gained --replace.

.github/workflows/android-fdroid-reproducible-aar.yaml
```diff
@@ -0,0 +1,133 @@
...
+      # https://github.com/actions/setup-java
+      - uses: actions/setup-java@v4
+        with:
+          distribution: 'temurin'
+          java-version: '21'
+
+      - name: Install the pinned NDK
+        shell: bash
+        run: |
+          # The page pins this revision; ANDROID_NDK_LATEST_HOME floats.
```
csukuangfj: Could you document which page pins this version?

How can we update the version in the future?

It would be great if you could post a link to the related F-Droid doc.

.github/workflows/android-fdroid-reproducible-aar.yaml
```diff
@@ -0,0 +1,133 @@
...
+          ls -lh "$ANDROID_HOME/ndk/$NDK_VERSION"
+
+      - name: build android arm64-v8a
+        shell: bash
+        run: |
+          export CMAKE_CXX_COMPILER_LAUNCHER=ccache
+          export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
+          export ANDROID_NDK=${ANDROID_NDK}
+
+          ./build-android-arm64-v8a.sh
```
csukuangfj, line 73: Please disable C_API explicitly by using
```
export SHERPA_ONNX_ENABLE_C_API=OFF
```

The script may change the default behavior in the future. Specifying it explicitly here is future-proof.

.github/workflows/android-fdroid-reproducible-aar.yaml
```diff
@@ -0,0 +1,133 @@
...
+          for arch in arm64-v8a armeabi-v7a x86 x86_64; do
+            mkdir -p android/SherpaOnnxAar/sherpa_onnx/src/main/jniLibs/$arch/
+            cp -v ./jniLibs/$arch/* android/SherpaOnnxAar/sherpa_onnx/src/main/jniLibs/$arch/
+          done
+          ls -lh android/SherpaOnnxAar/sherpa_onnx/src/main/jniLibs/*
+
+      - name: Build aar
+        shell: bash
+        run: |
+          cd android/SherpaOnnxAar
```
csukuangfj, line 133: Can you describe how the following items mentioned in `android/fdroid-reproducible-builds.md` are implemented or checked by the CI:

- 4. Disable baseline-profile generation in the consuming project
- 5. Per-ABI APKs
- 6. Verification tooling

Other than the naming and the NDK version, this YAML file is very similar to our existing`android.yaml`, right?

csukuangfj (changes requested):

.github/workflows/android-fdroid-reproducible-aar.yaml
paoloantinori: Done in aa3734eb. The install step now names section 3 of android/fdroid-reproducible-builds.md as the pinning page and links the recipe field doc, https://f-droid.org/docs/Build_Metadata_Reference/#build_ndk, which accepts both r27c and the full revision. To update: read the recipe's ndk: value, resolve the r-string via the android/ndk release notes (each one states its ndkVersion; r27c = 27.2.12479018), and bump the workflow and the page together.

.github/workflows/android-fdroid-reproducible-aar.yaml
paoloantinori: Done in 383f71ea: all four build steps now export SHERPA_ONNX_ENABLE_C_API=OFF. I agree on the future-proofing, especially since android.yaml exports ON for its own artifact, so the two workflows would silently diverge if the script default changed. No behavior change against the green run (https://github.com/RisorseArtificiali/sherpa-onnx/actions/runs/35517793437): OFF is already the build-android-*.sh default, the export only pins it.

.github/workflows/android-fdroid-reproducible-aar.yaml
paoloantinori: They are not, and that is deliberate: this CI implements sections 1-3 of the page (pinned checkout, C API off, pinned NDK) and produces the reference AAR; sections 4-6 live in the consuming app and its fdroiddata recipe, so I added a header comment stating that scope (383f71ea). Concretely: baseline profiles are disabled by the whenTaskAdded block in the app's build.gradle.kts, per-ABI APKs come from the app's splits config, and the zipalign/apksigcopier comparison runs in fdroidserver's reproducible verification, which is what section 6 documents the commands for. And yes, the skeleton intentionally mirrors android.yaml, same four build-android-*.sh steps, so the reference matches the library you ship; the differences are the reproducibility ones (dispatch-only, pinned NDK, C API off) plus the tail: a gradle-built AAR instead of the jniLibs tarball and HuggingFace upload.

paoloantinori:
Addressed in 383f71ea and aa3734eb: explicit SHERPA_ONNX_ENABLE_C_API=OFF in all four build steps, and the NDK pin now documents its source page, the F-Droid ndk: field link, and the update procedure. Thread replies cover the scope question. The built library is unchanged relative to the verified green run.

.github/workflows/android-fdroid-reproducible-aar.yaml
```diff
@@ -0,0 +1,148 @@
...
+          distribution: 'temurin'
+          java-version: '21'
+
+      - name: Install the pinned NDK
+        shell: bash
+        run: |
+          # Revision pinned by section 3 of
+          # android/fdroid-reproducible-builds.md; ANDROID_NDK_LATEST_HOME
+          # floats. To update: read the consuming recipe's ndk: value
+          # (https://f-droid.org/docs/Build_Metadata_Reference/#build_ndk),
```
csukuangfj, line 53: Image

From 
https://f-droid.org/docs/Build_Metadata_Reference/#build_ndk
 , it does not mention the NDK version has to be r27c, so it is free to switch to a different NDK version, as long as the same version is used to build the AAR and the Android APP, right?

paoloantinori: Right: any NDK version is fine, the requirement is only that the SAME revision builds the reference AAR and the app, because the packaged .so are stripped by that NDK's llvm-strip and the output differs across revisions. r27c is simply what our fdroiddata recipe pins today. I have opened section 3 with the general rule so it reads that way: abd5fff3.

csukuangfj (approved):

